PR #22, opened 2015-02-06
Pluralize services

zeppelin:
Not sure if the original signular `service` was intentional, but since I've only seen (and used) the plural `services` in ember-cli, I guess it worth correcting. Also, if more services pop up, this change should be performed anyway.


yahoocla:
CLA is valid!

jasonmit:
LGTM :+1:

Issue #23, opened 2015-02-06
Deprecate mandatory `messages.` prefix


zeppelin:
Due to the internals of the `intl` service, it is currently required to prefix key paths with `messages.` (ex. `messages.product.info` - instead of just typing `product.info`).

If `intl-get` only uses properties on the `messages` object, I feel this is unnecessary, because:
* It exposes the service internals to the end user which might change in the future
* Makes it harder to use with i18n backends, because they know nothing about this, and therefore every message key must be prefixed to match `ember-intl`'s semantics to be usable on the client

Related to #18

caridy:
I don't know all the details about the ember-intl implementation, but in other implementations like handlebars intl we do not lock down the accessor to messages, in fact you can do your own thing:

```
myTemplate({
    data: {
         mymsgs: { foo: 'hello' }
    }
})
```

zeppelin:
The problem is that the `intl-get` helper is only performing an `Ember.get` on the `intl` service: https://github.com/yahoo/ember-intl/blob/master/app/helpers/intl-get.js#L18

Which has a computed property for the messages: https://github.com/yahoo/ember-intl/blob/master/app/service/intl.js#L55-L72

With this implementation there is no way to use `intl-get` on keys that are not inside the messages hash. In this example `foo.bar` is not accessible using `{{intl-get 'foo.bar'}}`:

```js
export default {
  locale: 'en',
  foo: {
    bar: 'baaz'
  },
  messages: {
    product: {
      info: '{product} will cost {price, number, EUR} if ordered by {deadline, date, time}',
      html: {
        info: '<strong>{product}</strong> will cost <em>{price, number, EUR}</em> if ordered by {deadline, date, time}'
      }
    }
  }
};
```

zeppelin:
But this is the least of my problem, as I'm happy to throw everything inside the `messages` hash. The problem is that i18n backends on the server doesn't follow this convention.

caridy:
`intl-get` should be flexible, replaceable and non-mandatory, just like it is in all other integrations. it is a convenience method that can be used or not to facilitate message resolution. The real question here is: can ember-intl work without `intl-get`?

zeppelin:
How does the other libraries work? Is it common to reach out to `foo.bar` when fetching a message?

```js
{
  foo: {
    bar: 'baaz'
  },
  messages: {
    product: {
      info: 'info',
      html: {
        info: '<strong>info</strong>'
      }
    }
  }
}
```

[...]

Issue #18, opened 2015-02-05
Implement a "message store"


jasonmit:
The following API should be supported.

`this.intl.addMessage('en', 'greeting', 'hello');`
`this.intl.addMessage('fr-FR', 'greeting', 'bonjour');`

Push multiple messages.  Note: this should do a merge and should not replace the entire locale's message object and should support nested objects.

```js
this.intl.addMessages('en', {
  greeting: 'hello',
  bye:      'goodbye'
});
```

Ideally these would be used on `ready`, `initialize`, or a route hook to push messages to a locale but can happen at any point at runtime.

/cc @zeppelin @ericf 

Inspired from the feedback on https://twitter.com/xeppelin/status/563319454193291264

jasonmit:
Loosely related to #3

zeppelin:
:+1: I like the idea

caridy:
:+1:

caridy:
I like it, but lets keep in mind that this is suppose to be a one time operation, let's not over complicate it with a deep merge, we can do a swallow merge, and warn if we are overriding any top level entry.

jasonmit:
#26 will resolve this

jasonmit:
`addMessage`/`addMessages` is now on master.

bakura10:
Could this actually be made automatic when the locale change? For instance, there could be a new option called root URL (app.myapp.com/assets/locales), and whenever the locale change through an observer, the library automatically tries to retrieve the new locale by appending it (so if we switch to fr, it will load app.myapp.com/assets/locales/fr.json), and automatically push all the messages.

jasonmit:
@bakura10 there are so many options when it comes to how people store their messages.  For example, some may store them externally and fetch via an API call, others may store them on a CDN, and lastly others may choose the static approach which is baked in to ember-intl.

The best I can do without terribly overcomplicating the design is provide a hook to allow people to push messages into the locale object.  If you want to sideload your messages, `ember g locale <localename>` and do not populate the locale.  The reason you need to create an empty locale is because at build time, I construct an object with CLDR data based on the modules within app/locales/*.

Now if you want to side load, push your json files to the asset/public tree and you can load them via XHR.

All that said, I'm opened to implementing what you described as another addon which wraps ember-intl.

bakura10:
No problem, that makes sense. Although I think that it is quite safe to consider locales as a kind of static asset similar to JS or CSS. Therefore, serving them through a CDN makes sense ;).

[...]

PR #26, opened 2015-02-07
[Breaking] Locale modules now extend from a Locale model (WIP)

### LINKED ISSUE #24: Allow message keys to contain dots
Basically no Ember-related i18n library does allow dots in message keys, but I've found relying solely to `Ember.get` a bit counter-intuitive. Granted, the implementation is simpler, since it's a core Ember tool to fetch key paths and is working pretty well, but most of the i18n backeds use a key-value store like Redis. Therefore, when fetching the translations from the server, these messages must be un-flattened first to build up a hierarchical JS object structure, something Ember i18n libraries can deal with.

Instead, just serializing a KV table into

```json
{
  "product.info": "{product} will cost {price, number, EUR} if ordered by {deadline, date, time}",
  "product.html.info": "<strong>{product}</strong> will cost <em>{price, number, EUR}</em> if ordered by {deadline, date, time}"
}
```

is much more straightforward on the server then the currently required

```json
{
  "product": {
    "info": "{product} will cost {price, number, EUR} if ordered by {deadline, date, time}",
    "html": {
      "info": "<strong>{product}</strong> will cost <em>{price, number, EUR}</em> if ordered by {deadline, date, time}"
    }
  }
}
```

While the filesize is somewhat bigger by default, it is negligible when gzipped. Also, the former speeds up message lookups as a byproduct.

As these two semantics are not mutually exclusive (exact key lookup first, then fall back to `Ember.get`), I see no disadvantage in supporting both.

jasonmit:
* Add `addMessages` and `addMessage` to the intl service and which delegates to the locale module

```js
this.intl.addMessage('en', 'greeting', 'hello');
this.intl.addMessage('fr', 'greeting', 'hello');

// will do a shallow merge based on feedback from @caridy
this.intl.addMessages('en' {
  greeting: 'hello',
  bye: 'goodbye'
});
```

This is extensible if you do not want your messages to hang off "messages" on the locale module.  You can override `addMessage`/`addMessages` on the locale model.

* This locale model implements a accessor method so you can implement your own getter accessor logic.
  * This is the method invoked from intl-get and intl-get no longer acts on the service, but instead the active locale module instance.

It will resolve #24 #23 #18 

All existing locale modules no longer can be POJOs and should extend from:

`import Locale from 'ember-intl/models/locale’;`


jasonmit:
One issue with this is I will likely need to configure the container so that the locale types are registered as singletons since looking up the locale multiple times should not result in multiple instances.

jasonmit:
This may also cause formats to be embedded in the locale modules versus the global.  I could make formats a computed property within the Locale base class:

```js
formats: Ember.computed(function () {
  return this.container.lookup('formats:main');
}).readOnly()
```

Something I need to think about more before I merge this

jasonmit:
Thanks @zeppelin

PR #28, opened 2015-02-09
Smoke test

zeppelin:
This is currently failing due to the changes introduced in 8268e8c59aaa5c753e5a8db016e913c16f0b7567

This could come handy in the future to avoid breaking changes that unit tests won't cover.


yahoocla:
CLA is valid!

jasonmit:
This is likely failing because there isn't a formats module.  In any case, I want to make it optional so will fix that today.



—

PR #36, opened 2015-02-10
Fix smoke tests for browsers without innerText support

zeppelin:
Compatibility table: http://www.quirksmode.org/dom/html/#t01


yahoocla:
CLA is valid!

jasonmit:
:+1:

PR #59, opened 2015-02-24
[Bugfix] format-message should not take format arguments

### LINKED ISSUE #58: Custom date format not being used in format-message
I've added a custom date formatter to simply display the short form of a week-day, but that doesn't seem to be being used in a message. It's working with `format-date` but not with `format-message`.

Here's [an example repo](https://github.com/rlivsey/ember-intl-test) containing an Ember app with the code below:

Eg:

```javascript
export default {
  date: {
    shortWeekDay: {
      weekday: "short"
    }
  }
};
```

This is then used in a message:

```javascript
import Locale from 'ember-intl/models/locale';

export default Locale.extend({
  locale: 'en',
  messages: {
    test: "should be a week-day initial: {day, date, shortWeekDay}"
  }
});
```

When displaying this in a template I'd expect the same date format to be shown:

```handlebars
<p>format-date: {{format-date date format="shortWeekDay"}}</p>
<p>format-message: {{format-message (intl-get "messages.test") day=date}}</p>
```

But instead I get the following:

screenshot 2015-02-23 20 24 41

jasonmit:
fixes #58

### DIFF
--- a/addon/utils/data.js
+++ b/addon/utils/data.js
@@ -7,7 +7,7 @@
 var IntlMessageFormat  = window.IntlMessageFormat;
 var IntlRelativeFormat = window.IntlRelativeFormat;
 
-export function addLocaleData(data) {
+export function addLocaleData (data) {
     IntlMessageFormat.__addLocaleData(data);
     IntlRelativeFormat.__addLocaleData(data);
 }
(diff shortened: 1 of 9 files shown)


yahoocla:
CLA is valid!

PR #64, opened 2015-03-03
[Enhancement] Replace CLDR with formatjs-extract-cldr-data

### LINKED ISSUE #61: Consume new "slimmed-down" CLDR dep
Keeping this opened for tracking.

/cc @ericf

jasonmit:
Fixes #61 

/cc @caridy and @ericf


yahoocla:
CLA is valid!

caridy:
LGTM

blueprints/locale/index.js
```diff
@@ -3,23 +3,20 @@
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
 
+var locales     = require('formatjs-extract-cldr-data/lib/locales');
```
ericf: We shouldn't do this, it's an anti pattern.

index.js
```diff
@@ -7,47 +7,19 @@
...
 
-
 var SilentError = require('ember-cli/lib/errors/silent');
 var serialize   = require('serialize-javascript');
 var Filter      = require('broccoli-filter');
 var path        = require('path');
 
-var extract   = require('./lib/extract');
+var cldr        = require('formatjs-extract-cldr-data');
+var cldrLocales = require('formatjs-extract-cldr-data/lib/locales');
```
ericf: Ditto

ericf:
@jasonmit it's not a good ideal to use sub modules within a package directly, so we should avoid doing that. There's no need to normalize the locales or throw if we don't have data for them since `formats-extract-cdlr-data` will do that for you under the hood already.

blueprints/locale/index.js
```diff
@@ -3,23 +3,20 @@
...
 module.exports = {
     description: 'Extract a CLDR data into an ES6 module for a given locale code',
 
-    anonymousOptions: ['fields'],
-
     normalizeEntityName: function (localeName) {
         entityName = Blueprint.prototype.normalizeEntityName.apply(this, arguments);
 
-        if (!extract.isValidLocale(localeName)) {
+        if (!locales.has(localeName)) {
```
ericf: This will throw, since `locales` will not have a `has()` function.

index.js
```diff
@@ -71,12 +43,30 @@ LocaleProcessor.prototype.transform = function (localeName, fields, pluralFn) {
...
-        fields:  ['year', 'month', 'day', 'hour', 'minute', 'second']
+    localeName = cldrLocales.locales.normalize(localeName);
+
+    var out   = {};
+    var parts = localeName.split('-');
+    
+    var data = cldr({
+        locales:        [localeName],
+        pluralRules:    true,
+        relativeFields: ['year', 'month', 'day', 'hour', 'minute', 'second']
```
ericf: The fields list is fixed in the new package, simply return `true` here.

index.js
```diff
@@ -71,12 +43,30 @@ LocaleProcessor.prototype.transform = function (localeName, fields, pluralFn) {
 LocaleProcessor.prototype.processString = function (inputString, filename) {
     var localeName = path.basename(filename, path.extname(filename));
 
-    var cldr = process(localeName, {
-        plurals: true,
-        fields:  ['year', 'month', 'day', 'hour', 'minute', 'second']
+    localeName = cldrLocales.locales.normalize(localeName);
+
+    var out   = {};
+    var parts = localeName.split('-');
```
ericf: There's no need to do this, the underlying package will automatically do this.

index.js
```diff
@@ -71,12 +43,30 @@ LocaleProcessor.prototype.transform = function (localeName, fields, pluralFn) {
...
+        relativeFields: ['year', 'month', 'day', 'hour', 'minute', 'second']
+    });
+
+    parts.map(function(part, idx) {
+        /* 
+         * if the locale is 'zh-Hant-TW' this map will build an
+         * array of strings ['zh', 'zh-Hant', 'zh-Hant-TW']
+         * which is used to walk the CLDR results to progressively
+         * build up the results object
+         */
```
ericf: Unnecessary since this is handled automatically by the underlying package.

ericf:
This internal operation is run on every call to extract data and it will automatically normalize and expand locales up to their root: https://github.com/yahoo/formatjs-extract-cldr-data/blob/master/lib/expand-locales.js

You can look at [how the Grunt package](https://github.com/yahoo/grunt-extract-cldr-data/blob/26ccd6c86d41ca70b0570721e2299d93e8e5e31e/tasks/extract_cldr_data.js) buildles up of the extracted data into files.

blueprints/locale/index.js
jasonmit: Should we be exporting this from formatjs-extract-cldr-data?  I need a way to validate the locale

index.js
jasonmit: I found it wasn't working, at least as expected.  For example, if I am targeting `zh-Hant-TW` then the object I get back from cldr function above is an object containing `zh` and `zh-Hant` and `zh-Hant` was not an aggregating of `zh` and `zh-Hant`.  Which is why I had to walk the results to build it up..

index.js
jasonmit: See comment below

index.js
jasonmit: :+1:

index.js
jasonmit: Update: I am going to have to make much larger changes to accommodate this new result object since I am no longer building one ES6 module per locale, I need to create an ES6 module per each part of the result set to prevent duplication.

jasonmit:
I am going to have to make much larger changes to accommodate this new result object since I am no longer building one ES6 module per locale, I need to create an ES6 module per each part of the result set to prevent duplication.

PR #68, opened 2015-03-11
Implementing new CLDR extracting

jasonmit:
* [x] Improve fallback functionality
* [x] Implement new message format
* [x] Implement new relative format
* [x] Move cldr extracter logic to a more reusable broccoli library (still not standalone yet)
* [x] Translations can now be loaded async
* [x] Adapter pattern implemented for overriding how translations are retrieved

This is not yet complete, waiting on relative format npm publish, but verified everything works locally.


yahoocla:
CLA is valid!

blueprints/locale/index.js
```diff
@@ -3,23 +3,20 @@
...
 module.exports = {
     description: 'Extract a CLDR data into an ES6 module for a given locale code',
 
-    anonymousOptions: ['fields'],
-
     normalizeEntityName: function (localeName) {
         entityName = Blueprint.prototype.normalizeEntityName.apply(this, arguments);
 
-        if (!extract.isValidLocale(localeName)) {
+        if (!cldrLocales.locales.has(localeName)) {
```
jasonmit: Er, can be removed

lib/broccoli-cldr.js
```diff
@@ -0,0 +1,80 @@
...
+            data = wrapEntry(localeData.locale, {
+                parentLocale: localeData.parentLocale,
+                pluralFn:     localeData.pluralRuleFunction,
+                fields:       localeData.fields
+            });
+        }
+
+        var outFile = path.join(destPath, localeKey.toLowerCase() + '.js');
+        var file = options.prelude + data;
+        console.log(file);
```
jasonmit: Remove

app/adapters/-intl-adapter.js
```diff
@@ -0,0 +1,40 @@
...
+import Locale from 'ember-intl/models/locale';
+import IntlAdapter from 'ember-intl/adapter';
+
+function normalize (fullName) {
+    Ember.assert('Lookup name must be a string', typeof fullName === 'string');
+
+    return fullName.toLowerCase();
+}
+
+export default IntlAdapter.extend({
```
jasonmit: Not documenting this yet, but it will eventually be exposed for those people who want to specify their own way of loading translations.  The CLDR portion of this hasn't been thought out yet, but it will ideally be another hook off this.

See: https://github.com/yahoo/ember-intl/issues/69

package.json
```diff
@@ -27,18 +27,22 @@
...
+    "broccoli-caching-writer": "^0.5.5",
     "broccoli-filter": "^0.1.11",
-    "broccoli-funnel": "^0.2.0",
-    "cldr": "2.3.2",
+    "broccoli-funnel": "^0.2.2",
+    "formatjs-extract-cldr-data": "^1.1.0",
     "intl": "0.1.4",
-    "intl-messageformat": "1.0.4",
+    "intl-messageformat": "1.1.0",
     "intl-relativeformat": "1.0.3",
```
jasonmit: I plan to bump this now that you published

lib/broccoli-cldr.js
```diff
@@ -0,0 +1,79 @@
...
+            });
+        }
+
+        var outFile = path.join(destPath, localeKey.toLowerCase() + '.js');
+        var file = options.prelude + data;
+        fs.writeFileSync(outFile, file, { encoding: 'utf8' });
+    }, this);
+}
+
+LocaleWriter.has = function (locale) {
```
jasonmit: We talked about this a few weeks ago, and this is what I think we agreed on.  Let me know if you have a better idea now

README.md
```diff
@@ -52,8 +52,8 @@ export default Locale.extend({
 ```js
 	var App = Ember.Application.extend({
 		ready: function () {
-			// setup default locale to fr but will fallback to en
-			this.intl.set('locales', ['fr-FR', 'en']);
+			// setup default locale to navigator.language but will fallback to en
+			this.intl.set('locales', [navigator.language, 'en']);
```
ericf: You should link to this: http://formatjs.io/guide/#client-side since it talks about the fallback for older browsers.

index.js
```diff
@@ -141,5 +91,29 @@ module.exports = {
         }));
 
         return this.mergeTrees(trees, { overwrite: true });
+    },
+
+    _transformLocale: function (locale, result) {
+        var data = ['/*jslint eqeq: true*/', 'export default {'];
```
jasonmit: /*jslint eqeq: true*/ could go into the prelude

README.md
jasonmit: :+1:

app/adapters/-intl-adapter.js
```diff
@@ -0,0 +1,40 @@
...
+    findLocale: function (locale) {
+        if (locale instanceof Locale) {
```
ericf: What's `Locale` here, a class? If so can this be simplified since it is just data?

jasonmit: Ya, a class: https://github.com/yahoo/ember-intl/blob/master/addon/models/locale.js

app/services/intl.js
```diff
@@ -4,9 +4,9 @@
  */
 
 import Ember from 'ember';
-import Locale from 'ember-intl/models/locale';
-import createFormatCache from 'ember-intl/format-cache/memoizer';
 import { IntlRelativeFormat, IntlMessageFormat } from 'ember-intl/utils/data';
+import createFormatCache from 'ember-intl/format-cache/memoizer';
```
ericf: I think there might have been a bug fix recently in the `intl-format-cache` package. Are you using that package? If not can we switch to using it? Otherwise you might want to make sure any bug fix that happened in that package is done here too.

jasonmit: Ya, there was a bug in browserify that was in a more recent build.  First I need to fix ember-browserify in order to use this.  Plan on doing this though.

blueprints/locale/index.js
```diff
@@ -3,23 +3,20 @@
...
 var fs          = require('fs');
 
-var extract     = require('../../lib/extract');
-
 module.exports = {
     description: 'Extract a CLDR data into an ES6 module for a given locale code',
 
-    anonymousOptions: ['fields'],
-
     normalizeEntityName: function (localeName) {
```
ericf: Why do you need this here?

ericf: I see because this is driven by the app's translation files…

lib/broccoli-cldr.js
```diff
@@ -0,0 +1,79 @@
...
+    var destPath    = path.join(destDir, this.outputPath);
+    var options     = this.options;
+    var wrapEntry   = options.wrapEntry;
+    var localesData = cldr(options);
+
+    mkdirp.sync(destPath);
+
+    Object.keys(localesData).forEach(function (localeKey) {
+        var localeData = localesData[localeKey];
+        var data       = localeData;
```
ericf: Two vars pointing to the same object seems weird here.

lib/broccoli-cldr.js
```diff
@@ -0,0 +1,79 @@
...
+
+        if (typeof wrapEntry === 'function') {
+            data = wrapEntry(localeData.locale, {
+                parentLocale: localeData.parentLocale,
+                pluralFn:     localeData.pluralRuleFunction,
+                fields:       localeData.fields
+            });
+        }
+
+        var outFile = path.join(destPath, localeKey.toLowerCase() + '.js');
```
ericf: I don't think you want to do this because there will be missing information when a locale like `en-US` delegates to its parent `en` for all the real data.

lib/broccoli-cldr.js
jasonmit: err.. :)

lib/broccoli-cldr.js
ericf: I think you'll want to aggregate the data for this locale plus its ancestry in this one file if you want one file per locale vs. one file per language.

jasonmit: I really need a way to support both.  One where you get the partial objects and one where you get the concatted CLDR.  I can speak offline about the two use cases.

blueprints/locale/index.js
jasonmit: Correct, spits out an empty translation file which extends from `Locale` class.  I am really considering naming the locale files `translations`, because that's all we store in them at the moment.  It's clearly confusing :)

app/adapters/-intl-adapter.js
ericf: I feel like this is too much abstraction for data that can be represented as a hash. But this might need to stay for back-compat?

app/services/intl.js
ericf: What was the bug? That package has no deps and a single default export. But it's also not compiled to CommonJS, it's simply a pure ES6 module…maybe that's the issue?

blueprints/locale/index.js
ericf: Yeah, +1 for app strings being called "translations" or "messages". Whereas in the other FormatJS libs "locale data" is the generic, non-app-specific data for formatting dates/numbers/plurals/etc.

index.js
```diff
@@ -141,5 +91,29 @@ module.exports = {
...
+
+        data.push('  locale: "' + locale + '",');
+
+        if (result.parentLocale) {
+            data.push('  parentLocale: "' + result.parentLocale + '",');
+        }
+
+        if (result.fields) {
+            data.push('  fields: ' + serialize(result.fields) + ',');
```
ericf: You can build up a JSON/JavaScript object and then call `serialize()` once for the whole thing.

index.js
```diff
@@ -141,5 +91,29 @@ module.exports = {
         }));
 
         return this.mergeTrees(trees, { overwrite: true });
+    },
+
+    _transformLocale: function (locale, result) {
```
ericf: It seems like this whole function could simply be:

```js
return 'export default ' + serialize(result) + ';';
```

lib/broccoli-cldr.js
ericf: Instead of talking about `en.js` as a parent-level locale, we can simply call it a "language" since I think that will be a simpler mental model for people.

lib/broccoli-cldr.js
ericf: I find it weird to call a utility function on a broccoli plugin; I haven't seen that before. Is this common?

app/adapters/-intl-adapter.js
jasonmit: It was needed two resolve the following issues:
https://github.com/yahoo/ember-intl/issues/46 (getValue)
https://github.com/yahoo/ember-intl/issues/24 (getValue)
And https://github.com/yahoo/ember-intl/issues/18 (addMessage)

app/services/intl.js
jasonmit: That was the issue, I can revisit and figure out a solution for this.

The issue with browserify was not being able to import both mf and rf.  That should be fixed now so soon I plan to remove to global references to these.

index.js
jasonmit: :+1:

lib/broccoli-cldr.js
jasonmit: :+1:

lib/broccoli-cldr.js
jasonmit: Not common.  Would you prefer a named export exposing this method?

index.js
jasonmit: fixed

blueprints/locale/index.js
jasonmit: This PR is getting big, so I'll tackle this afterwards

index.js
jasonmit: :+1: fixed

lib/broccoli-cldr.js
jasonmit: > I don't think you want to do this because there will be missing information when a locale like en-US delegates to its parent en for all the real data.

If these files were to be written to the file system and kept as static files, this would be the case.  What this is doing ultimately is writing to the "application tree" by broccoli, which get bundled into app.js all together.  So if my app locale is 'en-us', it will write 'en-us' and 'en' with 'en-us' not having any data other than `parentLocale` set to `en`.

When the application is initialized I call addLocaleData on each of the CLDR objects written to the app tree.  What you mention is in fact an issue though when pulling out this broccoli plugin for others to have more control over this extraction process.  One use-case I've heard is the need to write them to the asset tree, so that they are not bundled at all .. instead individually stored.  That's where your proposal of "languages" comes in and will definitely solve this.

PR #72, opened 2015-03-20
[Enhancement] adding `now` intl.relativeformat  argument and removing buildOptions

### LINKED ISSUE #70: Add now prop to format relative
See: https://github.com/yahoo/react-intl/commit/016eafb

/cc @ericf

jasonmit:
Fixes #70 

/cc @ericf

### DIFF
--- a/addon/formatter-base.js
+++ b/addon/formatter-base.js
@@ -6,24 +6,6 @@
 import Ember from 'ember';
 
 var FormatBase = Ember.Object.extend({
-    buildOptions: function (value, hash) {
-        var args = [value];
-
-        if (typeof hash.format !== 'undefined') {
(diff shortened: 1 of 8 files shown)


yahoocla:
CLA is valid!

addon/formatter-base.js
```diff
@@ -45,7 +27,8 @@ var FormatBase = Ember.Object.extend({
 });
 
 FormatBase.reopenClass({
-    formatOptions: Ember.A()
+    formatOptions: Ember.A(['locales', 'format']),
```
jasonmit: treats locales and format as allowable arguments that get passed on through to the formatter constructors as the second argument.

Issue #25, opened 2015-02-06
Documentation on how to include ember-intl in test suite


terryroe:
It would be very helpful to have some documentation showing how to use ember-intl in ember-cli tests.  I currently get messages like this when running my tests:
```
Handlebars error: Could not find component or helper named 'format-relative'
Handlebars error: Could not find component or helper named 'intl-get'
```
In order to get past these, I stubbed out these methods in my setup method  like this:
```
    component = this.subject();
    // stubs for ember-intl
    component.intl = {};
    component['format-relative'] = component.intl.get = function (str) {
      return str;
    };
```
But I would rather know how to include the library in the test suite.

Thanks.

TR

jasonmit:
Unit testing may be a problem, acceptance shouldn't be.  

I'll think of a way to best solve this, perhaps by creating a utility function you would need to invoke in the setup hook of your unit test.  Would that be acceptable?

terryroe:
The above problems I ran into were in unit tests for components.  It would be nice if there were a better way to stub things out, but I can live with my approach for now.  A utility function for setup would be great.  Thanks.

jasonmit:
With my latest changes (1.1.0), it's a little bit easier.  Assuming you are on 1.10:
https://github.com/yahoo/ember-intl/blob/master/tests/unit/format-date-test.js#L13-L14

You'll always want to register an intl:main into the Container.  

```
import Intl from 'app/services/intl';

// in your setup
container.register('intl:main", Intl, { instantiate: true, singleton: true });
```

jasonmit:
#77 will resolve this. Feel free to test the branch out and follow the changeset in the new README.

Let me know if I should expand on anything further.

jasonmit:
1.3.0 released today, which resolves some testing concerns.

Also, here is a pretty simply test seeing if a component including format-message rendered: https://github.com/jasonmit/ember-intl-1.3.0-test-example

PR #77, opened 2015-03-22
[Enhancement] unit test documentation and enhancements

### LINKED ISSUE #76: Unable to lookup the "intl:main" service in unit tests
At the moment I'm unable to write unit tests for components that use of the ```format-message``` and ```intl-get``` helpers in the component template.

Initially, this is what I've tried when just setting the test up:
```
needs: [
  'locale:en-GB', 'helper:format-message', 'helper:intl-get'
]
```
This results in an exception when calling ```this.render()``` in the actual tests:
```TypeError: Cannot read property 'on' of undefined```, the stacktrace points to [this](https://github.com/yahoo/ember-intl/blob/master/app/helpers/intl-get.js#L45) line. Which makes sense as ```'intl:main'``` is missing from the needs block.

I tried adding ```'intl:main'``` to the needs block:
```
needs: [
  'intl:main', 'locale:en-GB',
  'helper:format-message', 'helper:intl-get'
]
```
Which also resulted in an exception:
```TypeError: Attempting to register an unknown factory: `intl:main````.

After digging through the ember-intl source I noticed that the ember-intl initializer sets up the intl service as ```intl:main``` instead of ```service:intl```. Trying to inject ```service:intl``` in the needs block doesn't work either.

My best guess is that initializers don't run in a (module) unit test, resulting in ```intl:main``` never being registered on the container/app. Because the ```intl-get``` helper injects ```intl:main``` rather than ```service:intl```, adding ```service:intl``` to the test's need block won't do anything.

At the moment, I don't even care what format-message/intl-get renders, stubbing out whatever ```format-message``` and/or ```intl-get``` needs would work just fine for me as well, I just want to be able to render the component's template without exceptions. (I just don't know how to do this)

My actual question(s):
* Am I doing something wrong?
* Is there any documentation that I missed that explains how to do this/make it work?
* If I'm not doing anything wrong, is there a way to work around this (eg. stubbing ```intl``` on ```intl-get```)?

Any answers or pointers in the right direction would be greatly appreciated!

jasonmit:
* [x] Document how to register CLDR objects
* [x] Export named initializer function from initializer module
* [x] 1.9 helpers should use the view's container
* [x] Renamed `intl:main` to `service:main` to adhere to new Ember lookup service `Ember.inject.service`

fixes #76, #25


yahoocla:
CLA is valid!

PR #82, opened 2015-03-25
[BREAKING] Removing intl injection into all popular types

### LINKED ISSUE #80: Remove injecting `intl` into all objects
Once we remove support for < 1.10 we can remove the Ember initializer which handles injecting `intl` instance into all popular types (view, controllers, models, components, routes).

Instead, consumers can opt for `Ember.inject.service` to optionally inject it:

```js
export default Ember.Component.extend({
  intl: Ember.inject.service()
});
```

This would obviously be a breaking change.

jasonmit:
The alternative is to use `Ember.inject.service`:

```js
export default Ember.Component.extend({
  /* note: intl is a computed property so you must use
Ember.get/Ember.set to access */
  intl: Ember.inject.service()
})
```

fixes #80

### DIFF
--- a/addon/formatter-base.js
+++ b/addon/formatter-base.js
@@ -6,6 +6,8 @@
 import Ember from 'ember';
 
 var FormatBase = Ember.Object.extend({
+    intl: Ember.inject.service(),
+
     filterFormatOptions: function (hash) {
         hash = hash || {};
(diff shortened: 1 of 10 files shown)


yahoocla:
CLA is valid!

PR #85, opened 2015-03-26
[Enhancement] Throw when intl-get is used standalone

### LINKED ISSUE #84: intl-get template helper now returns an object instead of just the message string
We were using intl-get in templates like so:
```
{{intl-get 'messages.key.otherKey'}}
```
and it would return the message string.

It appears that this no longer works and intl-get returns an object.  We now have to change all our code from above to something like this:
```
{{format-message (intl-get 'messages.key.otherKey')}}
```
Was the first example above an intended use of intl-get?  Or was it always supposed to be used with format-message?

jasonmit:
Fixes #84

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -79,13 +79,13 @@ export default Ember.Service.extend(Ember.Evented, {
     }),
 
     addMessage: function (locale, key, value) {
-        return this.getLanguage(locale).then(function (localeInstance) {
+        return this.findLanguage(locale).then(function (localeInstance) {
             return localeInstance.addMessage(key, value);
         });
     },
 
     addMessages: function (locale, messageObject) {
-        return this.getLanguage(locale).then(function (localeInstance) {
+        return this.findLanguage(locale).then(function (localeInstance) {
             return localeInstance.addMessages(messageObject);
         });
     },
@@ -175,7 +175,7 @@ export default Ember.Service.extend(Ember.Evented, {
         }
     },
 
-    getLanguage: function (locale) {
+    findLanguage: function (locale) {
         var result = this.get('adapter').findLanguage(locale);
 
         return Ember.RSVP.cast(result).then(function (localeInstance) {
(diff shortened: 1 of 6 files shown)


yahoocla:
CLA is valid!

addon/services/intl.js
```diff
@@ -175,7 +175,7 @@ export default Ember.Service.extend(Ember.Evented, {
         }
     },
 
-    getLanguage: function (locale) {
+    findLanguage: function (locale) {
```
jasonmit: These are not yet documented and private, I am changing them to match what gets invoked on the adapter for consistency.  In 1.3.0 I plan to document this and mark it as public.

tests/unit/format-message-test.js
```diff
@@ -63,6 +63,17 @@ test('should throw if called with out a value', function(assert) {
...
 });
 
+test('should throw intl-get is used standalone helper', function(assert) {
+    expect(1);
+    view = this.intlBlock('{{intl-get "messages.foo.bar"}}');
+
+    try {
+        runAppend(view);
+    } catch (ex) {
+        assert.ok(ex, 'intl-get threw when run standalone');
```
jasonmit: `assert.throws` doesn't pick up the Ember.assert so resorting to a good ol' try/catch + expect

app/helpers/intl-get.js
```diff
@@ -7,6 +7,8 @@ import Ember from 'ember';
 import { Stream, read, readHash, destroyStream } from 'ember-intl/utils/streams';
 
 export default function (value, options) {
+    Ember.assert('intl-get helper must be used as a subexpression', options.isInline === true);
```
jasonmit: Future versions I plan to deprecate intl-get with a warning in favor of a `t` helper.  Intl-get no longer shares the same concepts of its sibling libs since it only is used to resolve translations keys.

`(intl-get 'messages.foo.bar')` will become `(t 'foo.bar')`

This change will likely happen when I rename `locales` -> `translations`

PR #86, opened 2015-03-26
[Bugfix] Removing currentValue in base helper

gtimti:
We faced problems that are generally related to base helper.

There is a `currentValue` variable in the `helpers/base.js`, which is used to store the current value obtained from input stream. This variable is not safe, as it can be used for writing and reading at the same time.
For example, we use the `format-number` helper to display the calculated values. In our case a user enters the initial values ​​in specific fields. As a result, the form displays incorrect calculated values. There is no problem if we don't apply formatting (using the plain output). Investigating the problem we noticed that the `currentValue` variable still stores previous value at the time of applying formats. The problem was resolved by getting rid of the `currentValue` variable and read values ​​directly from the input stream.

Thanks.

### DIFF
--- a/addon/helpers/base.js
+++ b/addon/helpers/base.js
@@ -16,7 +16,7 @@ export default function (formatterName) {
             return throwError();
         }
 
-        var currentValue, outStream;
+        var outStream;
 
         function touchStream () {
             outStream.notify();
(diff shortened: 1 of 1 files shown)


jasonmit:
> This variable is not safe, as it can be used for writing and reading at the same time.

Just so I fully understand, you were doing the following:

```hbs
{{format-number num}}
{{input value=num}}
```

Is that correct?

jasonmit:
LGTM, merged.  I'm still interested in your use case so I can add tests around this :)

Issue #96, opened 2015-04-16
Missing message Error


al3x-edge:
Haven't yet tried stable, but noticed on 2.0 branch _(Yup. I'm that guy)_ that the error message returned when there is no message available to translate is this: `findTranslation should return an object of instance 'IntlGetResult'`.

I noticed that there is an error below that, but it is never hit due to the assert exiting.
```js
if (typeof translation === 'undefined') {
    throw new Error('translation: `' + key + '` on locale(s): ' + locales.join(',') + ' was not found.');
}
```

I bring this up because when switching to 2.0 for the ability to get simple values returned during translation to output in HTML (not output -- supposed to be copied into clipboard).  I had the previous `messages.something.translateMe` which was throwing the assert and I couldn't figure it out until about 30 minutes later I remembered I took those out. 

Are we able to throw the assert after the error so that people are able to see they are missing the key to be translated also?

jasonmit:
So you're using format-message and passing in a key that doesn't have a translation?  I'm unsure of the use-case there, can you explain more?  I am opened to changing this.

al3x-edge:
@jasonmit Aye. Exactly.  Not really a use case; more of a helping users thing.  The assertion happens and then the error isn't displayed to users that their key was wrong -- Just that they should expect an `IntlGetResult` -- Kind of cryptic ya know?

jasonmit:
:+1: I'll fix this

jasonmit:
LGTM

PR #97, opened 2015-04-17
Remove Ember.Assert for IntlGetResult

al3x-edge:
This will now display to others than they are missing a key in their
messages instead of a more cryptic message saying that a result is not
of type IntlGetResult.

Fixes issue #96.

Attempted to add a test to catch the case, but I test less than I should.  The `assert.throws` with a sub-expression seems not to pick up the error. Not sure if it's because the error happens in a different place than it's expecting due to it being a sub-expression.

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -6,7 +6,6 @@
 import Ember from 'ember';
 import { IntlMessageFormat } from '../utils/data';
 import createFormatCache from '../format-cache/memoizer';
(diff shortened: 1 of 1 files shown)


yahoocla:

al3x-edge:
Not sure if the CLA check will re-run, but the CLA has been signed.

yahoocla:
CLA is valid!

jasonmit:
:+1:

PR #99, opened 2015-04-21
Update README to new addon install syntax

elwayman02:
Also, call out major browsers that will need the Intl shim


yahoocla:
CLA is valid!

jasonmit:
Can you update it to support < 0.2.3 language as well since this command is relatively new

elwayman02:
Yes

elwayman02:
@jasonmit added

jasonmit:
:+1:

PR #108, opened 2015-05-19
Fix warnings to output locale code

al3x-edge:
Fixes the issue where `undefined` is coming up instead of the specific
translation code.

I'm not sure if there is an easier way to parse the locale while in the broccoli merge-tree stuff. Just was taking a crack to hopefully help you move the 2.0 WIP branch.

### DIFF
--- a/lib/translation-blender.js
+++ b/lib/translation-blender.js
@@ -92,13 +92,14 @@ TranslationBlender.prototype.updateCache = function (includePaths, destDir) {
     var translationsSrc = includePaths[0];
     var outputPath = path.join(destDir, this.options.outputPath);
     var base = this.readAsObject(path.join(translationsSrc, baseLocale + '.json'));
-    var translation, filename;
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

jasonmit:
Thanks for taking this on, it was a bug I had listed for 2.0 but just haven't had the time to get 2.0 moving forward the last few weeks!  All PRs are much appreciated :)

al3x-edge:
No problem @jasonmit =] -- I'll see what I can get finished up. Glimmer support is probably the biggest for us and just making sure we can stay on the beta track.  Is that still being held up by the `env` change for the most part?

jasonmit:
Yeah, we're blocked on that until the stateful helper feature lands.  I'll likely have to punt on that making it into 2.0, and will land an update the second there is support for it in canary. @al3x-edge

PR #113, opened 2015-06-03
Split initializers into initialzer and instance-initializer

al3x-edge:
Leverage instance-initalizer for injecting intead of initializer.

This currently doesn't have all of the registry lookups fixed, but that's going to have to wait until later. Thought I might get you something to look at.

### DIFF
--- a/.watchmanconfig
+++ b/.watchmanconfig
@@ -1,0 +1,3 @@
+{
+  ignore_dirs: ["tmp"]
+}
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

jasonmit:
Thanks again! Merging since I already know about this failing test and will fix it Friday.

al3x-edge:
@jasonmit Not a problem. Trying to help get you over the 2.0 hump and to get it stable. /2.0syndrome

PR #117, opened 2015-06-08
Fix deprecations with registry / container

al3x-edge:
Splitting the intializers into normal and instance initializers helped,
but it didn't remove the deprecations that are currently being output.
After speaking with @runspired, this is a beta idea to fix the
createLocale to be used inside and outside of an initializer without
deprecations.

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -117,15 +117,15 @@ export default Ember.Service.extend(Ember.Evented, {
         return message.format(values);
     },
 
-    createLocale(locale, payload) {
+    createLocale(instance, locale, payload) {
         let name = 'ember-intl@translation:' + locale;
-        let modelType = this.container.lookupFactory('ember-intl@model:translation');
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

runspired:
The build error is because phantom 1.9 doesn't have `function.bind`, there's a patch available or you can upgrade to phantom 2.0

al3x-edge:
@runspired Aye. Did notice that in local tests. Upgraded to phantom 2.0 locally.  Deciding it's probably better to include the polyfill for now until it's default on Travis -- others may not have Phantom 2.0 installed.

jasonmit:
LGTM :+1:

I'll include the polyfill and add a comment above the `service.createLocale = service.createLocale.bind(service, instance);` describing why it's there

al3x-edge:
Awesome. Thanks @jasonmit

PR #118, opened 2015-06-08
Squash dat getter-setter non-sense.

al3x-edge:
Removes deprecation for using the same function as getter and setter.

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -6,6 +6,7 @@
 import Ember from 'ember';
 import { IntlMessageFormat } from '../utils/data';
 import createFormatCache from '../format-cache/memoizer';
+import computedPolyfill from 'ember-new-computed';
 
 var makeArray = Ember.makeArray;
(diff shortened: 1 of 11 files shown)


yahoocla:
CLA is valid!

addon/services/intl.js
```diff
@@ -6,6 +6,7 @@
 import Ember from 'ember';
 import { IntlMessageFormat } from '../utils/data';
 import createFormatCache from '../format-cache/memoizer';
+import computedPolyfill from 'ember-new-computed';
```
jasonmit: call it `computed` and replace all instances of Ember.computed with ember-new-computed :+1:

al3x-edge:
@jasonmit Updated to change all references to `Ember.computed` to `computed` from `ember-new-computed`.

jasonmit:
Thanks again Alex :) :+1:

al3x-edge:
Not a problem :+1:

PR #119, opened 2015-06-10
Ember-try dependencies for older Embers

al3x-edge:
`ember-load-initializers#0.1.4` is incompatible with older versions of Ember. Just a concept to see if this is how to test old versions or if there is a workaround needed for the undefined initializers.


yahoocla:
CLA is valid!

jasonmit:
0.1.4 should be compatible still unless I am overlooking soemthing

al3x-edge:
Awesome. :+1:

PR #122, opened 2015-06-13
Pin Ember to 1.12.x until ready for 1.13

al3x-edge:
Currently, the caret will allow Ember to upgrade to 1.13 which causes all sorts of failed tests. This fixes that by being more specific.


yahoocla:
CLA is valid!

jasonmit:
:+1:

PR #123, opened 2015-06-15
yaml support

jasonmit:


yahoocla:
CLA is valid!

caridy:
+1 for what it worth.

PR #125, opened 2015-06-16
[WIP] Stateful Helper Glimmer Support

al3x-edge:
Very rough, nasty, test-failing monster.  Just wanted to get your thoughts or suggestions on fixing some of the lookup errors.

### DIFF
--- a/app/instance-initializers/ember-intl.js
+++ b/app/instance-initializers/ember-intl.js
@@ -16,6 +16,7 @@ import FormatMessage from 'ember-intl/helpers/format-message';
 export function instanceInitializer(instance) {
     let service = instance.container.lookup('service:intl');
 
+    service.set('appInstance', instance);
     service.createLocale = service.createLocale.bind(service, instance);
 
     filterBy(ENV, 'cldrs').forEach((key) => {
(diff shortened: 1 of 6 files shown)


yahoocla:
CLA is valid!

app/instance-initializers/ember-intl.js
```diff
@@ -10,6 +10,7 @@ import { filterBy } from 'ember-intl/utils/initialize';
 export function instanceInitializer(instance) {
     let service = instance.container.lookup('service:intl');
 
+    service.set('appInstance', instance);
```
jasonmit: Not sure we need this, I'll do some investgating on my part

al3x-edge: Aye. Felt dirty. Just not sure where else to get the reference for it. I think there is `this.container._registry` but being marked with an underscore and all I didn't investigate further.

jasonmit: Why not `this.container.lookup`?

al3x-edge: That's what it was before the change. I thought I was getting errors from it but maybe it was related to something else instead.

jasonmit: #127 it all seems to be working except for tests and one other comment I pointed out which we need to guard against for backwards compat

jasonmit:
Tests are now passing on #127 some clean up is needed before I merge it

PR #130, opened 2015-06-23
first pass at icu helper

jasonmit:
This PR removes the need for the intl-get helper and instead any arguments passed into format-message are assumed to be a path to a translation.  If you want to pass a literal string you now must use the `icu` helper.

```hbs
{{format-message (icu '{price, number, EUR}') price=1000}}
{{format-message 'product.info' price=1000 name='Macbook'}}
```

The reason for this is passing a string literal is not as popular of a use-case as passing the translation path.

* [ ] document
* [ ] support < 1.13
* [ ] fix failing tests


yahoocla:
CLA is valid!

caridy:
When we first tried this, we decided to drop it (in plain handlebars) due to the ambiguity associated with it. What should `icu` return? obviously not an string, but an object with some specific signature that I don't know. This makes this more complicated to understand, and less flexible. What if I don't want to use icu() and I don't want to use path? what should I do then? /cc @ericf

rlivsey:
We've a fairly large Ember app with several hundred usages of `format-message` and every single one of them is paired with `intl-get`. Dropping `intl-get` would clean up our templates immensely. 

Obviously a sample size of 1, but I can't personally think of anywhere we'd use the `icu` helper in our app, but maybe it could be a separate helper instead of overloading `format-message` with both behaviours?

```handlebars
{{format-message "path.to.translation"}}
```
and
```handlebars
{{format-icu '{price, number, EUR}') price=1000}}
```
Alternatively keep `format-message` taking an ICU string and have a new shorthand helper for taking a path, eg 

```handlebars
{{t "path.to.translation"}}
```

caridy:
@rlivsey it gets more complicated when you want to deal with fallbacks (e.g.: what happen when the path does not exists?), that's why we are in favor of `format-message` to only consume message objects, and by creating that indirection (in this case with `intl-get`, or `icu` or whatever please you), you have the ability to control the message resolution and fallback strategy and the library does not care about it.

caridy:
> sample size of 1

believe, when it comes to intl, everyone is doing it differently.

jasonmit:
> What if I don't want to use icu() and I don't want to use path? what should I do then?

@caridy if you're not passing a path or an ICU string then what are you passing?  I can only think of a case where you are using one or the other.  If you need to do something else you could likely invoke the service method directly to format message strings.  `this.intl.formatMessage(str, ...args);`  The service method for formatting a string will continue to expect an ICU string.

The ultimate goal in this change is to align with a similar syntax to ember-i18n.  The first major thing to is to align the helper APIs, alias `format-message` to `t` and make paths the default value.

caridy:
@jasonmit let me clarify, when I reference to `message objects`, I'm talking about a string representation of the ICU message, the one that we parse and cache internally. A `path` does not qualify, and in fact, the `path` is not something `format-message

jasonmit:
@caridy I am still unable to follow the reasons against the message helper accepting a path instead of an ICU string as a param.  I don't want to miss a use-case that I'm not thinking about, so lets jump on a call tomorrow so I can understand before this goes further.

jasonmit:
> Alternatively keep format-message taking an ICU string and have a new shorthand helper for taking a path, eg

> {{t "path.to.translation"}}

I feel this would just lead to confusion.

The reason for the ICU helper is to cover the case where you generate a translation string at runtime.  It's not popular, but there is a use for it.

> Obviously a sample size of 1, but I can't personally think of anywhere we'd use the icu helper in our app, but maybe it could be a separate helper instead of overloading format-message with both behaviours?

This may work, perhaps `format-inline-message`?

caridy:
let's talk today @jasonmit

jasonmit:
I will close for now until I get a chance to discuss with @caridy.  This won't likely land in 2.0 anymore.

Issue #133, opened 2015-06-27
[Beta] Could not find module 'ember-new-computed'


al3x-edge:
Currently getting `Uncaught Error: Could not find module 'ember-new-computed' imported from 'ember-intl/services/intl'` in a dummy-app (https://github.com/al3x-edge/ember-intl-beta1-dummy-app).

jasonmit:
Fixed, use beta.4 at the moment

PR #135, opened 2015-06-27
Fix ember-new-computed module not found

al3x-edge:
`ember-new-computed` needs to be a dependency instead of a devDependency.

Fixes issue #133.


yahoocla:
CLA is valid!

jasonmit:
Fixed this already, didn't see the PR.  Please use beta.4 :+1:

al3x-edge:
:+1:

PR #137, opened 2015-06-27
error message for invalid locale in translation path

jasonmit:
```
[ember-intl] 'en-foo' does not match a supported locale name.
Full list of supported locales: https://github.com/yahoo/formatjs-extract-cldr-data/tree/master/data/main
```


yahoocla:
CLA is valid!

PR #149, opened 2015-07-06
Instruct to setup application-wide locale when addon is installed

jasonmit:
After `ember install`, instruct how to setup your runtime locale


jasonmit:
Trying to kick travis

PR #183, opened 2015-07-29
Remove observer and add exists lookup method

jasonmit:
`exists` is to bring in the i18n API, similar to addTranslation/addTranslations for the eventually converging  of the two projects.

Use `setLocale` instead of `Ember.set` since it's possible for there to be some asynchronous behavior (fetching of translations) and validating that translations exist for the locale before setting.  So the helpers and whatever else is listening for locale changes should really be listening for an event instead, notifying that the locale change is ready.

### DIFF
--- a/addon/helpers/-base.js
+++ b/addon/helpers/-base.js
@@ -6,9 +6,9 @@
 import Ember from 'ember';
 import extend from '../utils/extend';
 
-const { get, getProperties } = Ember;
+const { get } = Ember;
 
-export default function (formatType) {
+const helperFactory = function (formatType) {
     function throwError () {
         return new Error(`${formatType} requires a single unname argument. {{format-${formatType} value}}`);
     }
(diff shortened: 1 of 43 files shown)


addon/helpers/-base.js
```diff
@@ -19,12 +19,15 @@ export default function (formatType) {
...
             this._super(...arguments);
+            const intl = get(this, 'intl');
             this.formatter = this.container.lookup(`ember-intl@formatter:format-${formatType}`);
+            intl.on('localeChanged', this, this.recompute);
         },
 
-        onIntlLocaleChanged: observer('intl.locale', function() {
-            this.recompute();
-        }),
+        willDestroy() {
```
stefanpenner: This should be destroy. Now will destroy

addon/helpers/intl-get.js
```diff
@@ -5,19 +5,28 @@
...
-        onIntlLocaleChanged: observer('intl.locale', function() {
-            this.recompute();
-        }),
+        init() {
+            this._super(...arguments);
+            const intl = get(this, 'intl');
+            intl.on('localeChanged', this, this.recompute);
+        },
+
+        willDestroy() {
```
stefanpenner: Destroy not will destroy

addon/helpers/-base.js
jasonmit: When to use `willDestroy`?  Seems `destroy` schedules a `willDestroy`

PR #191, opened 2015-08-11
Fixed initializers for Ember 2.0 (including beta/canary)

mdehoog:
Fixes the `is-modern` ember-cli util to include Ember 2.0 versions (including pre-release), so that the initializers and instance-initializers are correct.

Also fixes the registry lookup in the `createLocale()` function in the `intl` service.

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -103,6 +103,8 @@ const IntlService = Service.extend(Evented, {
 
         if (instance.registry) {
             container = instance.registry;
+        } else if (container.registry && container.registry.register) {
+            container = container.registry;
         }
 
         if (container.has(name)) {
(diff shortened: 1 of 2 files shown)


jasonmit:
Looks good.  Tests are failing on Ember 1.10.

mdehoog:
Thanks @jasonmit. Added an extra check for Ember 1.10, should be good now.

Interestingly, the `should respect format options for date ICU block` test in `format-message-test.js`, and the `Acceptance: Smoke: format-date` test both fail for me locally (in Australia). Setting my timezone to the US causes all tests to pass. Is it possible to pass timezone information to the `format-message` helper?

jasonmit:
>  Is it possible to pass timezone information to the format-message helper?

It is, https://github.com/yahoo/ember-intl/wiki/Format-Date-&-Time-Options

PR #194, opened 2015-08-12
[BUGFIX] Funnel#files is meant to be a fully expanded list of files, …

stefanpenner:
…not a glob

Funnel#includes can be used for globs.

This is meant as a 1.3.x release, which means it will never apply cleanly to master. We should likely create 1.3.x-stable branch, to PR against.

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@ module.exports = {
         // with the native Intl API
         trees.push(new Funnel(intlPath + '/locale-data/jsonp', {
             srcDir:  '/',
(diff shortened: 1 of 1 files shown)


jasonmit:
https://github.com/yahoo/ember-intl/pull/195 created a 1.3.1 branch and submitted your fix against it.  Thanks :+1:

jasonmit:
I'll also bring the changes to master

jasonmit:
1.3.1 published

PR #199, opened 2015-08-17
Fixed initializer for ember-canary

mdehoog:
Ember canary minor version has been bumped (2.0.0-x to 2.1.0-beta and 2.2.0-canary), which has broken the ember-intl instance initializer again. Updating the semver `satisfies` string fixes it.

Do we need to think of a way to change `is-modern.js` to support future version bumps of ember-canary?


jasonmit:
We can flip the behavior to isLegacy < 1.130

PR #200, opened 2015-08-17
add formatHtmlMessage to the intl service

null-null-null:
Adds `formatHtmlMessage` to the `intl` service. Not sure if there's a reason this formatter has been excluded until now, but if not, here's a simple PR that'll bring it into the fold.


jasonmit:
LGTM.

I meant to do this but was blocked until recently.  Thanks :+1:

PR #205, opened 2015-08-19
Added `Intl.js.map` to public assets (fixes #204)

### LINKED ISSUE #204: ember-cli prod build with source maps fails with ember-intl
When running `ember build -prod`, Javascript minification is enabled by default. Source maps can be enabled in `ember-cli-build.js` as follows:
```
    sourcemaps: {
      enabled: true,
      extensions: ['js']
    }
```

When running a production build of an app using `ember-intl`, the build fails with:
```
Build failed.
ENOENT, no such file or directory '/ember-app/tmp/broccoli_merge_trees-output_path-HMTQQyCi.tmp/assets/intl/Intl.js.map'
```

This is because the `Intl.js.map` is not included in the `treeForPublic()` function in `index.js`.

mdehoog:
See issue #204

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,7 +158,7 @@ module.exports = {
         }
 
         trees.push(utils.lowercaseTree(new Funnel(path.join(intlPath, 'dist'), {
-            files  : ['Intl.complete.js', 'Intl.js', 'Intl.min.js'],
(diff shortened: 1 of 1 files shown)


jasonmit:
:+1:

PR #217, opened 2015-08-27
format_date formats null dates as empty strings by default

adamacus:
This is for https://github.com/yahoo/ember-intl/issues/212.  I took a crack at this, not sure it's the best way, looking for feedback.

### DIFF
--- a/addon/formatters/-base.js
+++ b/addon/formatters/-base.js
@@ -65,7 +65,7 @@ const FormatBase = Ember.Object.extend({
 });
 
 FormatBase.reopenClass({
-    supportedOptions: ['locale', 'format'],
+    supportedOptions: ['locale', 'format', 'allowEmpty'],
     concatenatedProperties: ['supportedOptions']
 });
 
--- a/addon/formatters/format-date.js
+++ b/addon/formatters/format-date.js
@@ -30,9 +30,7 @@ const FormatDate = Formatter.extend({
         if(value || filteredOpts.allowEmpty === false){
           return this._format(value, filteredOpts);
         }
-        else{
-          return '';
-        }
+        return '';
     }
 });
 
--- a/addon/formatters/format-date.js
+++ b/addon/formatters/format-date.js
@@ -24,9 +24,15 @@ const FormatDate = Formatter.extend({
     }),
 
     format(value, options) {
-        value = new Date(value);
+        value = value == null ? null : new Date(value);
         assertIsDate(value, 'A date or timestamp must be provided to format-date');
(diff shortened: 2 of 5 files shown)


addon/formatters/format-date.js
```diff
@@ -24,9 +24,15 @@ const FormatDate = Formatter.extend({
     }),
 
     format(value, options) {
-        value = new Date(value);
+        value = value == null ? null : new Date(value);
```
jasonmit: We'll want to support undefined, null, and empty string

addon/formatters/format-date.js
```diff
@@ -24,9 +24,15 @@ const FormatDate = Formatter.extend({
...
         assertIsDate(value, 'A date or timestamp must be provided to format-date');
-        return this._format(value, this.filterSupporedOptions(options));
+        var filteredOpts = this.filterSupporedOptions(options);
+        if(value || filteredOpts.allowEmpty === false){
+          return this._format(value, filteredOpts);
+        }
+        else{
```
jasonmit: else not needed

addon/formatters/-base.js
```diff
@@ -65,7 +65,7 @@ const FormatBase = Ember.Object.extend({
 });
 
 FormatBase.reopenClass({
```
jasonmit: I think we should handle this at the helper level and not the formatter.  Basically don't invoke the formatters if value is empty when allowEmpty is truthy

jasonmit: Actually, let me think about this some more.  There are trade offs to both approaches.

adamacus: Yeah I actually tried this approach first but then was worried I didn't really understand how broad of an impact a change here would have.

jasonmit: The only issue is it's going to be used on things that aren't time-based (ie, format-message).  And also not sure how we need to handle when time is used as part of a ICU string.

ie, 'order by {deadline, date, time}'

jasonmit: The more I think about this, the more it becomes evident we move this one level higher to the helper instead.  We can leave the service methods unguarded as the person invoking these methods has control there to prevent the behavior of passing an empty string to the service.

jasonmit:
https://github.com/yahoo/ember-intl/pull/218 I iterated on it, feel free to push the changes in.  Also would be sweet if we can support a global setting via `config/environment`

jasonmit:
Okay, updated my PR to https://github.com/yahoo/ember-intl/pull/219 and did some other house keeping in the PR for the legacy helper.

PR #222, opened 2015-09-01
Pass formats through to formatter in intl service

rlivsey:
The message formatter takes [3 arguments - `value`, `options` and `formats`](https://github.com/yahoo/ember-intl/blob/a3600ac0dc79aa73977d0a33c5d97fdee6d0dd9b/addon/formatters/format-message.js#L21)

The `formatMessage` method on the service [only passes through `value` and `options`](https://github.com/yahoo/ember-intl/blob/a3600ac0dc79aa73977d0a33c5d97fdee6d0dd9b/addon/services/intl.js#L25) which means you can't format dates/times etc when using the service.

This PR passes through all three arguments, defaulting the formats if not specified.

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -11,7 +11,7 @@ import Translation from '../models/translation';
 const { assert, get, set, RSVP, Service, Evented, Logger:logger } = Ember;
 
 function formatterProxy (formatType) {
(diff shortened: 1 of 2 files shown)


jasonmit:
Awesome!  Thanks :+1:

PR #227, opened 2015-09-11
load translation sub-directories

null-null-null:
For https://github.com/yahoo/ember-intl/issues/224. Will merge translations from top-level and/or sub-directories into a single file.

In the process of writing this PR, I also encountered a problem when using vim. Vim creates .swp files when editing, which `ember-intl` would attempt to read and then throw the error: `Unable to parse .en-us.yaml.swp`. To stop this behavior, I removed [this line](https://github.com/yahoo/ember-intl/pull/227/files#diff-42ac393812b7b68b3df21fae0ea6b645L66), then threw in a console warning [here](https://github.com/yahoo/ember-intl/pull/227/files#diff-42ac393812b7b68b3df21fae0ea6b645R91).

Also, I was unsure where to put the tests for this PR, so I threw it in with the other acceptance tests in `smoke-test`. If you've got a better idea, let me know and I'll give it a shot.


jasonmit:
Great work, and I'm fine with the warning versus a throw :+1:   Acceptance tests seems like the best place.

jasonmit:
Just a thought, maybe we should warn on collisions?  Not going to block the PR on it, but might be nice UX.

PR #245, opened 2015-10-28
Add `t` method to provide formatted message translation by key

elwayman02:
https://github.com/yahoo/ember-intl/issues/243

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -52,6 +52,11 @@ const IntlService = Service.extend(Evented, {
   formatTime: formatterProxy('time'),
   formatDate: formatterProxy('date'),
 
+  t(key, options, formats) {
+    const translation = this.findTranslationByKey(key, options && options.locale);
+    return this.formatMessage(translation, options, formats);
+  },
+
   adapter: computed({
     get() {
       return this.container.lookup('ember-intl@adapter:-intl-adapter');
--- a/tests/unit/services/intl.js
+++ b/tests/unit/services/intl.js
@@ -1,0 +1,75 @@
+import { moduleFor } from 'ember-qunit';
+import test from 'dummy/tests/ember-sinon-qunit/test';
+
+let service;
+
+moduleFor('service:intl', 'Unit | Service | intl', {
+  setup() {
+    service = this.subject();
+  }
+});
+
+let message, translation, key, result, formatStub, findStub, formatArgs, findArgs;
+
+function setupT(options, formats) {
+  message = 'foobar';
+  translation = 'foo';
+  key = 'foo.bar';
+  formatStub = this.stub(service, 'formatMessage', function () {
+    return message;
+  });
+  findStub = this.stub(service, 'findTranslationByKey', function () {
+    return translation;
+  });
+
+  result = service.t(key, options, formats);
+
+  formatArgs = formatStub.firstCall.args;
+  findArgs = findStub.firstCall.args;
+}
+
+test('t', function (assert) {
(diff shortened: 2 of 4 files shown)


addon/services/intl.js
```diff
@@ -52,6 +52,10 @@ const IntlService = Service.extend(Evented, {
   formatTime: formatterProxy('time'),
   formatDate: formatterProxy('date'),
 
+  t(...args) {
+    return this.formatMessage(this.findTranslationByKey(...args));
```
jasonmit: formatMessage takes 3 arguments: the translation, the value hash, and the format hash

addon/services/intl.js
```diff
@@ -52,6 +52,10 @@ const IntlService = Service.extend(Evented, {
   formatTime: formatterProxy('time'),
   formatDate: formatterProxy('date'),
 
+  t(...args) {
```
jasonmit: Can you cover with some unit tests

elwayman02: :+1:

addon/services/intl.js
elwayman02: What is the current signature of the `{{t}}` helper? I'd like to maintain the same arguments that it supports.

jasonmit: `{{t 'foo' bar='bar' locale='fr'}}`

The formats gets threaded through automatically.  But when invoking it programmatically, you have the ability to thread a format POJO.  Ultimately, this gets invoked: https://github.com/yahoo/ember-intl/blob/master/addon/formatters/format-message.js#L21 and that all should be satisfied

addon/services/intl.js
```diff
@@ -52,6 +52,10 @@ const IntlService = Service.extend(Evented, {
   formatTime: formatterProxy('time'),
   formatDate: formatterProxy('date'),
 
+  t(key, options, formats) {
+    return this.formatMessage(this.findTranslationByKey(key, options.locale), options, formats);
```
jasonmit: will throw if options is undefined on options.locale.  it's optional

addon/services/intl.js
```diff
@@ -52,6 +52,10 @@ const IntlService = Service.extend(Evented, {
   formatTime: formatterProxy('time'),
   formatDate: formatterProxy('date'),
 
+  t(key, options, formats) {
+    return this.formatMessage(this.findTranslationByKey(key, options ? options.locale : undefined), options, formats);
```
jasonmit:
```js
let translation = this.findTranslationByKey(key, options && options.locale);

return this.formatMessage(translation, options, formats);
```

I would rather split this up for debugging purposes

elwayman02: ok

tests/unit/services/intl.js
```diff
@@ -0,0 +1,75 @@
+import { moduleFor } from 'ember-qunit';
+import test from 'dummy/tests/ember-sinon-qunit/test';
+
+let service;
+
+moduleFor('service:intl', 'Unit | Service | intl', {
```
jasonmit: I'd rather see more of an integration style test to get the whole flow.  I'm fine with keeping these unit tests though.

elwayman02: I agree, but I don't have time to address that today. Is this sufficient to merge the changes and we can log an issue to provide better overall unit/integration tests for the service?

jasonmit: Okay, will hold off on versioning though.

elwayman02: :+1:

PR #254, opened 2015-11-05
Don't lowercase the sourcemap file

elucid:
Doing so causes an error when building with souremaps enabled in production on case-sensitive file systems.

I came across this error https://github.com/yahoo/ember-intl/issues/204 when using a version of ember-intl that included the fix. You can see a simple sample application here https://github.com/elucid/ember-intl-test that demonstrates the issue is still present when building on a case-sensitive filesystem like ext4.

As the issue mentions, Intl.js.map must be included in `treeForPublic()` but it is only picked when run on case-insensitive filesystems. This moves Intl.js.map out into its own filter which leaves the case intact. I've tested this on both case-insensitive and case-sensitive filesystems.

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,9 +131,15 @@ module.exports = {
       trees.push(publicTree);
     }
 
+    trees.push(new Funnel(this.trees.intl, {
+      srcDir: 'dist',
+      files: ['Intl.js.map'],
+      destDir: path.join(assetPath)
+    }));
+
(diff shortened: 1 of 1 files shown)


elucid:
NOTE: The suite is passing on everything but canary. I'm not sure this is related to my change.

jasonmit:
Thanks, yes the canary break is unrelated.

PR #265, opened 2015-11-29
makes default locale optional

jasonmit:
* This PR brings back support for defining an array for your application locale.  This array is used to resolve a translation key for the fallback case.

```
this.get('intl').setLocale([primary, secondary, fallback]);
```

* This adds a missing-message utility, similar to that of ember-i18n
* This makes default locale optional, as it would interfere with application that rely entirely on side loading translations

### DIFF
--- a/addon/helpers/-base.js
+++ b/addon/helpers/-base.js
@@ -28,14 +28,13 @@ function helperFactory(formatType, optionalGetValue, optionalReturnEmpty) {
 
     init(...args) {
       this._super(...args);
-      let intl = get(this, 'intl');
-      intl.on('localeChanged', this, this.recompute);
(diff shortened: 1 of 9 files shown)


addon/helpers/-base.js
```diff
@@ -45,6 +44,12 @@ function helperFactory(formatType, optionalGetValue, optionalReturnEmpty) {
         throw new Error(`format-${formatType} helper requires value`);
       }
 
+      let locale = get(intl, '_locale');
+
+      if (Array.isArray(locale)) {
+        locale = locale[0];
```
jasonmit: Use the primary locale always, the only time we use the fallbacks is when resolving translations.

PR #284, opened 2016-02-01
Auto inject polyfill into the head of index.html

jasonmit:
Not merging this for now since it's a breaking change, might need to hold off until 3.0.

Another option is to have an interactive installer that has this option off by default but suggests it and programmatically constructs the configuration object.

Begins to work needed for #324

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -57,6 +57,32 @@ module.exports = {
     return this.app;
   },
 
+  outputPaths: function() {
+    var assetPath = 'assets/intl';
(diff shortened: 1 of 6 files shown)


packages/ember-intl/index.js
```diff
@@ -57,6 +57,32 @@ module.exports = {
     return this.app;
   },
 
+  outputPaths: function() {
+    var assetPath = 'assets/intl';
```
jasonmit: `assets` should come from some whatever the the path that treeForPublic is written to, I believe there is an option where this can be configurable so we don't want to make assets static.

PR #297, opened 2016-03-15
Blueprint to override default translation model to support dot notate…

### LINKED ISSUE #296: Add support for flat json structure
The current implementation does not seem to allow for flat-style keys.

Example:
```
{
    "login.authenticate": "Authenticate",
    "login.forgot-password": "Forgot Password"
}
```

It would be nice if ember-intl could detect this flat structure and parse them into something usable.
Currently, using a flat structure as per above causes Ember throw an error for not finding the parent key when it attempts to break it down:
```
Error: Property set failed: object in path "login" could not be found or was destroyed.
```

jasonmit:
…d translation keys

Fixes #296

### DIFF
--- a/packages/ember-intl/addon/adapters/default.js
+++ b/packages/ember-intl/addon/adapters/default.js
@@ -5,7 +5,6 @@
 
 import Ember from 'ember';
 import getOwner from 'ember-getowner-polyfill';
-
 import Translation from '../models/translation';
 
 const { assert } = Ember;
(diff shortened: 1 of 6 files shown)


packages/ember-intl/addon/adapters/default.js
```diff
@@ -28,6 +26,7 @@ const DefaultIntlAdapter = Ember.Object.extend({
 
     const owner = getOwner(this);
     const lookupName = `ember-intl@translation:${normalize(localeName)}`;
+    const Translation = owner._lookupFactory('model:ember-intl-translation') || owner._lookupFactory('ember-intl@model:translation');
```
jasonmit: Tries to resolve from the app namespace first

PR #299, opened 2016-03-17
Creates a standalone config file

jasonmit:


packages/ember-intl/blueprints/ember-intl-config/files/config/ember-intl.js
```diff
@@ -0,0 +1,12 @@
+/*jshint node:true*/
+
+module.exports = function(environment) {
```
jasonmit: Reminder to self: document each setting in the blueprint

Issue #271, opened 2015-12-08
Can't be used as a dependency in an addon


ajgordo:
We have an addon where we have ember-intl included. We have found that unless a project\application adds ember-intl as a dependency then our addon will fail.
If we have ember-intl as dependency then and add to index.js of our addon:
```js
  included: function(app) {
    this._super.included(app);
    .
    .
    .
```
 we get errors the error `Cannot read property 'translations' of undefined`. It seems the only option right now is to include ember-intl in to the application project.

jasonmit:
Which version of ember-intl are you targeting?  Where you intending to keep the translations scoped to the consuming app or would the best shipped with your addon?

ajgordo:
Ideally we would like to provide the translations within our addon, but allow others to override them if they need to. We haven't fully explored that yet (priorities), so for those that want to use our addon we are providing a list of translations to add to their application.
I had also considered creating the translations files as part of a blueprint in our addon, but that becomes hard to maintain.

jasonmit:
Which version of ember-intl are you on right now?  It's likely possible if you take the fully programmatic approach of loading translations into the service.

ajgordo:
We are currently using 2.0.0-rc.5. I'll have to take a look at using a service when I get back around to that. Thanks for the suggestion on that one.

jasonmit:
Also, the issues around the translation folder being required were solved rc.7.

xdumaine:
@jasonmit I'm trying to understand the issue as described here. My own sounds the same: I'd like to be able to provide translations within the addon itself. Is this something that should be supported?

jasonmit:
@xdumaine not currently but it's planned for 3.0 where you will create a translations file in your addon root and the consuming app will bundle them

https://github.com/yahoo/ember-intl/issues/285

vlascik:
This is a problem also when trying to use ember-intl with engines, `Cannot read property 'translations' of undefined` error can be fixed with changing index.js to something like: 
```
  included: function(app) {
    if (app.app) {
      app = app.app;
    }
    this.app = app;
```
but other things fail then anyway.

jasonmit:
https://github.com/yahoo/ember-intl/pull/307 is adding support for this.  I just need to work through one issue and then I will publish.

xdumaine:
@jasonmit awesome! Psyched that you made time for this, thanks! We'll take a look. Do you have an example or docs for how to set it up properly, or should it just work™?

[...]

PR #307, opened 2016-03-28
initial spike for addon support

jasonmit:
Adds first class support for addons to supply translations up to the parent application.

`<addon root>/translations/[en-us.yml, fr-fr.yml, ...]`

Todo:
* [x] If the parent application defies a translation key that collides, the parent should win
* [x] allow addons to provide a translation path if they want to store them somewhere other than `<addon>/translations` via package.json configuration.
```
{
  "name": "intl-addon",
  ...
  "ember-addon": {
    "configPath": "tests/dummy/config",
    "translationPath": "x-translations"
  }
}
```

#271


packages/ember-intl/index.js
```diff
@@ -37,17 +37,30 @@ module.exports = {
     }
 
     this.app = app;
+    this._intlConfig = this.intlConfig(this.app.env);
+    var translationFolder = this._intlConfig.inputPath;
 
-    this._intlConfig = this.intlConfig(app.env);
-    this.hasTranslationDir = existsSync(this.project.root + '/' + this._intlConfig.inputPath);
+    this.hasTranslationDir = existsSync(this.project.root + '/' + translationFolder);
```
jasonmit: hasTranslationDir needs to be removed as part of this

PR #314, opened 2016-03-30
Translations should live in a nested object on the model

jasonmit:


packages/ember-intl/addon/models/translation.js
```diff
@@ -5,25 +5,31 @@
...
 
-const { get, set } = Ember;
+const { get, set, computed } = Ember;
 
 const TranslationModel = Ember.Object.extend({
+  init() {
+    this._super(...arguments);
+
+    if (!this.translations) {
+      this.translations = {};
```
jasonmit: debating using a weakmap here to ensure this object remains private

PR #316, opened 2016-04-01
Pick only exact matched locale for polyfill

### LINKED ISSUE #315: Polyfill builds too many locales
We have a few translations defined in "translations" directory, they all use the "short" version: "en", "de".
This causes very long builds, as Brocolli processes every file that maches those strings by regexp.

With `disablePolyfill: false`:  1m41.592s
With `disablePolyfill: true`: 0m31.455s

Problematic line: https://github.com/jasonmit/ember-intl/blob/0c36b5b91e8d81e5a5a23aa03b2db8afd25f57f9/packages/ember-intl/index.js#L132

It causes to match (and build) files like "en-tv", "en-ws" or even "kde-tz". We deifinitely don't need all those variants as we activate only those translations which we have defined and we never activate any of the subvariants.

I know I could use the polyfill from CDN, but I hoped that we could get the build mechanism working as expected as well.

Edit:
Part of detailed debug output from `DEBUG=broccoli* ember build -prod`
```
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/de-lu-405086ee17e88ad5f80c18d36e99c404.js +196ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/dsb-de-f153bdfb64b2d533e7f6580816d48848.js +197ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-001-9ee3517647c34a880aebfd68f466e3ab.js +197ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-150-55f454a71a5d71933f0d5b503e8e9079.js +191ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-36656d5d0b30ee32b78ecff4fd0f3b98.js +191ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-ag-9b35d29388bcf26b380409b8c19b583b.js +192ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-ai-c3815d1bea2f2cf9e53208990132dede.js +192ms
  broccoli-filter:AssetRewrite cache prime: assets/intl/locales/en-as-31425900b3092790e4743d57c1e00c07.js +197ms
```

jgoclawski:
This fixes #315.

I'll be able to run more test on Monday, but it should be ok.
Is there any other usage I'm not aware of that could be broken by this approach?

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -129,7 +129,7 @@ module.exports = {
 
       if (this.projectLocales.length) {
         localeFunnel.include = this.projectLocales.map(function(locale) {
-          return new RegExp(locale, 'i');
+          return new RegExp('^' + locale + '.js$', 'i');
         });
       }
 


jasonmit:
@jgoclawski no issues with this, thanks for reporting and also submitting the patch.

I will version shortly.

Issue #268, opened 2015-12-01
Support sideloading of CLDR data


mjuniper:
For apps that need to support many languages, sideloading the CLDR data would be great.

buschtoens:
In order to support this, we essentially just need to do what we already do with the regular translations and `this.opts.publicOnly`:

https://github.com/ember-intl/ember-intl/blob/c92422d416de65db3fb70f68c57fec9b782dc40c/index.js#L93-L144

Then defuse the warning here:

https://github.com/ember-intl/ember-intl/blob/c92422d416de65db3fb70f68c57fec9b782dc40c/addon/hydrate.js#L13-L43

And document for users that they now not only need to call `addTranslations`, but also `addLocaleData`:

https://github.com/ember-intl/ember-intl/blob/c92422d416de65db3fb70f68c57fec9b782dc40c/addon/services/intl.js#L161-L173

We can't repurpose the `publicOnly` option, as this would be a breaking change for users of it. Any suggestions for a new option name?

PR #317, opened 2016-04-01
Write a CLDR module per lang instead of per locale

jasonmit:
This is needed to begin addressing #268

### DIFF
--- a/packages/ember-intl/lib/broccoli/cldr-writer.js
+++ b/packages/ember-intl/lib/broccoli/cldr-writer.js
@@ -7,11 +7,137 @@
 * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
 */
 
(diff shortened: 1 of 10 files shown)


packages/ember-intl/lib/broccoli/cldr-writer.js
```diff
@@ -39,19 +42,33 @@ Plugin.prototype.build = function() {
...
+    var langData = map.get(lang) || [];
 
-  Object.keys(localesData).forEach(function(locale) {
-    var result = localesData[locale];
+    return map.set(lang, langData.concat(data));
+  }, new Map());
+
+  mkdirp.sync(destPath);
 
+  cldrDataByLang.forEach((cldrData, lang) => {
```
jasonmit: This is ripped from react-intl build script

PR #331, opened 2016-04-20
notifyPropertyChange on locale to bust cache for observers

### LINKED ISSUE #330: Document localeChanged
See: http://discuss.emberjs.com/t/trouble-understanding-how-observers-work/10487/2

Also, we likely only want to trigger localeChanged when the locale on did change.

```
intl.setLocale('en');
intl.setLocale('en');
```

Should only trigger the event once, right now it would trigger it twice before we're not comparing previous and present values.

jasonmit:
Fixes #330


packages/ember-intl/tests/unit/services/intl.js
```diff
@@ -1,75 +0,0 @@
...
-
-moduleFor('service:intl', 'Unit | Service | intl', {
-  setup() {
-    service = this.subject();
-  }
-});
-
-let message, translation, key, result, formatStub, findStub, formatArgs, findArgs;
-
-function setupT(options, formats) {
```
jasonmit: tests were in dev awhile back but were never complete, now they're obsolete from 1.x and I need to rewrite them now that services/intl is running in the tests

PR #333, opened 2016-04-21
Fixes options locale bleeding out into the options hash passed into `service.t`

jasonmit:
Fixes options locale bleeding out into userland by mutating the options hash passed into `service.t`

See test to see the issue and the fix.

### DIFF
--- a/packages/ember-intl/addon/helpers/-format-base.js
+++ b/packages/ember-intl/addon/helpers/-format-base.js
@@ -45,10 +45,14 @@ function helperFactory(formatType, helperOptions = {}) {
       }
 
       const intl = get(this, 'intl');
-      let locale = get(intl, '_locale');
+      let locale = hash.locale;
 
-      if (Array.isArray(locale)) {
-        locale = locale[0];
+      if (!locale) {
+        locale = get(intl, '_locale');
+
+        if (Array.isArray(locale)) {
+          locale = locale[0];
+        }
       }
 
       let format = {};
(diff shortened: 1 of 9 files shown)


packages/ember-intl/addon/helpers/-format-base.js
```diff
@@ -60,8 +63,11 @@ function helperFactory(formatType, helperOptions = {}) {
 
       return get(this, 'formatter').format(
         value,
-        extend({ locale }, format, hash),
-        get(intl, 'formats')
+        assign(assign({}, format), hash),
```
jasonmit: since Ember.merge doesn't take N option args

packages/ember-intl/addon/helpers/-format-base.js
```diff
@@ -46,10 +45,14 @@ function helperFactory(formatType, helperOptions = {}) {
...
       const intl = get(this, 'intl');
-      let locale = get(intl, '_locale');
+      let locale = hash.locale;
+
+      if (!locale) {
+        locale = get(intl, '_locale');
 
-      if (Array.isArray(locale)) {
-        locale = locale[0];
+        if (Array.isArray(locale)) {
```
jasonmit: This now appears unnecessary, I need to validate

jasonmit: This is another bug, I'm going to fix this as well.  It should accept an array.

https://github.com/yahoo/intl-messageformat#intlmessageformat-constructor

PR #339, opened 2016-04-25
Added support for translations beyond the first layer of addons.

JackieChiles:
Implements support for nested addons by recursively looking for translations in each addon referenced by an addon.


jasonmit:
Looks good, thanks!  It will be part of 2.11.0 which will be versioned some point this week.

JackieChiles:
Excellent, thank you!

PR #345, opened 2016-05-04
Attempts to try to set bare-strings to true in template-lintrc

jasonmit:
/cc @rwjblue

https://github.com/rwjblue/ember-template-lint/pull/27

### DIFF
--- a/packages/ember-intl/blueprints/ember-intl/index.js
+++ b/packages/ember-intl/blueprints/ember-intl/index.js
@@ -1,5 +1,8 @@
 'use strict';
 
+var path = require('path');
+var existsSync = require('exists-sync');
+
 /**
  * Copyright 2015, Yahoo! Inc.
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

rwjblue:
Looks good! I think I'd still like a flag on the addon so that I can warn in ember-cli-template-lint when the rule isn't set explicitly (either true or false) in the config file.

rwjblue:
After thinking about it some more, maybe ember-cli-template-lint should do this in its blueprint instead of both ember-intl and ember-i18n both doing it?  In ember-cli-template-lint I would check for the flag discussed above (and in the ember-template-lint issue) and generate the default blueprint with `bare-strings` enabled.  Then each time we go to do a lint run, I could also check and confirm that `bare-strings` is present in the rules list...

rwjblue:
What do you think?

jasonmit:
Fully on board with the flag on the addon.  Will close this and wait for that to land.

PR #353, opened 2016-06-07
Fixing config lookup logic for ember >= 2.6.0 beta.3

### LINKED ISSUE #352: Intl config not being read
At least on Windows (haven't tested elsewhere), in `readConfig` function in `index.js` the config path is wrong
```
var configPath = path.join(project.root, project.configPath(), '..', 'ember-intl.js');
```
leads to paths like `C:\Project\C:\Project\config\environment\..\ember-intl.js`, as `project.configPath()` returns the full path (in 2.6.0). Intl config file is then not found and read.
Changing the line to
```
var configPath = path.join(project.configPath(), '..', 'ember-intl.js');
```
fixes the problem.

ember-intl: 2.11.1
ember: 2.6.0.beta.4
ember-cli: 2.6.0-beta.3
node: 6.2.1
os: win32 x64

jasonmit:
Fixes #352

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -144,10 +144,14 @@ module.exports = {
 
   readConfig: function(environment) {
     var project = this.app.project;
-    var configPath = path.join(project.root, project.configPath(), '..', 'ember-intl.js');
 
-    if (fs.existsSync(configPath)) {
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

PR #360, opened 2016-06-28
Fix #359 Safari's Can't find variable: Intl error on startup

### LINKED ISSUE #359: Safari - `Intl is not defined` error causes apllication to abort startup
[The instance-initilizer module checks for `Intl` variable](https://github.com/jasonmit/ember-intl/blob/master/packages/ember-intl/app/instance-initializers/ember-intl.js#L29). Only in Safari, this results in following error during application startup causing it to fail

    ReferenceError: Can't find variable: Intl
        didBecomeReadyember.debug.js:4874
        invokeember.debug.js:332
        flushember.debug.js:396
        flushember.debug.js:204
        endember.debug.js:559
        ...

Weird that the same project works in Firefox and Chrome without complaining about Intl not being there. Fixing this would atleast let the application run without il8n support when Intl is not present.

kashiif:
Fixes Issue #359

### DIFF
--- a/packages/ember-intl/app/instance-initializers/ember-intl.js
+++ b/packages/ember-intl/app/instance-initializers/ember-intl.js
@@ -26,7 +26,7 @@ export function instanceInitializer(instance) {
     service = instance.container.lookup('service:intl');
   }
 
-  if (!Intl) {
+  if (typeof Intl === 'undefined') {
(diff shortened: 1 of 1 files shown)


yahoocla:

jasonmit:
Thanks!

PR #368, opened 2016-07-06
Fix: show url in console message when browser lacks Intl Api

kashiif:


yahoocla:
CLA is valid!

jasonmit:
Thanks, will version tonight

PR #370, opened 2016-08-07
respect the new rootURL behavior

### LINKED ISSUE #369: Incompatibility with Ember 2.7.0 due to relative paths
Since Ember deprecated the `baseURL` option and removed the `<base>` tag entirely (http://emberjs.com/blog/2016/04/28/baseURL.html) this addon stopped working for me.

I tested that this is an issue by

- create a new Ember 2.7.0 app
- `ember install ember-intl`
- enable the polyfill `autoPolyfill: true,` in `config/ember-intl.js`
- start up the app
- create a route that's at least two slashes deep (i.e. `/name/:id`)
- navigate to that route and observe that the asset files are referenced without a leading slash leading the browser to attempt to retrieve them from something like `http://localhost:4200/name/:id/assets/intl/intl.min.js`.

I was able to work around this issue by overriding `app.intl` in `ember-cli-build.js` but I only guessed to do this after reading the source code and seeing that the plugin looks for this value then reading documentation about creating ember-cli addons and where things like their `included` gets its arguments. I suggest that the documentation be improved to mention this option on this page https://github.com/jasonmit/ember-intl/wiki/Intl.js-Polyfill and that the page be updated to not display relative paths by default once this is fixed.

jasonmit:
fixes #369

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,7 +300,7 @@ Phantom does support the Intl API, so in order for for you ember-intl to work in
 
 To resolve this, add the following above all script tags in `tests/index.html`:
 ```html
-<script src="assets/intl/intl.complete.js"></script>
+<script src="{{rootURL}}assets/intl/intl.complete.js"></script>
 ```
 
 ## Asynchronously loading translations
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

PR #377, opened 2016-10-03
Format-block component

taheilo:
**What have changed**
- Added a component `format-block`. Takes `t` and `locale` as params.
  - `t` is required
  - `locale` is optional

**Why**
- Enabled us to build more complex translations, and use helpers and components in our translations.

This is literally my first time working with add-ons in Ember, so I apologies if this is not the way to do it. Let me know what your thoughts are.

### DIFF
--- a/packages/ember-intl/addon/components/format-block.js
+++ b/packages/ember-intl/addon/components/format-block.js
@@ -1,0 +1,33 @@
+import Ember from 'ember';
+import links from '../utils/links';
+
+const { inject } = Ember;
+
+export default Ember.Component.extend({
+  tagName: 'span',
+
+  intl: inject.service(),
+
+  init: Ember.on('init', function() {
+    this._super(...arguments);
+
+    let key = this.get('t');
+    let locale = this.get('locale') || this.get('intl._locale');
+
+    if (!key) {
+      throw new Error(
+        `No translation specified.`
+      );
+    }
+
+    if (!locale) {
+      throw new Error(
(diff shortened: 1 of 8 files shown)


yahoocla:

jasonmit:
Paging @taheilo

taheilo:
Sorry about the late response.

I have been looking into the failing test `not ok 46 PhantomJS 1.9 - component:format-block: should throw error if no key` but it passes for me locally. Do you have any pointers as to what I might be doing wrong?

The request from yahoocla has already been done.

jasonmit:
Left a question above since I don't think I understand the use case.

taheilo:
This is the scenario I needed it for:
I have a `post` with a bunch of editors. The translation for this would in english be:
`Edited by Paul, Brian, Brad +4` with the `+4` indicating that there are 4 more. The +4 is wrapped in a tooltip component so they show up on hover. It could however be pretty much anything you put in the yield.

jasonmit:
@taheilo I need to think more about this.  I'm actually in the process of moving around the internals and would like to revisit once that is complete.  Should be within the next two-weeks.

The gist is I plan to make the message compiler swappable for others that do not support the ICU message syntax for translations.  And in doing so, will make https://github.com/jasonmit/ember-intl/pull/377/files#diff-fa056eb2fd4f3b51ff7bd33831c63070R30 not compatible with this approach since that could interfere with message syntax for non-ICU.

Issue #350, opened 2016-05-31
custom translation compilers


chrisjshull:
I'm comparing ember-intl with ember-18n and don't see docs here about custom translation compliers:
https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translation-Compiler

Is this possible with ember-intl?

jasonmit:
Not currently, but have considered implementing this.

I will leave the issue opened for tracking.

jasonmit:
I plan to implement a first pass at this over the next week.

taheilo:
+1

jasonmit:
I meant to implement this over the weekend but got side tracked, it's still high on my list of things-to-do and will likely be the next feature of ember-intl worked on.

It will move the formatter message into an optional addon `ember-intl-messageformat` that the default blueprint will install.  I'll also create a blueprint for the ruby-style translations that will uninstall `ember-intl-messageformat` as part of the blueprint and drop in a new compiler that by default will support the following syntax: `{{count}} Clicks`

If you want to implement support for any other type of syntax you would need to just override the format method of the formatter it drops in the app namespace.

PR #378, opened 2016-10-11
Allow for formatters to be looked up against the app namespace

jasonmit:
Small change to begin #350, next PR will involve ensuring ember-intl@formatters/base can be public


yahoocla:
CLA is valid!

PR #379, opened 2016-10-11
Rename format method to compute

jasonmit:
Also related to #350 
* centralizes the lookupFormatter logic to the service
* renamed `format` to `compute` on the formatter base to get ready to make it public
* reduce `ember-intl/formatters/base` down to just the _current_ public interface


yahoocla:
CLA is valid!

PR #380, opened 2016-10-12
Move message formatter into ember-intl-messageformat

jasonmit:
Moving out ember-intl-messageformat so that there is no cost to bringing in the parsers and what not when you opt out of ember-intl-messageformat with a custom compiler #350 

The conditional logic to strip out ember-intl-messageformat will follow soon.


yahoocla:
CLA is valid!

PR #383, opened 2016-10-12
Move the locale data loader to ember-intl-messageformat

jasonmit:
More decoupling of Intl.MessageFormat into ember-intl-messageformat addon #350


yahoocla:
CLA is valid!

PR #384, opened 2016-10-12
Combine messageformat and relativeformat into ember-intl-icu-bundle

jasonmit:
Combine the two messageformat and relative format into `ember-intl-icu-bundle` since they're dependent on each other.  Opting out of messageformat will then opt out of both these formatters and helpers.

#350


yahoocla:
CLA is valid!

PR #387, opened 2016-10-13
Add translation synchrnous, expose add as the public API for adding t…

jasonmit:
…ranslations on the locale model


yahoocla:
CLA is valid!

PR #388, opened 2016-10-13
Initial commit

jasonmit:
The goal of this addon will be to serve as an example of how to swap out the compiler while also leveraging the CLDR data for pluralization.


yahoocla:
CLA is valid!

PR #389, opened 2016-10-14
WIP - 3.0.0-alpha.1

jasonmit:
- [x] moved all icu message dependent code into ember-intl-icu-bundle
- [x] addTranslation(s) are synchronous, no longer return a promise
- [x] combined ember-intl-relativeformat and ember-intl-messageformat into ember-intl-icu-bundle
- [ ] create optional addon ember-intl-ember-i18n-bundle which swaps out all ICU logic with a compiler that supports ember-i18n's translation format
  - [ ] pluralization functions will be provided by `make-plural` at build time using a broccoli plugin to push into the addon tree


yahoocla:
CLA is valid!

PR #393, opened 2016-10-28
Port of ember-i18ns macro to ember-intl

### LINKED ISSUE #392: Computed property macros
`ember-i18n` has a `translationMacro`, which essentially generates a computed property listening to `i18n.locale` (see https://github.com/jamesarosen/ember-i18n/blob/master/addon/utils/macro.js). Would something like that be desirable in `ember-intl` too?

jasonmit:
Given how similar the APIs are, it was pretty much a 1-1 copy.

Closes #392


yahoocla:
CLA is valid!

PR #397, opened 2016-11-02
[WIP] Computed Property Macros

### LINKED ISSUE #392: Computed property macros

Turbo87:
This PR introduces a `formatNumber()` computed property macro as discussed in https://github.com/jasonmit/ember-intl/issues/392#issuecomment-257812181. This is marked WIP since I'd like to add more macros, but also because the discussion in #392 wasn't quite finished and I wanted to present a concrete proposal to discuss.

Resolves #392

### DIFF
--- a/packages/ember-intl/addon/macros.js
+++ b/packages/ember-intl/addon/macros.js
@@ -1,0 +1,57 @@
+import Ember from 'ember';
+
+const { get, assert, computed } = Ember;
+const keys = Object.keys;
+
+function extractDependencies(obj) {
+  return keys(obj)
+    .map(key => obj[key])
+    .filter(value => value.isReference)
+    .map(it => it.key);
+}
+
+function resolve(context, value) {
+  return value.isReference ? get(context, value.key) : value;
+}
+
+function resolveHash(context, hash) {
+  let result = {};
+
+  keys(hash).forEach(function(key) {
+    result[key] = resolve(context, hash[key]);
+  });
+
+  return result;
+}
+
+export function raw(value) {
+  return { value, isRaw: true };
+}
+
+export function ref(key) {
+  return { key, isReference: true };
+}
+
+/**
+ * Creates a computed property that calls `intl.formatNumber()`
+ *
+ * @param ref name of another property on the object (or a number wrapped with `raw()`)
+ * @param options static options hash (use `ref()` to reference other properties on the object)
+ */
+export function formatNumber(ref, options) {
+  let hash = options || Object.create(null);
+  let dependentKeys = extractDependencies(hash);
+  if (!ref.isRaw) {
+    dependentKeys.push(ref);
+  }
+  dependentKeys.push('intl.locale');
+
+  return computed(...dependentKeys, function() {
+    let intl = get(this, 'intl');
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

jasonmit:
Thanks for taking the time to land this :)

LGTM -- just need to address failing tests.  Ping me on slack if you want to chat.

I've also made you a contributor as I desperately need help managing this project (do so as you feel the need to obviously).

Turbo87:
One project more or less to maintain doesn't even matter anymore... :joy:

packages/ember-intl/tests/unit/macros/format-number-test.js
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach() {
+    let intl = this.subject();
+    intl.setLocale('en');
+
+    this.object = Ember.Object.extend({
+      intl,
+      someNumber: 12345,
+      fractionDigits: 3,
+      formattedNumber: formatNumber('someNumber'),
+      rawFormattedNumber: formatNumber(raw(10001)),
```
marcoow: Is `raw` actually needed at all when there's also `ref`? It seems it would be easier to simply expect everything to be a value unless explicitly marked as a reference via `ref`?

Turbo87: yeah, that is the one thing I'm not sure about yet. there are macros where it's a lot more common to use references over static values, at least for the positional parameters, and forcing to use `ref` there might clutter the code quite a bit. I'm 50/50 on this...

kellyselden:
FWIW, I'm planning on extracting the primitives from ember-awesome-macros (`raw`, `computed` (composable version), any others). Not saying you should block on this, but it should be done soon.

packages/ember-intl/addon/macros.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  return result;
+}
+
+export function raw(value) {
+  return { value, isRaw: true };
+}
+
+export function ref(key) {
```
jasonmit: I think we should have only `raw`.  Not using a wrapper, when providing a string, should default to `ref`.  != String should default to `raw`.

jasonmit:
Can I do anything to move this along?  Happy to take it over if it's a time thing.

Turbo87:
> Can I do anything to move this along? Happy to take it over if it's a time thing.

It's unfortunately not a time thing, I just wasn't sure anymore what the right path would be...

jasonmit:
I liked the direction.  The only things I'd ask is to drop `ref` and always default it.  We should also only have to wrap string literals in `raw`.  If the value type is not a string than it's safe to treat it as raw.

(There might be a use case to keep `ref` around, just unsure what it is.)

Turbo87:
yeah, though I'm not a huge fan of that (see the discussion in https://github.com/jasonmit/ember-intl/issues/392), but ultimately it's your decision.

jasonmit:
Consider me persuaded.  I see your point in it helping to reduce confusion

PR #398, opened 2016-11-02
Import wiki pages into "docs" folder

### LINKED ISSUE #396: Move documentation into repository
With a v3.0 planned (#485) it might make sense to move the documentation into the repository to version it properly. That way we can more easily view the documentation corresponding to a particular release. Please let me know if a PR that does that would be acceptable.

Turbo87:
Resolves #396

### DIFF
--- a/docs/addon-support.md
+++ b/docs/addon-support.md
@@ -1,0 +1,12 @@
+
+Addon support
+==============================================================================
+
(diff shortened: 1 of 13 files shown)


yahoocla:
CLA is valid!

jasonmit:
This is great, thanks!

I'll likely just point the wiki to the docs folder and remove the current docs that are there.

PR #404, opened 2016-11-28
wiki links were removed from docs and codes

ykaragol:
Wiki links are removed and wiki "is now part of the repository https://github.com/jasonmit/ember-intl/tree/master/docs."

So I've removed wiki urls from readme, docs and all codes. Please do check before merging it.


Turbo87:
shit, totally forgot about those links in #398 ... thanks @ykaragol!

PR #405, opened 2016-11-30
[DNM] flagging mechanism for fallback translations

MarkhamShofner:
PR for conversational purposes - 

Our team needed a way to indicate when translations resort to fallback strings (specifically for us - there is always a lag in sending/receiving translations, even though we are constantly deploying in between these events. QA has been reporting translation related bugs that will actually resolve themselves once the next round of translations are returned, so we needed a way to indicate that the strings, while missing in locales currently, will no longer be an issues in the near future). 
Could potentially be turned into something more general for other teams to find useful.

Photo example of the pattern. 
image

### DIFF
--- a/packages/ember-intl/addon/adapters/default.js
+++ b/packages/ember-intl/addon/adapters/default.js
@@ -50,12 +50,19 @@ const DefaultIntlAdapter = Ember.Object.extend({
     const len = locales.length;
     let i = 0;
 
+    const chosenLocale = locales[0];
+
     for (; i < len; i++) {
       const locale = locales[i];
       const translations = this.translationsFor(locale);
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@MarkhamShofner doesn't this approach fail when you have a `baseLocale` set in your config?

jasonmit:
@Turbo87 it would.

I also don't feel this would be terribly useful in all cases.  Sometimes it's perfectly fine to fallback w/o any need for indication.

MarkhamShofner:
@jasonmit - agree that fallback indicator isn't always (or even frequently) needed. In this case, it's pretty specific to what our team needed - but it seems like other people could want to flip similar functionality.

Also - thanks for ember-intl!

PR #408, opened 2016-12-14
Add information about `ember-cli-polyfill-io`

alexlafroscia:
Update the docs to mention using `ember-cli-polyfill-io` rather than adding the `<script>` tag to the app directly.


jasonmit:
👍  thanks!

PR #410, opened 2016-12-30
ensure app is set

null-null-null:
I'm getting this error when using `ember-intl` within an addon:

```js
TypeError: Cannot read property 'project' of undefined
    at Class.module.exports.intlConfig (/home/null/code/affinity/stage/directions/text/node_modules/affinity-engine-translator-ember-intl/node_modules/ember-intl/index.js:165:36)
```

This PR simply sets `this.app` before calling `intlConfig`.

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -31,8 +31,8 @@ module.exports = {
     this._super.included.apply(this, arguments);
 
     var app = this._findHost();
-    this.opts = this.intlConfig(app.env);
     this.app = app;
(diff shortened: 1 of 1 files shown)


jasonmit:
Thanks, I'll version tonight

jasonmit:
2.16.2, thanks!  Unsure when that started.  I need to figure out a way to test my addon when used within another addon.

null-null-null:
Thanks for the quick turnaround, @jasonmit!

PR #417, opened 2017-01-09
this.intl.tH() - added function support HTML via service

aalasolutions:
New function added to support HTML based message from component. I required this as I am using Ember Notify which generates messages from component.

### DIFF
--- a/packages/ember-intl/addon/services/intl.js
+++ b/packages/ember-intl/addon/services/intl.js
@@ -89,6 +89,15 @@ const IntlService = Service.extend(Evented, {
 
     return this.formatMessage(translation, ...args);
   },
+  
+  
+  tH(key, ...args) {
+    const [ options ] = args;
+    const translation = this.findTranslationByKey(key, options && options.locale);
+
+    return this.formatHtmlMessage(translation, ...args);
+  },
+
 
   exists(key, optionalLocales) {
     let locales = optionalLocales;


packages/ember-intl/addon/services/intl.js
```diff
@@ -89,6 +89,15 @@ const IntlService = Service.extend(Evented, {
 
     return this.formatMessage(translation, ...args);
   },
+  
+  
+  tH(key, ...args) {
```
jasonmit: I'm in agreement that this is needed, I just need to consider the API.  The method name is bothering me more than it should :)

We could possibly make it an option of the `t` method:
`this.t('foo.bar', { html: true })`

aalasolutions: Well making it argument will be confusing with translations with variables. If HTML is filtered via arguments it will increase logic on ember-intl end also bit difficult on component end. 

tH bothered me too but what is the shortest and closest to existing this.t() function?

jasonmit: Could be a symbol to avoid collision with arguments but complicates the API even more.

```
import { html } from 'ember-intl';
// ...
this.t('foo.bar', { [html]: true });
```

Perhaps we go with what you've done but rename it to `tHtml`.  Could you add tests to cover the feature?

aalasolutions: I am a lost soul for unit testing :blush: :smirk:

vvohra:
Was there an alternative for this using t-html in `intl` service? Does this PR just need tests?

vvohra:
@jasonmit bump regarding this. Are just tests needed?

aalasolutions:
@vvohra I had used this .th feature in my code and is working for me. I really need to learn testing. As far as I know, only tests are pending for this

vvohra:
Oh, I just figured out that I think the idea is that `intl.t()` will accept `htmlSafe` to signify if the string is html.

@aalasolutions I've also used a similarly named method, but was wondering if we would see this in Ember-Intl.

jasonmit:
@vvohra @aalasolutions sorry for not circling back down the right avenues I was commenting on other threads with the reasoning behind this.

@vvohra is exactly right.  What I want to arrive at is a similar API to ember-i18n and that will involve introducing `htmlSafe` as an attribute for the `t` method on the service (and the helpers).

The helper work landed in 2.25.0 to deprecate some of the older functionality for the helpers and the next phase is introduce `htmlSafe` into the `t` method of the service.  I'll try and land that tomorrow since it seems like a relatively safe introduction without a breaking change.

aalasolutions:
@jasonmit thank you for giving this full thoughts and just not doing monkey patch like I did. 

@vvohra 👍

jasonmit:
Just looping back to say I published 2.26.0 to introduce htmlSafe support within the `t` method.

Usage:
```
intl.t('title.greeting', {
  htmlSafe: true,
  name: model.name /* name will be escaped */
});
```

jasonmit:
https://github.com/ember-intl/ember-intl/commit/4d62f2b9331181b0555afcf3b3e96dc8c23ea26c

In an attempt to reduce the API surface area, I've removed `formatHtmlMessage` and `format-html-message` in favor of `htmlSafe=true` on `formatMessage` and `format-message`

This is going to be part of 3.0 (on master currently).

Issue #411, opened 2016-12-30
getLocalesByTranslations not documented


patric-eberle:
By luck, I stumbled over the getLocalesByTranslations method, when searching for a solution to get the available translations. Can you please add that to your service documentation?

vvohra:
I was going to document it but I realized that `getLocalesByTranslations` does not include locales that were sideloaded. Is that intentional?

jasonmit:
Yes this is known and one of the reasons I didn't mark it officially public. I do plan to update the API to clean up how translations are resolved and in turn will fix this.

High on my list of things to do since it should simplify testing by removing the instance initializer.

jasonmit:
@vvohra PR opened to add a public API for this and addressed the issue you're seeing.

jasonmit:
https://github.com/jasonmit/ember-intl/releases/tag/2.17.0 closes this out

PR #418, opened 2017-01-10
Adding a public API for seeing all available locales registered by translations

jasonmit:
Related to #411, fixing how translations are discovered to include support for lazily-loaded translations.  Also adds a new public API which is a computed property called `locales` on the intl service.  I've deprecated `findLocalesByTranslations` in favor of `locales`.


jasonmit:
I'll add another PR documenting the new `locales` property before versioning.

PR #422, opened 2017-01-14
Removing the need for an instance initializer

jasonmit:
Complicates the hell out of the test story, isn't needed inside of an initializer.

### DIFF
--- a/packages/ember-intl/addon/services/intl.js
+++ b/packages/ember-intl/addon/services/intl.js
@@ -1,3 +1,5 @@
+/* globals requirejs, Intl */
+
 /**
  * Copyright 2015, Yahoo! Inc.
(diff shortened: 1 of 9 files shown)


packages/ember-intl/addon/services/intl.js
```diff
@@ -91,6 +113,33 @@ const IntlService = Service.extend(Evented, {
     return get(this, 'adapter.seen').map(l => l.localeName);
   }).readOnly(),
 
+  /**
+   * Peeks into the requirejs map and registers all locale data objects found.
+   * This is also very likely to be removed soon.
+   *
+   * @private
+   */
+  _hydrate() {
```
jasonmit: This garbage is going away, will be lazily done as part of the addTranslation logic and will go through the owner lookup/resolveRegistration api.

PR #423, opened 2017-01-15
3.0

jasonmit:
* [x] BREAKING: remove deprecated APIs (`intl-get`, `createLocale`)
* [x] BREAKING: removed `baseLocale` from configuration
* [x] BREAKING: `addTranslation`/`addTranslations` are now synchronous by default (no longer return promise)
* [x] BREAKING: remove `translationsFor` from intl API
**Migration**: `localeFactory` is the new method name as they return locale models, which contain the translations, but also other utility methods i.e., `lookup`, `has`, `addTranslation`, `addTranslations`
* [x] BREAKING: remove baseLocale logic where baseLocale's translations were merged into every locale
**Migration**: Explicitly set a fallback locale, ie: `intl.setLocale(['fr-ca', 'fr-fr']);`
* [x] BREAKING: lazily instantiate translations models and load CLDR data
  * removes dubious logic that walks undocumented/private `requirejs._eak_seen`
  * Changes behavior of `locales`, `getLocalesByTranslations`.  They'll now only return locales that have been seen/used by the application
  * [x] expose a `config/environment` property for all known locales that were detected at build time so that there is an method equivalent for past behavior for those upgrading to 3.0
    * **Migration**: `config.intl.locales` now contains all locales that were discovered during building of the app


jasonmit:
Build will continue to fail until I publish due to `yarn install` not respecting linked packages on install.

vvohra:
I've been following this work. I'm wondering why it was closed? Are these changes going to be done in another PR?

jasonmit:
@vvohra I plan to redo in smaller PRs, broken up by feature.  Current commit history doesn't tell a good story.

PR #437, opened 2017-02-26
Listing translation files 

estshy:
Fixing listing translation files so only `.json` and `.yaml` would be considered as valid translation files.

I need that because I am keeping my translations inside pods directory and now it's trying to find translations inside `.js` files which is wrong for sure.  It might be also useful for performance reason.


estshy:
@jasonmit That's correct. I've updated PR.

jasonmit:
I think I'll end up using broccoli-funnel on the translation node to include only *.json,yaml,yml which should resolve this issue.  I'll PR it tomorrow.

jasonmit:
@estshy https://github.com/ember-intl/ember-intl/pull/440 should also fix the issue, just filtering it out further upstream

PR #438, opened 2017-02-27
Show emoji flags for log lines with locales

Turbo87:
bildschirmfoto 2017-02-27 um 22 10 42

this is using:

- https://github.com/iarna/has-unicode to figure out if the terminal supports unicode (and possibly emojis)
- https://github.com/10xjs/locale-emoji to turn locale string to emoji

/cc @jasonmit

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,10 +72,12 @@
     "ember-intl-relativeformat": "^2.4.0",
     "exists-sync": "0.0.3",
     "extend": "^3.0.0",
+    "has-unicode": "^2.0.1",
     "intl": "1.0.1",
     "intl-messageformat-parser": "^1.2.0",
     "js-yaml": "^3.3.1",
     "json-stable-stringify": "^1.0.0",
+    "locale-emoji": "^0.1.0",
     "mkdirp": "^0.5.0",
     "silent-error": "^1.0.0",
     "walk-sync": "^0.2.6"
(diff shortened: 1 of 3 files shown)


jasonmit:
Love it 👍   Merge when ready

package.json
```diff
@@ -72,10 +72,12 @@
     "ember-intl-relativeformat": "^2.4.0",
     "exists-sync": "0.0.3",
     "extend": "^3.0.0",
+    "has-unicode": "^2.0.1",
```
jasonmit: Should also update the yarn lockfile

Turbo87: I used yarn to install so lockfile update is apparently not necessary as some subdependency depends on it already. Likely npm via ember-cli...

Turbo87:
Yeah, will need a bit of cross platform testing, but looks promising so far.

package.json
jasonmit: 👍

Turbo87:
just noticed that it's using the UK flag for en-us though... Something is not working as expected yet :thinking_face:

Turbo87:
bildschirmfoto 2017-02-27 um 22 31 48

Looks like we need to pass the country part in uppercase. I'll file an issue to support lowercase too.

stefanpenner:
🎉

jasonmit:
@Turbo87 I kicked off another build since the last one hung.

PR #449, opened 2017-03-02
3.0

jasonmit:
## What's left?
* [ ] document breaking changes in a markdown within `docs`
* [ ] convert over the commits which made the CLDR/translations resolve lazily
  * removes dependency on private APIs in favor of public owner API


jasonmit:
Going to reboot this effort again soon since a lot of changes have been going into master.

Issue #109, opened 2015-05-20
Collaborate with ember-i18n / ember-cli-i18n


acorncom:
Howdy, saw your extension recently.  It looks like there are some nice parts to it that aren't yet nicely handled in the Ember ecosystem (all the formatters especially).  However, other parts seems to overlap to different degrees with ember-i18n and ember-cli-i18n.  Would it be possible for you guys to collaborate with @jamesarosen / @bcardarella to end up with one awesome add-on?

@jamesarosen is working on v4.0 of his add-on and expressed willingness to discuss things here: https://github.com/jamesarosen/ember-i18n/pull/211#issuecomment-103727431

And there has been discussion of working together between the other two extensions over the last 5 months or so here: https://github.com/dockyard/ember-cli-i18n/issues/56

Thanks, look forward to seeing how this might go!

jasonmit:
I stumbled upon the restart-cli PR over the weekend after seeing a tweet about it and noticed how closely the two projects are organically aligning and was also considering to reach out to @jamesarosen -- so thanks for taking the initiative :)

I'm all in favor of merging the two projects together to focus on building one solution.  Right now, I'm the sole maintainer of the project and have been actively looking for others to join in on the effort of making an approachable internationalization addon.

stefanpenner:
> I'm all in favor of merging the two projects together to focus on building one solution. Right now, I'm the sole maintainer of the project and have been actively looking for others to join in on the effort of making an approachable internationalization addon.

:+1:

thoov:
:+1:

acorncom:
:+1:

jamesarosen:
@jasonmit what do you see as the key distinctions between this any my [v4.0-alpha proposal](https://github.com/jamesarosen/ember-i18n/pull/213)?

Some background: I enumerated some of the [core differences](https://github.com/dockyard/ember-cli-i18n/issues/56) between ember-i18n and ember-cli-i18n a while back. Here's how I addressed those differences in v4:

 * ordered vs named interpolations: I stuck with named interpolations, but made it easy to swap in a custom compiler.
 * rerendering when the locale changes: I adopted ember-cli-i18n's use of `Stream`s. Unfortunately, that's still a private API. At least it's well isolated.
 * globals vs ES6 modules: I adopted ES2015, which is the standard for ember-cli addons
 * HTML safety: the default compiler treats interpolations as unsafe by default and emits safe strings by default. Again, the custom compiler can easily change this
[...]

PR #450, opened 2017-03-02
Lazily materialize translations and lazily register CLDR data

jasonmit:
The goal of this is to get off walking `requirejs._eak_seen` to register bundled translations and CLDR.

Additionally, this will allow us to do add support for eagerly compile translations when the locale is initialized (usually when the app is initialized).  This is how ember-i18n is implemented and was mentioned early on in #109.

Breaking changes:
* intl.locales is now a method instead of a CP.  Will be mentioned in the release notes and 3.0 migration doc.


addon/services/intl.js
```diff
@@ -37,205 +38,226 @@ function formatterProxy(formatType) {
...
 
+  /* @private */
   localeFactory(localeName) {
     return get(this, 'adapter').localeFactory(normalizeLocale(localeName), true);
   },
 
-  findTranslationByKey(key, localeName, options) {
-    return this.lookup(key, localeName, options);
+  /* @private */
+  registerLocale(localeName) {
```
jasonmit: This is invoked when a localeName hasn't been seen by the service.  It will attempt to find and register the bundled translations and CLDR "on-demand" versus when the service is initialized (which involved abusing private APIs).

tests/unit/helpers/format-message-test.js
```diff
@@ -179,7 +179,7 @@ test('able to discover all register translations', function(assert) {
   assert.expect(1);
   this.intl.addTranslation('es_MX', 'foo', 'bar'); /* tests that the locale name becomes normalized to es-mx */
   this.intl.exists('test', 'fr-ca');
-  assert.equal(this.intl.get('locales').join('; '), 'en-us; es-es; fr-fr; es-mx');
```
jasonmit: I don't have a solution for this yet, but it may involve determine locales that have been bundled at build time and making that list available as `ENV.intl.bundledLocales`.

addon/adapters/default.js
```diff
@@ -4,62 +4,55 @@
  */
 
 import Ember from 'ember';
-import Translation from '../models/translation';
 
 const { computed, get, A:emberArray, getOwner } = Ember;
 
 const DefaultTranslationAdapter = Ember.Object.extend({
+  /* private */
```
Turbo87: how about using `/** @private */` instead?

Turbo87:
I'm not that deeply familiar with the code yet so this is a little hard to review with the little context I got.

> this will allow us to add support for eagerly compile translations when the locale is initialized

I'm not sure I understand. Do we now compile *all* translations at startup vs. only the needed translations lazily once used? That wouldn't sound like an advantage to me. 🤔

jasonmit:
Right now we load all the translations on at boot (no compiling until the translation is used).  What this PR does is only load the translations for the locale that is set and remove the use of accessing the private ember loader APIs.

The eagerly precompile is something ember-i18n does and was mentioned in the discussion of merging the two projects.  This just makes that approach possible in the future but for now isn't the objective.

addon/adapters/default.js
jasonmit: :+1:

Turbo87:
> only load the translations for the locale that is set

what do you mean by "load"? they will by default still be bundled in the app.js, correct?

> The eager compiling on app-boot is something ember-i18n does and was mentioned in the discussion of merging the two projects.

I understand that reducing differences is a good thing, but I'm not sure yet where the advantage is of eager compiling? It seems to me that this adds a certain amount of startup cost without any advantages. Am I missing something?

jasonmit:
Yes the translations are still bundled by default.   This changes how those bundles are discovered and loaded into the service.  Instead of walking the requirejs._eak_seen on app boot to add the translations into the service, we wait until the locale is actually used before doing that work.  This isn't a performance optimization but rather cleaning up debt that I previously couldn't due to the API breaking nature of this change.

You're not missing anything, compiling on boot is something I was never convinced on either.  What I think they want is precompiling at build which neither project does.  Ember i18n compiles eagerly on app boot (for the active locale) and if they want that feature I should at least make that option possible in the future.

Turbo87:
> What I think they want is precompiling at build which neither project does.

I don't think that's possible in the current state. We could precompile to the ICU AST, but then the payload size would be huge

> Ember i18n compiles eagerly on app boot (for the active locale) and if they want that feature I should at least make that option possible in the future.

I don't think I agree. Why offer and develop a feature that degrades performance?

jasonmit:
> I don't think that's possible in the current state. We could precompile to the ICU AST, but then the payload size would be huge

Totally agree.  I mentioned that early on: https://github.com/ember-intl/ember-intl/issues/109#issuecomment-105688470

> I don't think I agree. Why offer and develop a feature that degrades performance?

The gist is in the future I'd like to support ember-i18n's functionality as it exists today in a separate addon.  That addon would be an extension/plugin to ember-intl "core" and the ICU functionality (message format, relative format) would go into a separate plugin.  There wouldn't be a performance hit to how things behave today in respect to the ICU plugin and ember-i18n's functionality would continue to behave how it behaves today.

tests/unit/helpers/format-message-test.js
Turbo87: correct me if I'm wrong: we usually know the available locales at build-time, what's keeping us from passing that information to the runtime service? in the case of sideloading translations that are not compiled by us we could offer a property somewhere that you could overwrite at build-time (e.g. by traversing a file system structure).

jasonmit: That's what I've been considering, likely will write it to the config.

jasonmit: @Turbo87 PR updated to include this behavior.

Turbo87: would it be possible to split this PR up into multiple PRs? it has become pretty larger and hard to review... 😞 

e.g. the `@private` comments seem unrelated and the build-time locale list building could probably be implemented independent from the original changes in this PR, right?

jasonmit:
Closing to reduce down into multiple PRs.

PR #454, opened 2017-03-08
add int.locale as depKey to programmatic examples

sly7-7:
This way the cp are invalidated when the end-user change the locale


jasonmit:
Thanks!

Issue #480, opened 2017-03-25
Consistency regarding format-message and t


vvohra:
- [x] I am on the [latest](https://github.com/jasonmit/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/jasonmit/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 2.8
- Ember CLI Version: 2.6.3
- Ember Intl Version: 2.24.1

### Question

So:
Service methods: `formatMessage` does NOT accept a translation key, but `t` does

Helper methods: both `formatMessage` and `t` accept translation keys.

Is that intentional? Should it be consistent?

jasonmit:
Legacy logic that was moved forward while I waited to see how ember-i18ns API settled around formatting html-safe translations.

`format-message` helper in 3.0 will take a string literal, to match the service API.  I first need to deprecate the method in 2.x to encourage anyone using it to converting to the `t` helper.

The `t` helper will take an `htmlSafe` attribute to signify returning an html safe string.  The `t` method on the service will also respect this attribute.

jasonmit:
I released 2.25.0 with all the deprecations I was referring to.

jasonmit:
https://github.com/ember-intl/ember-intl/commit/aec30932e7d125fb3e552c51571e25948a4838e0 resolves this.

These changes will all land in 3.0 (master)

PR #486, opened 2017-03-27
deprecate current use of {{format-message}} and {{format-html-message}}

jasonmit:
This is to deprecate the translation lookup behavior with `{{format-message}}` in favor of using the `{{t}}` helper for lookups & compiling.

This PR also introduces `htmlSafe` attribute to the `t` helper, which will escape arguments and render an ElementNode.  If you're using `{{format-html-message}}` with lookups you'll want to convert those to the `t` helper and pass `htmlSafe=true`.

In 3.0.0, `{{format-message}}` and `{{format-html-message}}` will just accept string literals (like the service APIs do today with `formatMessage` & `formatHtmlMessage`).  If you're using the literal helper today, i.e,:
`{{format-message (l '{name} took {createdOn, date, long}') name=model.name createdOn=model.createdOn}}` this will continue to be supported and will not be deprecated.  In 3.0.0, you'll be able to drop the `{{l}}` helper in favor of passing the string literal as the first argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ Recompute the relative timestamp on an interval by passing an `interval` argumen
 [List of supported format date options](https://github.com/ember-intl/ember-intl/blob/master/docs/format-relative-options.md)
 
 ### Format Message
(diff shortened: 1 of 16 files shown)


jasonmit:
Fixes #480

PR #516, opened 2017-07-12
Eagerly instantiate intl service

jasonmit:
Since looking up a locale is async, it's possible to render before the hydration has finished.


jasonmit:
Beta is failing but always has been failing.  Will need to figure that out another day.

Issue #401, opened 2016-11-16
PROPOSAL: allow for translations to be grouped with components/by route


jasonmit:
The title is vague intentionally, this discussion spawned from #400.

The gist is the ability to group translations with components or by route and in a way that is addon agnostic that works with both ember-i18n and ember-intl.

What I had in mind was naively glob for **/*.(yaml,json), filter on everything that looks like a locale, merge by locale, and remove them from the app tree as they're merged together.

But given a few people have interest in this, I figured we can open it up to discussion since there has been interest in supporting specific app shapes (pods, unified module structure) along with auto-namespacing/scoping translations.

/cc @alexlafroscia

alexlafroscia:
Definitely interested in grouping translations with a component's "pod", regardless of the new/current directory structure.  I've been kicking around a few different questions as I've thought of this, both of which Jason proposes solutions for

1. How do we organize the translations within a component? Should there be a `translations` directory inside the component, or a single `translations.{json, yml}` where the top-level keys are the locale?
2. Should a component-scoped translation file get "namespaced" by the component name? This would definitely be harder to implement, but might be worth it to avoid collisions. Further questions include:
  - How do the `t` helper automatically add the namespace based on where it is being used?

The solutions proposed above are:

1. Go based on file name for each locale, let people decide if they want to use a `translations` directory by grabbing anything that looks like a translation
2. No namespacing; warn on collisions

That's fine with me, although I'm a little apprehensive about not namespacing the component's translations. I wonder if it would be possible to do something similar to `ember-css-modules` where the "real" values of the keys get obfuscated in some way...

jasonmit:
I helped write `local-class` and while I like the idea, it is dependent on htmlbar preprocessors and a private API looking at the debug container key.  That will mean we could also need to reopen Ember.Component and expose a lookup method, deviating from the current API a bit.

All of this is why I'm apprehensive to do adopt this, along with the side-loading issues you mention and the inability to share the work this lib with ember-i18n.

rlivsey:
Doing this in 2 phases seems like a sensible approach then (where phase 2 may never happen but can be discussed separately!)

1. load translations.yml from components with global keys
2. at some future time, maybe, possibly, add local scoping

Regarding file vs directory, I think I'd probably prefer a file per locale in a dir on the projects where we have a lot of translations. On smaller projects I can see the benefit of having just one translations.yml file with multiple locales in, but directory also has the benefit of consistency with the top level ones.

[...]

PR #518, opened 2017-07-24
treeForTranslations hook

jasonmit:
The goal is summed up here https://github.com/ember-intl/ember-intl/issues/401#issuecomment-316629409

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,25 +42,37 @@ module.exports = {
     if (this._isHost) {
       this._options = this.intlConfig(host.env);
       this._hostsLocales = this.findLocales();
-      this._translationTree = this.getTranslationTree();
+      this._translationTree = this.buildTranslationTree();
     }
   },
 
   /*
-   * @method getTranslationTree
+   * @method buildTranslationTree
    * @private
    */
-  getTranslationTree() {
-    let nodes = [];
+  buildTranslationTree() {
+    let registry = this.app.registry;
+    let addonNodes = [];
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -19,114 +18,205 @@ const TranslationReducer = require('./lib/broccoli/translation-reducer');
...
+    addons.forEach(addon => this.processAddon(addon, nodes));
+  },
+
+  /*
+  * @method processAddon
+  * @param {object} addon model
+  * @param {array} array of broccoli node/tree/directories
+  * @private
+  */
+  processAddon(addon, nodes) {
```
jasonmit: Ripped from ember-cli-fastboot.  They are the only addon I know that introduces an addon hook and it looked like a good implementation for what I need.

index.js
```diff
@@ -19,124 +18,218 @@ const TranslationReducer = require('./lib/broccoli/translation-reducer');
...
+    /* walk through each addon and grab its translation tree */
+    let addonTranslationPath = path.join(addon.root, this.translationPathForAddon(addon));
+    let _translationTree;
+
+    if (existsSync(addonTranslationPath)) {
+      /* verify these are not setting up watchers */
```
jasonmit: I probablly want `UnwatchedDir` here

vlascik:
Nice, could use a mention in docs though? People might not find out about this otherwise.

jasonmit:
@vlascik it's on master, which is not versioned.  It will be documented when it is released and the APIs are marked "public".

PR #521, opened 2017-07-28
Use "New Module Imports"

Turbo87:
see https://github.com/ember-cli/ember-rfc176-data


Turbo87:
CI failure seems unrelated 😞

jasonmit:
@Turbo87 I'll need to dig into all these CI failures across the addons I maintain - unsure what changed upstream.  At the least I should be allow beta and canary to fail.

package.json
```diff
@@ -40,7 +40,7 @@
     "chalk": "^1.1.3",
     "cldr-core": "28.0.0",
     "ember-assign-polyfill": "^2.0.1",
-    "ember-cli-babel": "^6.0.0",
+    "ember-cli-babel": "^6.6.0",
```
jasonmit: @Turbo87 I suspect the polyfill is occurring here

jasonmit: I confirmed this, disregard.  https://github.com/babel/ember-cli-babel/blob/master/package.json#L39

jasonmit:
@Turbo87 LGTM.  I will version a minor soon

Turbo87:
@jasonmit I actually was just made aware that I introduced a bug here since the new imports are not available in the app folder. Will fix tomorrow :(

PR #523, opened 2017-07-29
Use "New Module Imports"

Turbo87:
This time without mistakenly converting the files in the `app` folder too. Sorry about that... 😞 

proper solution for the `app` folder is to just reexport stuff in the `app` folder and have the real code in the `addon` folder.


jasonmit:
> proper solution for the app folder is to just reexport stuff in the app folder and have the real code in the addon folder.

Agreed.  I'll go through once I merge this and work on reexporting from addon

PR #535, opened 2017-09-21
Initial support for module unification

### LINKED ISSUE #534: Support for apps using module unification resolver
- [x] I am on the [latest](https://github.com/jasonmit/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/jasonmit/ember-intl/issues) of this repo and believe that this is not a duplicate

[Module Unification RFC](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)

Where full support may be premature, it would be nice to have at least the basic functionality. Right now addon crashes due to custom `config:*` registrations.

jasonmit:
Fixes #534

TODO:
* [ ] fix tests
* [ ] filter `src` out of the app tree if `EMBER_RESOLVER_MODULE_UNIFICATION` is falsey
* [ ] verify no APIs were broken

### DIFF
--- a/addon/adapters/default.js
+++ b/addon/adapters/default.js
@@ -3,56 +3,56 @@
  * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
 
-import EmberObject, { get, computed } from '@ember/object';
-import { A as emberArray } from '@ember/array';
 import { getOwner } from '@ember/application';
+import { A as emberArray } from '@ember/array';
 import Translation from '../models/translation';
+import EmberObject, { computed } from '@ember/object';
+import EmptyObject from '../utils/empty-object';
 
 const DefaultTranslationAdapter = EmberObject.extend({
-  _seen: null,
-
-  /** @private **/
-  locales: computed('_seen.[]', function() {
-    return get(this, '_seen').map(l => l.localeName);
-  }).readOnly(),
-
-  /** @private **/
   init() {
-    this._super();
+    this._super(...arguments);
     this._seen = emberArray();
+    this._owner = getOwner(this);
+    this._cache = new EmptyObject();
+    this._Factory = this._owner.factoryFor('model:ember-intl-translation');
   },
 
   /** @private **/
+  _seen: null,
+
+  /** @private **/
+  _cache: null,
+
+  /** @private **/
+  _Factory: null,
+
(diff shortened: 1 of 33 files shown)


tchak:
I can report that it works on https://github.com/emberjs-paris/website. Thanks!

jasonmit:
Closing for now since there are a bit of conflicts and module unification still needs work before I'm willing to officially support.

NullVoxPopuli:
update:

instead of changing node_modules...

`src/init/initializers/initl-hacks.ts`:
```ts
import Application from '@ember/application';
import IntlService from 'ember-intl/services/intl';

export function initialize(application: Application): void {
  IntlService.reopen({
    lookup(key: string, localeName: string, options: any = {}) {
     const localeNames = this.localeWithDefault(localeName);
     const translation = this._adapter.lookup(localeNames, key);

     if (!options.resilient && translation === undefined) {
       const missingMessage = this._owner.resolveRegistration('util:intl/missing-message');

       return missingMessage.call(this, key, [localeNames]);
     }

     return translation;
   },
  });
}

export default {
  initialize
};

```

walter:
Does this work without typescript in the host app?

I'm migrating a small app that already uses `ember-intl` to the "modern ember" stack with the eventual addition of typescript. However, I want to add typescript last after MU and decorators. Was hoping to get things in a working state before jumping to typescript.

NullVoxPopuli:
Oh yeah. Typescript is always optional. It's just some sugar on top of JavaScript

walter:
I was talking specifically about the code you posted in `src/services/intl.ts`. Do you have a non-typescript version or compiled to js version?

walter:
Plus you removed this block:

```TypeScript
// DO NOT DELETE: this is how TypeScript knows how to look up your services.
declare module '@ember/service' {
  interface Registry {
    'intl': EmberClearIntl
  }
};
```

Thanks for verifying what I suspected. It's good to be explicit about these things for those reading in the future.

I'll have a crack at getting it working and report back if there are any other differences.

NullVoxPopuli:
That registry bit isn't even required for typescript. It's just a helper for the non-decorated service-injection.

walter:
Thanks for your help @NullVoxPopuli, got it working with your additions.

PR #548, opened 2018-01-31
Update to ember 2.18

cibernox:
This is the first step towards making this addon work on apps without ember-cli-shims


tests/unit/helpers/format-message-test.js
```diff
@@ -160,7 +160,8 @@ test('able to discover all register translations', function(assert) {
 
 test('should respect format options for date ICU block', function(assert) {
   assert.expect(1);
-  this.render(hbs`{{format-message (l 'Sale begins {day, date, shortWeekDay}') day=1390518044403}}`);
+  this.day = 1390518044403;
```
cibernox: The reason for this change is that glimmer-vm has a bug in which VERY big integer throw an error: https://github.com/emberjs/ember.js/issues/15635

tests/unit/services/intl-test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ember/new-module-imports */
```
Turbo87: I *think* you could use https://github.com/rwjblue/ember-debug-handlers-polyfill instead of the `Ember.warn` override below, but seem fine to explore that in a separate PR

Turbo87:
@jasonmit I'm deferring to you for the final decision

jasonmit:
Thank you, and super excited that this fixes travis CI :)  Much appreciated!

PR #549, opened 2018-01-31
Use RSVP.resolve instead of RSVP.cast

cibernox:
This breaks in the latest Ember apparently.


jasonmit:
Thanks for capturing this, I'll see about updating the smoke tests to handle regressions.

Versioned as 2.23.1

PR #550, opened 2018-02-01
Ensure addon works without jquery and/or app-shims

cibernox:


tests/unit/helpers/format-html-message-test.js
```diff
@@ -1,83 +1,88 @@
...
-});
+  test('message is formatted correctly with argument', async function(assert) {
+    assert.expect(1);
+    await render(hbs`{{format-html-message (l "Hello {name}") name="Jason"}}`);
+    assert.equal(this.element.textContent, 'Hello Jason');
+  });
 
-test('should throw if called with out a value', function(assert) {
-  assert.expect(1);
+  test('should throw if called with out a value', async function(assert) {
```
jasonmit: debuggers can be removed.  was there an issue with this test and if so want me to look into it?

cibernox: There is a problem, but it's not specific to this project.
It seems that is hard to catch assertions thrown during render. The usual things you'd expect like

```
render(hbs`thing-that-fails`).catch((e) => asserterror);
// or
try {
  await render(hbs`thing-that-fails`);
} catch (e) {
  assertError
}
```

don't work. 
I believe @rwjblue was inventigating how to make that more natural.

cibernox:
I'm keeping an eye on https://github.com/emberjs/ember-test-helpers/issues/310
Once this is fixed fixing the remaining tests should be a moment.

jasonmit:
https://github.com/ember-intl/ember-intl/pull/560 thanks @cibernox  :)

PR #552, opened 2018-02-05
Remove deprecated api

cibernox:
This deprecation is added in Ember 3.0


Turbo87:
I was about to open another issue but then I though: maybe check the latest PRs first... 😆 

Thanks again!

jasonmit:
Versioned as 2.33.0 & sorry for the late version cut.

Issue #553, opened 2018-02-12
Optionally throw hard error for missing translations


Turbo87:
Related to my suggestion in https://github.com/ember-intl/ember-intl/issues/547 it would be nice if the build could be configured to throw a hard error when missing translations are detected.

jasonmit:
```js
// app/utils/intl/missing-message.js
export default function()
```

jasonmit:
Reopening, I now realized you want this to be a build-time error.

Turbo87:
> I now realized you want this to be a build-time error

yeah, exactly. I would like our builds to fail if people submit/merge PRs with missing translations. This would need to be combined with https://github.com/ember-intl/ember-intl/issues/547 though as we would want to ignore a subset of translations that are indeed only relevant for certain locales.

makepanic:
A couple weeks ago I tried to implement this as a failing test but that was awkward as it had to be in the lintTree hook.

Is it ok if I implement this again on current master with throwing an error?

jasonmit:
@makepanic yes please

makepanic:
I think this is also fixed with https://github.com/ember-intl/ember-intl/pull/588

Issue #547, opened 2018-01-29
Ignore certain "missing message" warnings


Turbo87:
In our app we have several translations that are only used on specific locales, but `ember-intl` is constantly warning about them being missing in the other locales, which makes the signal-to-noise ratio quite bad. It would be great if there was an API that would allow us to filter out these messages before they are being passed to the console.

Maybe something like this:

```js
let app = new EmberApp(defaults, {
  'ember-intl': {
    ignoreMissingMessage(message, locale) {
      // your custom logic here
    },
  },
});
```

PR #561, opened 2018-05-06
feat: if failOnMissingTranslation option is set, throw missing errors

makepanic:
This is an initial stab at fixing #553.

It introduces the `failOnMissingTranslation` `config/ember-intl.js` option which throws an error if any missing translation keys are detected.

If the general concept is ok, I'll add some documentation and additional tests.

There are a couple open questions:

- should we throw in the broccoli reducer and failing the whole build?
- would this be better by injecting failing tests, like ember-template-lint does it?

This currently ignores #547 but could be easily extended to solve it, if we add support for specifying `ignoreMissingMessage` (or similar) in the `config/ember-intl.js` config.

### DIFF
--- a/lib/broccoli/translation-reducer.js
+++ b/lib/broccoli/translation-reducer.js
@@ -62,7 +62,7 @@ class TranslationReducer extends CachingWriter {
     return locale;
   }
 
-  _logLocale(locale, message, options) {
+  _buildLogMessage(locale, message) {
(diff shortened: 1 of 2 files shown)


makepanic:
I think following template-lint and injecting failing tests is the better approach as it won't cause any downtime when developing an app.

I'll wait for #563 and redo this PR.

Issue #562, opened 2018-05-06
Missing keys warning requires baseLocale


makepanic:
When sideloading translations, there is no warning about missing translations.
Technically it should be possible by just picking any of the `locales` and using its keys as `defaultTranslationKeys` when [finding missing keys](https://github.com/ember-intl/ember-intl/blob/master/lib/broccoli/translation-reducer.js#L163).

This could be implemented as fallback when no `baseLocale` is set.

If this is something ember-intl wants to do, I can open a PR to implement the behavior.

- [x] I am on the [latest](https://github.com/jasonmit/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/jasonmit/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3
- Ember CLI Version: 3
- Ember Intl Version: 2.33
- Browser(s): Chrome
- Node Version: 10

### Steps to Reproduce

**given**
1. set `baseLocale` to `null` as required for sideloading translations.
2. have missing translations when diffing all translation files

**expected**
Intl should emit missing translation warnings

**actual**
No warning gets printed

jasonmit:
`baseLocale` has no impact other than being the "source of truth" to compare against.  Unsure I see value in randomizing the `baseLocale` from `locales`.  The comment within ember-intl.js could be updated to be more clear.

`"This property is optional, and if you rely on sideloading translations then this should be null"`  could be updated to `"This property is optional.  If your translations are external to your application, and side loaded, this property should be `null`"`

Open to suggestions.

makepanic:
I'm unsure if I misunderstood you,
 but a possible suggestion is to only use a randomized baseLocale for [finding missing translation keys](https://github.com/ember-intl/ember-intl/blob/master/lib/broccoli/translation-reducer.js#L162), as they're not found without a baseLocale being set.

It shouldn't really influence any current behavior except the missing translation logging.

jasonmit:
> possible suggestion is to only use a randomized baseLocale

What if instead we search for all known keys across all translation files.  Then, compare each locale against that manifest of translation keys and spit out a report.  This could then ultimately remove `baseLocale` all together (which I really want to since it's a source of confusion where folks interpret it as being the default locale).

```
"welcome.title" was not found in "en-us", "en-ca", & "fr-fr"
"welcome.footer" was not found in "en-ca"
```

PR #563, opened 2018-05-08
feat: find missing translations without baseLocale being set

makepanic:
This is an initial stab at fixing #562.

It reduces the usage of `baseLocale` to only being used when merging a translation with the default translation.

### DIFF
--- a/lib/broccoli/translation-reducer.js
+++ b/lib/broccoli/translation-reducer.js
@@ -80,12 +80,22 @@ class TranslationReducer extends CachingWriter {
     }
   }
 
-  findMissingKeys(target, defaultTranslationKeys, locale) {
-    let targetProps = propKeys(target);
+  findMissingTranslations(translations) {
+    const locales = Object.keys(translations);
 
-    defaultTranslationKeys.forEach(property => {
-      if (targetProps.indexOf(property) === -1) {
-        this._logLocale(locale, `"${property}" message is missing`);
+    // [['de',['foo.bar', 'bar.baz'], ['en',['foo.baz','bar.baz']]
+    const localeKeys = locales.map(locale => [locale, propKeys(translations[locale])]);
+
+    // create flattened list of all unique translation keys
+    const allTranslationKeys = new Set(Array.prototype.concat(...localeKeys.map(([, keys]) => keys)));
+
+    allTranslationKeys.forEach(key => {
+      const notInLocales = localeKeys
+        .filter(([, translationKeys]) => !translationKeys.includes(key))
+        .map(([locale]) => `"${locale}"`);
+
+      if (notInLocales.length) {
+        this._log(`"${key}" was not found in ${notInLocales.join(', ')}`);
       }
     });
   }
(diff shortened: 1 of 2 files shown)


lib/broccoli/translation-reducer.js
```diff
@@ -80,12 +80,22 @@ class TranslationReducer extends CachingWriter {
...
+    const locales = Object.keys(translations);
 
-    defaultTranslationKeys.forEach(property => {
-      if (targetProps.indexOf(property) === -1) {
-        this._logLocale(locale, `"${property}" message is missing`);
+    // [['de',['foo.bar', 'bar.baz'], ['en',['foo.baz','bar.baz']]
+    const localeKeys = locales.map(locale => [locale, propKeys(translations[locale])]);
+
+    // create flattened list of all unique translation keys
+    const allTranslationKeys = Array.from(new Set(Array.prototype.concat(...localeKeys.map(([, keys]) => keys))));
```
jasonmit: Do not need to convert back to an array, can stay a Set

tests-node/unit/broccoli/translation-reducer-test.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+let mocha = require('mocha');
+let expect = require('chai').expect;
+
+let subject = require('../../../lib/broccoli/translation-reducer');
```
jasonmit: I'll add some additional integration tests w/ broccoli after merging

lib/broccoli/translation-reducer.js
```diff
@@ -144,25 +154,21 @@ class TranslationReducer extends CachingWriter {
     let outputPath = `${this.outputPath}/${this.options.outputPath}`;
     let baseLocale = this.options.baseLocale;
     let translations = this.readDirectory(this.inputPaths[0], this.listFiles());
-    let defaultTranslationKeys, defaultTranslation, translation;
+    let defaultTranslation, translation;
     mkdirp.sync(outputPath);
 
     if (baseLocale) {
```
jasonmit: I'll remove baseLocale once this lands

lib/broccoli/translation-reducer.js
makepanic: Thanks, I've changed it.
Forgot that `Set` has a tiny amount of array methods built-in.

jasonmit:
Thank you, thank you, thank you.

Master is a build up of 3.0.0 which will be a light release of removing deprecated APIs, removing `baseLocale` and some internal cleanups (converting addTranslation(s) to be synchronous).  I plan to version a beta release in a few days.. possibly today if I can get all the tests passing.

jasonmit:
Everything is now available in the beta tag.

ember-intl@beta

Be mindful of the APIs removed: https://github.com/ember-intl/ember-intl/commits/master

Likely the big one being the removal `format-html-message` & `t-html`.  You now pass `htmlSafe=true` as arguments to the `t` helper and `format-message`.

I plan to put together a migration doc before releasing as `latest`.

PR #566, opened 2018-05-10
Update Readme

baroquon:
Replaced reference to addon with reference to blueprint for ember-intl-dot-notation


jasonmit:
Much appreciated :)

PR #576, opened 2018-05-29
Update Integration Testing with new format tests

martynling:
The old snippet didn't work with the latest Ember test formats, so updated to use the new format tests


jasonmit:
Thank you!

PR #578, opened 2018-05-31
add migration tool to readme

snewcomer:


walter:
Looks to me like `ember-i18n-intl-migrator` package hasn't been published yet. Should we hold off merging this until it has been?

jasonmit:
It's since been published, thanks for the callout @walter.

@snewcomer I'm also going to add you as a collaborator.  Much appreciated!

PR #580, opened 2018-06-13
Adding support for treeForTranslations

jasonmit:
Part of #401 

@cibernox let me know if this solves what we discussed.

Usage:

```js
// index.js

module.exports = {
  name: 'an-ember-addon',

  /**
   * NOTE: addon's translation tree provided as an arg.
   * No need to return it if you are reimplementing behavior.
   * If you want to programmatically modify the translation node,
   * then feel free to mutate the passed in tree and return it.
   */
  treeForTranslations(/*tree*/) {
     return new BroccoliTranslationPlugin();
  }
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,7 @@ module.exports = {
   },
 
   _buildTranslationTree() {
-    const projectTranslations = path.join(this.project.root, this.opts.inputPath || 'translations');
+    const projectTranslations = path.join(this.project.root, this.opts.inputPath);
     const translationTrees = [];
 
     this._processAddons(this.project.addons, translationTrees);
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -220,36 +246,6 @@ module.exports = {
...
-    let addons = this.app.project.addons;
-    let registered = new Set();
-
-    let find = function(list, addon) {
-      // Only handle each addon once
-      if (registered.has(addon.name)) {
-        return list;
-      }
-
-      let translationPath = addon.pkg['ember-addon'].translationPath || 'translations';
```
jasonmit: I'm dropping `translationPath` in favor of `treeForTranslations`.  Addons can return a `this.treeGenerator(<any path>)`

index.js
```diff
@@ -21,30 +20,64 @@ const TranslationReducer = require('./lib/broccoli/translation-reducer');
...
+    }
+
+    if (addon.treeForTranslations) {
+      let additionalTranslationTree = addon.treeForTranslations(addonGeneratedTree);
+
+      if (additionalTranslationTree) {
+        translationTrees.push(funnel(additionalTranslationTree, { destDir: addon.name }));
+      }
+    } else if (addonGeneratedTree !== undefined) {
+      translationTrees.push(funnel(addonGeneratedTree, { destDir: addon.name }));
```
jasonmit: Translations from addons are namespaced within a folder based on the addon's name.  This is so that the app and the addon translations do not collide.  In the case of a rare case of collision of a file (i.e., the app creates a folder with the name of an addon) then the app will win.

index.js
```diff
@@ -21,30 +20,64 @@ const TranslationReducer = require('./lib/broccoli/translation-reducer');
...
   included() {
     this._super.included.apply(this, arguments);
 
-    let app = this._findHost();
-    this.app = app;
-    this.addonOptions = this.intlConfig(app.env);
+    this.app = this._findHost();
+    this.opts = this.getOptions(this.app.env);
+    this.locales = this.findLocales();
+    this.translationTree = this._buildTranslationTree();
```
jasonmit: I may end up moving this to be called where needed (within both treeFor hooks vs on `included`)

PR #587, opened 2018-06-26
BREAKING(on-hold): passthrough options to missing message and move fallback logic

### LINKED ISSUE #579: Passthrough options hash to missing-message
It would be great, if the [`missing-message`](https://github.com/ember-intl/ember-intl/blob/2273bbe1269e125135ab5da9c3077e9dd1280e7b/docs/missing-translations.md) util could also get access to the options hash. This would allow a setup like this:

```hbs
{{t "some.optional.translation" allowMissing=true}}
{{t "some.required.translation"}} {{!-- allowMissing=false is the implicit default --}}
```

```js
// app/utils/intl/missing-message.js:

export default function missingMessage(key, locales, options) {
  if (options.allowMissing) {
    return ''; // return empty string in order to not crash consumers
  }
  assert(`[ember-intl] Missing translation for key: "${key}" for locales: "${locales}"`);
}
```

This is not a dupe of https://github.com/ember-intl/ember-intl/issues/553, which is about conditionally throwing hard errors at build time. This issue is about conditionally throwing errors at run time.

buschtoens:
Closes #579.

This achieves two things:

- pass the complete `options` hash to the `missing-message` util
- offload the `default` fallback logic to the `missing-message` util, in that `missing-message` could theoretically log all steps of the resolution or even return a missing message before all `default` locales have been tried

This ultimately gives the user more control over how to handle message resolution and is really useful for mocking in tests.

Two things I am not happy about:

#### `defaults` / `default` inconsistency

IMHO the `default` should have been called `defaults` in the first place, because `default` is a reserved keyword and makes for weird code. But yeah. This'll never change. 😅 

#### Possible breakage for users of custom `missing-message` & `default`

Users that have their own `missing-message` util and also use the `default` option, will run into issues, because they are potentially not returning `undefined`, if there are still defaults to try.

The only thing I wanted to do with #579 in the first place (after I cleared some misunderstanding on my side), was to just have access to the `options` hash, when we've actually tried all `default` locales and could not find a translation.

If this potential breakage is too dangerous for your taste, I can move the fallback logic back into the `lookup` method and just pass the options through. If not, also fine – I'd like using this to log fallthroughs to the backend. :)

### DIFF
--- a/addon/utils/missing-message.js
+++ b/addon/utils/missing-message.js
@@ -2,13 +2,21 @@ import { isEmpty } from '@ember/utils';
 import { warn } from '@ember/debug';
 import links from './links';
 
-export default function missingMessage(key, locales) {
+export default function missingMessage(key, locales, options) {
   if (isEmpty(locales)) {
     warn(`[ember-intl] no locale has been set. Documentation: ${links.unsetLocale}`, false, {
       id: 'ember-intl-no-locale-set'
     });
 
     return `No locale defined.  Unable to resolve translation: "${key}"`;
+  }
+
+  if (options.default.length > 0) {
+    return undefined; // make `lookup` continue trying the other fallback defaults
+  }
+
+  if (options.resilient) {
+    return ''; // return empty string in order to not crash consumers
   }
 
   const localeNames = locales.join(', ');
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -13,6 +13,7 @@ import Evented from '@ember/object/evented';
 import { assert, warn } from '@ember/debug';
 import { makeArray } from '@ember/array';
 import Service from '@ember/service';
+import { assign } from '@ember/polyfills';
 
 import { FormatDate, FormatMessage, FormatNumber, FormatRelative, FormatTime } from '../-private/formatters';
 import isArrayEqual from '../-private/is-array-equal';
@@ -105,10 +106,12 @@ export default Service.extend(Evented, {
     const localeNames = this.localeWithDefault(localeName);
     const translation = this._adapter.lookup(localeNames, key);
 
-    if (!options.resilient && translation === undefined) {
+    if (typeof translation === 'undefined') {
       const missingMessage = this._owner.resolveRegistration('util:intl/missing-message');
+      // make sure to always pass an array to `missingMessage`
+      const defaults = options.default ? [].concat(options.default) : [];
 
-      return missingMessage.call(this, key, localeNames);
+      return missingMessage.call(this, key, localeNames, assign({}, options, { default: defaults }));
     }
 
     return translation;
@@ -124,9 +127,13 @@ export default Service.extend(Evented, {
     }
 
     while (!msg && defaults.length) {
-      msg = this.lookup(defaults.shift(), options.locale, {
-        resilient: defaults.length
-      });
+      msg = this.lookup(
+        defaults.shift(),
+        options.locale,
+        assign({}, options, {
+          default: defaults
+        })
+      );
     }
 
     return this.formatMessage(msg, options);
--- a/tests/unit/services/intl-test.js
+++ b/tests/unit/services/intl-test.js
@@ -92,6 +92,16 @@ module('service:intl', function(hooks) {
     );
   });
 
+  test('`t` should return empty string, if `resilient: true`', function(assert) {
+    assert.strictEqual(
+      this.intl.t('does.not.exist', {
+        default: ['also.does.not.exist', 'should_also_exist'],
+        resilient: true
+      }),
+      ''
+    );
+  });
+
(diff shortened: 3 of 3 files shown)


buschtoens:
> This ultimately gives the user more control over how to handle message resolution and is **really useful for mocking in tests**.

With this test helper, you can mock translations on the fly:

```ts
import castArray from 'lodash.castarray';
import last from 'lodash.last';
import { get } from '@ember/object';
import transitions from 'client/transitions';
import translations from 'client/initializers/translations';

/**
 * Takes an object and stringifies it in a deterministic way. It will also only
 * include top-level entries.
 *
 * @param obj
 */
const stringifyDeterministically = (obj: object): string =>
  JSON.stringify(obj, Object.keys(obj).sort());

/**
 * Serializes a translation invocation deterministically.
 *
 * @param key translation key
 * @param options options and variables passed along
 */
const serializeTranslation = (key: string, options?: object): string =>
  options ? `t:${key}:${stringifyDeterministically(options)}` : `t:${key}`;

/**
 * Used to overwrite the default `intl/missing-message` implementation in order
 * to display a deterministic serialization of the translation for easier
 * assertions in the tests.
 *
 * @param key translation key
 * @param locales list of locales to search through
 * @param options options and variables passed along
 */
const missingMessage = (key: string, locales: string[], options: object) =>
  serializeTranslation(key, options);

/**
 * Picks the last locale from a locales array. In case of a string value,
 * returns that string.
 *
 * @param locales
 */
const pickLastLocale = (locales: string | string[]): string =>
  last(castArray(locales));

/**
 *
 *
 * @param hooks
 * @param translations
 */
export default function(hooks, translations = {}) {
  hooks.beforeEach(function() {
    this.owner.register('util:intl/missing-message', missingMessage);
    this.t = serializeTranslation;

    const locale = 'de-de';
    this.intl = this.owner.lookup('service:intl');
    this.intl.setLocale(locale);

    /**
     * Convenience util for adding translations to the last enabled locale.
     *
     * @param translations
     */
    this.addTranslations = translations =>
      this.intl.addTranslations(
        pickLastLocale(get(this.intl, 'locale')),
        translations
      );

    this.addTranslations(translations);
  });
}
```

```js
module('some test', function(hooks) {
  setupTest(hooks);
  setupIntl(hooks);

  test('some test' , async function(assert) {
    await render(hbs`{{t 'some.translation' foo="bar"}}`); // => t:some.translation:{"foo":"bar"}

     assert.dom().hasText(this.t('some.translation', { foo: 'bar' }));
  });
});
```

addon/utils/missing-message.js
```diff
@@ -11,6 +11,14 @@ export default function missingMessage(key, locales) {
     return `No locale defined.  Unable to resolve translation: "${key}"`;
   }
 
+  if (options.default.length > 0) {
```
jasonmit: Can you update https://github.com/ember-intl/ember-intl/blob/master/docs/missing-translations.md to cover this change

buschtoens:
> Understandable. This is based on the ember-i18n API so maybe we can encourage change there.

Sure, I know. Do you think it's worth it though?

> I'm not afraid of releasing another major version.

Great news! Just thought it might be a bit silly tonl release another major version just a few days after the other major version was released. Haha. 

Anyhow. Are you happy with the semantics `missing-message` or is there anything you'd like to change, seeing that we can do a major release anyway?

jasonmit:
> Are you happy with the semantics missing-message or is there anything you'd like to change, seeing that we can do a major release anyway?

I'm debating if the missing translation is doing too much

snewcomer:
I do like the test helper!  Really cool.  What is the use case for deterministically mocking the translation?  Is it if you deliver your translations over the wire and don't have a test fallback?

Also, just my 2 cents - I kind of view the `util` as a sane blueprint for the end user.  For example, I just override it and return out a key with `**${key}**` so I don't let any missing keys by but others might find these additions quite useful.  With these additions, it *seems* a bit more sane.  Although I would imagine some might want resilient to not be an option in dev/test (caveat - I'm not familiar with the history of this option), which in that case, they can just override this util with this as a blueprint to reference.

jasonmit:
> Anyhow, to get this out now before another major release and enable the test helper (PR still offered, if you like it), we can also deliver this in smaller iterations. For now, I can submit another PR which just passes through the options hash.

I would love that (& the test helpers!).

Thanks a lot for seeing this through, I could see the test helpers being really useful.

> While we load them asynchronously in dev mode, we don't want to depend on them in testing.

`treeForTranslations` could be used here if you wanted to write them to assets or bundle your translations from your third-party API.  @cibernox and & I recently paired on a similar solve for this which is what inspired the hook.  The downside being a translation change would require a deployment.  The plus side, your app is entirely vendored/immutable and your tests should just work (tm).

PR #588, opened 2018-06-26
feat: add failOnMissingTranslation and allowMissingTranslation option for additional missing translations handling

### LINKED ISSUE #547: Ignore certain "missing message" warnings

makepanic:
fixes #547 and #553

Please don't merge this yet. I'll add tests and documentation once we're ok with the general concept.
This PR introduces two new `config/ember-intl.js` fields:

* `failOnMissingTranslation`: boolean
  * this causes the build pipeline to throw an error if a missing translation was found
* `allowMissingTranslation`: (key, locale) => boolean
  * this allows to whitelist some missing translations

Everything is open for discussion.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,9 @@ module.exports = {
       wrapEntry: _bundlerOptions.wrapEntry,
       log() {
         return addon.log.apply(addon, arguments);
-      }
+      },
+      requiresTranslation: this.opts.requiresTranslation,
+      throwMissingTranslations: this.opts.throwMissingTranslations
     });
   },
 
@@ -169,7 +171,9 @@ module.exports = {
         disablePolyfill: false,
         autoPolyfill: false,
         inputPath: 'translations',
-        outputPath: 'translations'
+        outputPath: 'translations',
+        throwMissingTranslations: false,
+        requiresTranslation: (/* key, locale */) => true
       },
       this.readConfig(environment)
     );
--- a/blueprints/ember-intl/files/config/ember-intl.js
+++ b/blueprints/ember-intl/files/config/ember-intl.js
@@ -58,6 +58,28 @@ module.exports = function(/* env */) {
      * @type {String}
      * @default "translations"
      */
-    inputPath: 'translations'
+    inputPath: 'translations',
+
+    /**
+     * cause a build error if missing translations are detected.
+     *
+     * See https://github.com/jasonmit/ember-intl/blob/master/docs/missing-translations.md#throwing-a-build-error-on-missing-required-translation
+     *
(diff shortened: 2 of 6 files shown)


makepanic:
I've made some changes to make it easier to test.

Also logging is disabled if verbose but `failOnMissingTranslation` is set to true. 
I think it's pretty useless to log and fail at the same time as both contain somewhat the same information.

index.js
```diff
@@ -169,7 +171,9 @@ module.exports = {
         disablePolyfill: false,
         autoPolyfill: false,
         inputPath: 'translations',
-        outputPath: 'translations'
+        outputPath: 'translations',
+        failOnMissingTranslation: false,
```
jasonmit: `allowMissingTranslation() { return false; }` could be the same as `failOnMissingTranslation: true` - correct?

makepanic: Any suggestions on how to improve this?
Right now the missing translation predicate also causes log to not emit missing translations if it returns true.

jasonmit: Let me give it some thought and get back!

jasonmit: Thoughts on combining these two into a single method `onMissingTranslation` where by default it will log but if the user implements this hook, they'll be able to `throw` or conditionally log in the case they want to fail on missing translations.

That should solve both of the issues, let me know if you disagree!

makepanic: I'm not sure if it's a good idea to let the user build the error to throw.

I'm also not sure how this will work if a user wants to throw all missing translations. If the method is called for each missing translation then a user has to collect all missing translations and throw in the end.

Maybe we could rename the options to make it more clear?
e.g. `requiresTranslation(key, languages)` and `throwMissingTranslations`.

jasonmit: @makepanic good feedback.  I'm on board with the renaming.

makepanic:
I've rebased with changed option names

jasonmit:
@makepanic LGTM.  One last thing, can you update the docs to include the new APIs?

makepanic:
Sure, I'll update them

makepanic:
I've rebased onto current master and added some documentation.

jasonmit:
Thank you @makepanic.  I'll version soon!

PR #590, opened 2018-06-27
New test helpers

buschtoens:
As discussed in #587.

- [x] even moaaar tests
- [x] docs


buschtoens:
Alright, got it working. 🎉
I just need to add some more tests and docs now.

Are you happy with this API? I had to alter the translation serialization logic a bit, because a serialized translation can obviously not contain `{` and `}` as these would get picked up by the message formatter. I've just replaced them with `(` and `)` now.

```js
import { setupIntl, t } from 'ember-intl/test-support';
 
module('Integration | Test Helpers', function(hooks) {
  setupRenderingTest(hooks);
 
  module('setupIntl(hooks)', function(hooks) {
    setupIntl(hooks);
 
    test('hooks were properly executed and helpers work', async function(assert) {
      assert.strictEqual(this.intl, this.owner.lookup('service:intl'), '`this.intl` shorthand is available');
 
      assert.strictEqual(
        this.intl.t('some.translation'),
        't:some.translation:()',
        '`t` method serializes translation without variables'
      );
      assert.strictEqual(
        this.intl.t('some.translation', { foo: 'bar' }),
        't:some.translation:("foo":"bar")',
        '`t` method serializes translation with variables'
      );
 
      assert.strictEqual(
        this.intl.t('some.translation'),
        t('some.translation'),
        '`t` helper serializes translation without variables'
      );
      assert.strictEqual(
        this.intl.t('some.translation', { foo: 'bar' }),
        t('some.translation', { foo: 'bar' }),
        '`t` helper serializes translation with variables'
      );
    });
  });
});
```

buschtoens:
What do you think about `this.intl`? On the one hand it's nice to have quick access to the service, on the other hand we're polluting the context here. I could add a `getIntl()` helper instead.

But since [`ember-cli-mirage` is doing it](https://github.com/samselikoff/ember-cli-mirage/blob/690182f81baad538ac5ff7faee7d5eb1ce1fe769/addon-test-support/setup-mirage.js#L19), I think we're in the clear 😄

buschtoens:
btw, the CI failures [appear to be unrelated](https://twitter.com/buschtoens/status/1012012108605779968), since it's just yarn having issues with the registry. *Apparently.*

buschtoens:
Done 😊🏁

jasonmit:
> What do you think about this.intl? On the one hand it's nice to have quick access to the service, on the other hand we're polluting the context here. I could add a getIntl() helper instead.

I'm fine with this, we also do something similar within pollyjs.

jasonmit:
I'll review tonight, thanks again :)

docs/integration-testing.md
```diff
@@ -2,35 +2,238 @@
...
+import { setupIntl } from 'ember-intl/test-support';
 
-module('Integration | Component | x-product', function(hooks) {
+module('setupIntl demo', function(hooks) {
   setupRenderingTest(hooks);
+  setupIntl(hooks);
+
+  test('it serializes missing translations and injects the `intl` service', async function(assert) {
+    await render(hbs`{{t "some.translation" someVariable="Hello"}}`);
+    assert.dom().hasText('t:some.translation:("someVariable":"Hello")');
```
jasonmit: Love this

jasonmit:
This is great, thanks for landing all of the test-helpers + the missing-message changes!  I'll version later on tonight.

jasonmit:
Fixes #579

PR #593, opened 2018-07-10
fix(index): ensure that `requiresTranslation` option is a function

### LINKED ISSUE #592: default ember-intl blueprint throws an error
With the current ember-intl@3.1.0 there's a bug where `requiresTranslation` is not a function.
This causes the build to fail unitl a user provides a `requiresTranslation` function.

The issue is probably caused by [`Object.assign(defaults, provided)`](https://github.com/ember-intl/ember-intl/blob/master/index.js#L167) overwriting the default `requiresTranslation`.

One possible solution would be to check that addonConfig.requiresTranslation is a function and if not, overwrite it with the default function.

(Sorry for introducing that bug :) )

I can open a PR with the possible solution mentioned above.

Alternatively we could comment the `requiresTranslation` option from the blueprint to unset it.

makepanic:
fixes #592

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,17 @@ const utils = require('./lib/utils');
 const buildTree = require('./lib/broccoli/build-translation-tree');
 const TranslationReducer = require('./lib/broccoli/translation-reducer');
 
+const defaultConfig = {
+  locales: null,
+  publicOnly: false,
+  disablePolyfill: false,
+  autoPolyfill: false,
+  inputPath: 'translations',
+  outputPath: 'translations',
+  throwMissingTranslations: false,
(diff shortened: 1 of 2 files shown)


makepanic:
I'm unsure if we should log or not.
If a user consumes ember-intl and uses the default configuration, it'll always log the warning.

buschtoens:
> Maybe we should comment the requiresTranslation option from the config blueprint to avoid this.

That would be great. Just `//` and maybe a add another `// when defined, this must be a function`, if not already specified in the blueprint docs comment.

buschtoens:
I'll add it myself :)

PR #600, opened 2018-08-18
chore: Converts internals of ember-intl to es6 classes

jasonmit:


jasonmit:
Unsure why CI is failing, tests pass locally.  I'll need to dig into that.

buschtoens:
I'm traveling right now, but will give this a thorough review on Monday. But LGTM so far 😍

snewcomer:
Wow really nice work!  I think ppl can use this PR is a great blueprint for others looking to update addons!

Issue #607, opened 2018-09-13
Loading async translations with JSONP instead of XHR


estshy:
I have a website running on multiple domains with several languages, which are loaded asynchronously depending on context. I am working on improving application boot time and I can see that loading of translations through XHR is currently a bottleneck as such request is blocking everything else (deferReadiness used). I was wondering whether it would not be possible to go with JSONP instead. It would allow to start loading of the translations even before ember was be fetched. 

I also had a few difficulties with proper handling on fingerprint on translation files as ember does not provide access to fingerprint at runtime. 

Why does ember intl comes with XHR mechanism built in? Were there any advantages that I am overlooking right now?

buschtoens:
> Why does ember-intl comes with XHR mechanism built in? Were there any advantages that I am overlooking right now?

ember-intl doesn't come with XHR built in. It is entirely agnostic to how you load translations, if you don't bundle them in.

The docs on [asynchronously loading translations](https://github.com/ember-intl/ember-intl/blob/2.x/docs/asynchronously-loading-translations.md) demonstrate that:

```js
// app/routes/application.js
export default Ember.Route.extend({
  intl: Ember.inject.service(),
  async beforeModel() {
    let translations = await fetch('/api/translations.json');
    this.get('intl').addTranslations('en-us', translations);
    this.get('intl').setLocale('en-us');
  }
});
```

[`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is native browser API. Instead of using `fetch` to asynchronously load the translations during app boot, you could instead inject `<script>` tags into your HTML, that load your translations in a JSONP manner.

Depending on your supported browser targets, you could also use [`<link rel="preload" href="/api/translations.json" as="fetch">`](https://stackoverflow.com/a/46536558) to preload the JSON request on page load.

estshy:
@buschtoens Unfortunately the approach with link preload as fetch doesn't work at the moment. I am not sure whether preload as fetch is already widely supported.

![image](https://user-images.githubusercontent.com/5459938/45482059-5880c900-b74d-11e8-9a06-22ed17c19191.png)

```
<link rel="preload" href="http://localhost:4200/assets/translations/pl-pl.json" as="fetch">
```

```
    fetch(path)
    .then(response => response.json())
    .then((response) => {
      container.register('website:translations', response, {
        instantiate: false
      });
    })
    .finally(() => container.advanceReadiness())
```

buschtoens:
[...]

PR #608, opened 2018-09-13
Making it possible to specifcy custom handler for wrapEntry and filename

estshy:
In relation to a need described in issue (https://github.com/ember-intl/ember-intl/issues/607) I've created a possibility to specify custom handler for `wrapEntry` and `filename` methods used inside translation reducer.

Thanks to them one can achieve jsonp compatibile output from ember intl

```
filename: function filename(key) {
    return key + '.js';
},

wrapEntry: function wrapEntry(obj) {
    return 'loadTranslations(' + stringify(obj) + ');';
}
```

### DIFF
--- a/config/ember-intl.js
+++ b/config/ember-intl.js
@@ -69,6 +69,22 @@ module.exports = function(/* environment */) {
      * @type {Function?}
      * @default "function(){return true}"
      */
-    requiresTranslation: undefined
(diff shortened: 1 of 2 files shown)


jasonmit:
Tests failing but still would like to continue this conversation is #607 so I better understand the use-case.

PR #612, opened 2018-09-18
isDevelopingAddon is ignored

jlami:
## Summary
Use the addon that has translations to decide whether to use WatchedDir or UnwatchedDir based on `isDevelopingAddon()` (via `addon.treeGenerator` instead of `emberCliAddon.treeGenerator`)

## Reproduction
Make an application that uses an addon that is set to `isDevelopingAddon() { return true; }` with translations directory inside. Or for example an in-repo-addon: `ember generate in-repo-addon addon-with-translations`.
If you do `ember s` now, modifications in the addon translations are not Watched. (maybe again windows only?)

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,7 @@ module.exports = {
   },
 
   generateTranslationTree(bundlerOptions) {
-    const translationTree = buildTree(this.project, this.opts.inputPath, this.treeGenerator.bind(this));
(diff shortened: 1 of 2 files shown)


jasonmit:
Thanks! Versioned as 3.2.6

jlami:
3.2.6 works! Thanks!

PR #614, opened 2018-09-25
feat(translationMacro): auto-lookup `intl` service, if not injected

### LINKED ISSUE #613: Make `translationMacro` work without `intl` service injection
Right now the [Computed Property macro](https://github.com/ember-intl/ember-intl/blob/e664d0eb8fbe9dc47c44e9d1df38eb2defa81416/docs/translating-text.md#computed-property-macro) relies on the `intl` service to be injected on the host object.

https://github.com/ember-intl/ember-intl/blob/e664d0eb8fbe9dc47c44e9d1df38eb2defa81416/docs/translating-text.md#L36-L67

We _could_ work around this by looking up the `intl` service on the owner, if there is no `intl` service injected.

https://github.com/ember-intl/ember-intl/blob/e664d0eb8fbe9dc47c44e9d1df38eb2defa81416/addon/macro.js#L25-L38

`L30-L34` would be changed into:

```js
let intl = get(this, 'intl');
if (!intl) {
  const owner = getOwner(this);
  assert(
    `Cannot translate "${key}".\n${this} does not have an 'intl' property set or an owner to look up the service from.`,
    owner
  );

  intl = owner.lookup('service:intl');
  assert(
    `Cannot translate "${key}".\n${this} does not have an 'intl' property set and there is no 'intl' service registered with the owner.`,
    intl
  );
}
```

This way, the injected `intl` service still takes precedence (in case users are doing some non-standard magic), but the macro will fallback to the global `intl` service for easier usage.

buschtoens:
Closes #613.


buschtoens:
🎉

Issue #615, opened 2018-09-28
Provide ability to know when a translation is missing for specific locale


artinboghosian:
- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.1.4
- Ember CLI Version: 3.1.4
- Ember Intl Version: 3.2.0
- Browser(s):
- Node Version: 8.11.3

### Steps to Reproduce

I am trying to add the ability to notify whenever a translation is missing for a particular locale. In production we use a fallback strategy which loads multiple locales including our default of en-US. This means we cannot use missing-message since that is only called if no translation is found from any locale. I want to be able to send a notification when a locale does not have a translation but in the currently implementation of intl service and default adapter this is not possible.

We can simply add a hook to the default adapter that is called whenever a translation cannot be found for a locale: 

```
if (model && model.has(translationKey)) {
  return model.getValue(translationKey);
} else {
  // provide hook here
}
```

That feels a bit odd to me and i think maybe the adapter should simply provide the functionality to lookup a translation for one key and one locale. Then the logic for iterating through multiple locales would be provided at a higher level, such as the service. And in this way you can call missing-message every time a translation is not found for a specific locale. Or another concept can be added that means something like missing-message-for-locale not necessarily that you couldn't not find any translation.

Let me know your thoughts.

jasonmit:
> That feels a bit odd to me and i think maybe the adapter should simply provide the functionality to lookup a translation for one key and one locale.

Should be a relatively easy change to make without breaking public APIs.

> And in this way you can call missing-message every time a translation is not found for a specific locale.

Missing message is responsible for returning the placeholder text, i.e., translation not found along with emitting a console warning.  Struggling to see how we can invoke this on every missing locale.  Seems sensible to stop once all the locales are iterated through.

>  Or another concept can be added that means something like missing-message-for-locale not necessarily that you couldn't not find any translation.

Might solve my concern in the paragraph above but would be curious to see what that looks like.

Are you more interested in an event?

artinboghosian:
[...]

PR #617, opened 2018-10-08
Adapter should only lookup key for specific locale whereas service sh…

artinboghosian:
…ould iterate active locales to find translation

https://github.com/ember-intl/ember-intl/issues/615


artinboghosian:
No problem, thanks for awesome library!

PR #624, opened 2018-10-16
feat: translation decorator `@t`

### LINKED ISSUE #623: Provide @t decorator
Please provide a `@t` decorator. Just for convenience not to use the `@macro` helper everytime. Would be a good to be served by ember-intl directly.

buschtoens:
This enables users to:

```ts
import Component from '@ember/component':
import { service } from '@ember-decorators/service';
import { translationDecorator as t } from 'ember-intl';

export default class ExampleComponent extends Component {
  // Injecting the service is optional
  @service intl;

  // A simple translation.
  @t('user.edit.title') title;
  followersCount = 1;

  // A translation with interpolations. This computed property
  // depends on `count` and will send `{ count: this.followersCount }`
  // in to the translation.
  @t('user.followers.title', { count: 'followersCount' })
  followersTitle;
}
```

Closes #623.

This is the naive and easy implementation. The problem is that we now throw the full `@ember-decorators/object` package onto every ember-intl user, which I do not really feel comfortable with.

Things we could do:

- create a separate `ember-intl-decorators` package 😒
- lift `macro` out of `@ember-decorators/object` and maybe put it in `@ember-decorators/utils`?
- try setting `@ember-decorators/object` as a `peerDependency` and tell users that it works, _if_ they install `@ember-decorators/object`
- manually create the decorator

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2017,
     sourceType: 'module'
--- a/addon/decorator.js
+++ b/addon/decorator.js
@@ -1,0 +1,4 @@
+import { macro } from '@ember-decorators/object/computed';
+import translationMacro from './macro';
+
+export default macro(translationMacro);
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,4 +1,5 @@
 import Service from './services/intl';
 import translationMacro from './macro';
+import translationDecorator from './decorator';
 
-export { Service, translationMacro };
+export { Service, translationMacro, translationDecorator };
--- a/docs/translating-text.md
+++ b/docs/translating-text.md
@@ -70,3 +70,30 @@ injection, the macro will lookup the `intl` service via the owner of the host
 object.
 
 _This feature is based on ember-i18n's `translationMacro`._
+
+## Computed Property Decorator
+
+If you are using ES6 class syntax, you can't easily use computed property macros.
+This is why ember-intl also provides the exact same functionality as a decorator.
+
+```js
+import Component from '@ember/component':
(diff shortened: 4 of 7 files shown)


jasonmit:
Have you considered introducing an API that allows for separating string literals and path keys?

Perhaps something like:

```
import Component from '@ember/component':
import { service } from '@ember-decorators/service';
import { translationDecorator as t, key } from 'ember-intl';

export default class ExampleComponent extends Component {
  // Injecting the service is optional
  @service intl;

  // A translation with interpolations. This computed property
  // depends on `count` and will send `{ count: this.followersCount }`
  // in to the translation.
  @t('user.followers.title', { count: key('followersCount') })
  followersTitle;
}
```

buschtoens:
Interesting concept.

This is pretty much the same, but the exact inverse of how [`raw`](https://github.com/kellyselden/ember-macro-helpers#raw) in [`ember-awesome-macros`](https://github.com/kellyselden/ember-awesome-macros) / [`ember-macro-helpers`](https://github.com/kellyselden/ember-macro-helpers) works.

Do you think having static inputs is more common than dynamic ones?

> I can make you admin of the org if you already don't have it.

I'm not yet. 😉 🍻

gossi:
I would expect, that `@t('user.followers.title', { count: 'followersCount' })` does not look up the object with the property `followersCount` - to me this is very, very legacy ember behavior and actually doesn't align with embers clarity movement. I'm actually more in favor of @jasonmit approach here tbh. Although I just read, that what @buschtoens proposed is the current ember-intl behavior - still don't like that kind of "magic" API 😝

buschtoens:
Published a first draft of [`ember-intl-decorators`](https://github.com/ember-intl/ember-intl-decorators#readme) to [npm](https://www.npmjs.com/package/ember-intl-decorators). 🚀

PR #626, opened 2018-10-17
feat(translation-macro): add `raw` literals

### LINKED ISSUE #625: Allow literal / raw values in translation macro / decorator
As suggested in https://github.com/ember-intl/ember-intl/pull/624#issuecomment-430397615.

The translation macro (and soon decorator #623) currently only allow the user to pass keys that will then be used to retrieve the values to be used for interpolating the translation strings: [Docs](https://github.com/ember-intl/ember-intl/blob/57144a6bb9a6f9e64f74273177f250a6d6d8010c/docs/translating-text.md#computed-property-macro)

```js
import Component from '@ember/component':
import { translationMacro as t } from 'ember-intl';

export default Component.extend({
  followersCount: 1,

  // A translation with interpolations. This computed property
  // depends on `count` and will send `{ count: this.followersCount }`
  // in to the translation.
  followersTitle: t('user.followers.title', { count: 'followersCount' })
});
```

Assume the `count` variable is static. We can't pass `{ count: 1 }`, instead we need to go through one unnecessary step of redirection by setting `followersCount: 1` and then passing `{ count: 'followersCount' }`.

We propose adding a [`raw` function](https://github.com/kellyselden/ember-macro-helpers#raw) like in [`ember-awesome-macros`](https://github.com/kellyselden/ember-awesome-macros) / [`ember-macro-helpers`](https://github.com/kellyselden/ember-macro-helpers):

```js
import Component from '@ember/component':
import { translationMacro as t, raw } from 'ember-intl';

export default Component.extend({
  // A translation with static interpolations. This computed property
  // has no dependencies (other than the implicit `intl.locale`)
  // and will send `{ count: 1 }`
  followersTitle: t('user.followers.title', { count: raw(1) })
});
```

buschtoens:
Closes #625.

### DIFF
--- a/addon/macro.js
+++ b/addon/macro.js
@@ -3,29 +3,72 @@
  * https://github.com/jamesarosen/ember-i18n/blob/master/addon/utils/macro.js
  */
 import { assert } from '@ember/debug';
-import { computed, get } from '@ember/object';
+import { computed, get, getProperties } from '@ember/object';
 import { getOwner } from '@ember/application';
 import EmptyObject from 'ember-intl/-private/empty-object';
 
-const { keys } = Object;
+function partitionDynamicValuesAndStaticValues(options) {
+  const dynamicValues = new EmptyObject();
+  const staticValues = new EmptyObject();
 
-function values(object) {
-  return keys(object).map(key => object[key]);
+  Object.keys(options).forEach(key => {
+    const value = options[key];
+    if (value instanceof Raw) {
+      staticValues[key] = value;
+    } else {
+      dynamicValues[key] = value;
+    }
+  });
+
+  return [dynamicValues, staticValues];
 }
 
 function mapPropertiesByHash(object, hash) {
-  const result = {};
+  const result = new EmptyObject();
 
-  keys(hash).forEach(function(key) {
+  Object.keys(hash).forEach(key => {
     result[key] = get(object, hash[key]);
   });
 
   return result;
 }
 
+/**
+ * This class is used to box primitive types and mark them as raw literals that
+ * should be used as is by the translation macro.
+ *
+ * This class is internal. Instead of using this class directly, use the `raw`
+ * utility function, that creates an instance of this class.
+ */
+class Raw {
+  constructor(value) {
+    this._value = value;
+  }
+
+  valueOf() {
+    return this._value;
+  }
+
+  toString() {
+    return String(this._value);
+  }
+}
+
+/**
+ * Use this utility function to mark a value as a raw literal.
+ *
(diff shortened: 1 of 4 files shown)


buschtoens:
Whoops, forgot the docs.

addon/macro.js
```diff
@@ -3,29 +3,72 @@
...
+      dynamicValues[key] = value;
+    }
+  });
+
+  return [dynamicValues, staticValues];
 }
 
 function mapPropertiesByHash(object, hash) {
-  const result = {};
+  const result = new EmptyObject();
```
buschtoens: Is `EmptyObject` actually faster? I always thought that modern browsers have optimized the :poop: out of `{}`. 😄

addon/macro.js
```diff
@@ -3,29 +3,72 @@
...
+function partitionDynamicValuesAndStaticValues(options) {
+  const dynamicValues = new EmptyObject();
+  const staticValues = new EmptyObject();
 
-function values(object) {
-  return keys(object).map(key => object[key]);
+  Object.keys(options).forEach(key => {
+    const value = options[key];
+    if (value instanceof Raw) {
+      staticValues[key] = value;
```
buschtoens: Could potentially already eagerly call `value.valueOf()` here to allow the `Raw` instance to be garbage collected and avoid repeated (but lazy) type coercion every time the macro recomputes.

addon/macro.js
```diff
@@ -3,29 +3,72 @@
...
 import EmptyObject from 'ember-intl/-private/empty-object';
 
-const { keys } = Object;
+function partitionDynamicValuesAndStaticValues(options) {
+  const dynamicValues = new EmptyObject();
+  const staticValues = new EmptyObject();
 
-function values(object) {
-  return keys(object).map(key => object[key]);
+  Object.keys(options).forEach(key => {
```
buschtoens: Intentionally using this old clunky syntax instead of `for (const [k, v] of Object.entries(...))` here to avoid transpilation bloat for ancient browsers.

addon/macro.js
```diff
@@ -44,6 +87,6 @@ export default function createTranslatedComputedProperty(key, options) {
       );
     }
 
-    return intl.t(key, mapPropertiesByHash(this, hash));
+    return intl.t(key, Object.assign({}, staticValues, mapPropertiesByHash(this, dynamicValues)));
```
buschtoens: `Object.assign` is not natively available in IE11. I guess we should use [`assign` from `@ember/polyfills`](https://emberjs.com/api/api/ember/release/functions/@ember%2Fpolyfills/assign) instead.

PR #649, opened 2018-10-23
feat: set document `lang` attribute

### LINKED ISSUE #533: Improve screen reader support
The Lighthouse report often shows the following warning:

> `<html>` element does not have a `[lang]` attribute.

If an Ember app only has a single language then we can set that language in the `index.html`, but when supporting multiple languages it gets a little more tricky. I would like to propose doing something like this:

```js
setLocale(locale) {
  // ...
  document.documentElement.lang = locale;
}
```

This should result in something like `<html lang="de">` and will hopefully improve the Lighthouse score and screen reader support.

buschtoens:
Closes #533.

Based on the code provided by @Turbo87 in https://github.com/ember-intl/ember-intl/issues/533#issuecomment-392483328


addon/services/intl.js
```diff
@@ -38,7 +40,7 @@ export default Service.extend(Evented, {
 
       if (!isArrayEqual(proposed, this._locale)) {
         this._locale = proposed;
-        this.trigger('localeChanged');
+        next(() => this.trigger('localeChanged'));
```
Turbo87: why this change?

addon/services/intl.js
```diff
@@ -206,7 +208,17 @@ export default Service.extend(Evented, {
     if (Array.isArray(localeName)) {
       return localeName.map(normalizeLocale);
     }
-  }
+  },
+
+  /** @private **/
+  updateDocumentLanguage: on('localeChanged', function() {
```
Turbo87: why do you use the `Evented` system for this instead of calling `updateDocumentLanguage()` directly from the setter?

addon/services/intl.js
buschtoens: The locale does not actually change until _after_ this setter has run through. So emitting this event _during_ the setter causes all (synchronous) listeners accessing the `locale` property to receive the old value.

The reason that it works in the `{{t}}` helper is that [`recompute`](https://github.com/emberjs/ember.js/blob/v3.5.0/packages/ember-glimmer/lib/helper.ts#L90-L119) marks the helper as dirty and it's only recomputed later in a different run loop queue.

https://github.com/ember-intl/ember-intl/blob/89009718cb36ebfce0b2783e9aff05702d9f6431/addon/helpers/-format-base.js#L22

I only noticed this "bug" now, because I used the event further down in the class.

Turbo87: this should be an independent bugfix PR then, right? I guess it would also be good to have some sort of test for this so that we don't regress on this again.

addon/services/intl.js
buschtoens: Because it was already there and (in theory) usable. Assuming that `localeChanged` was implemented correctly in the first place, this would have been a "minimally invasive" addition. 😄 

Doing it this way also allows us to potentially add further listeners, without having to explicitly list them all.

But on second thought, I think you are right. The `Evented` thing is foreseen to be deprecated at some point (AFAIK) and the order the listeners would wbe called in is not explicitly defined.

I'll refactor it.

addon/services/intl.js
buschtoens: I'll pull it out. 👌

jasonmit:
@Turbo87 correct, no fastboot tests at the moment.

jasonmit:
@buschtoens I made some adjustments to fix the failing test and rebased.  Merge whenever you're ready.

buschtoens:
Awesome, thanks! :)

PR #650, opened 2018-10-23
fix(intl): emit localeChanged event *after* locale has changed

buschtoens:
Split from https://github.com/ember-intl/ember-intl/pull/649

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -31,7 +31,7 @@ export default Service.extend(Evented, {
 
   /** @public **/
   formats: null,
-  
+
   _timer: null,
 
   /** @public **/
@@ -104,7 +104,7 @@ export default Service.extend(Evented, {
 
     hydrate(this, this._owner);
   },
-  
+
   willDestroy() {
     this._super(...arguments);
     cancel(this._timer);
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -14,7 +14,7 @@ import { assert, warn } from '@ember/debug';
 import { makeArray } from '@ember/array';
 import { assign } from '@ember/polyfills';
 import Service from '@ember/service';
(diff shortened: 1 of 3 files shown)


tests/unit/services/intl-test.js
```diff
@@ -238,4 +240,16 @@ module('service:intl', function(hooks) {
...
     assert.equal(this.intl.exists('bar'), false);
   });
+
+  test('changing the locale emits the `localeChanged` event and the new locale is available', async function(assert) {
+    assert.expect(1);
+    
+    const newLocale = ['de', 'en-us'];
+
+    addListener(this.intl, 'localeChanged', () => {
+      assert.deepEqual(get(this.intl, 'locale'), newLocale);
```
jasonmit: @buschtoens this gets called twice where I'd expect it to be called once.  Could be an artifact of setting the locale during the test setup, didn't get a chance to pull down the branch and investigate.

addon/services/intl.js
```diff
@@ -38,7 +39,7 @@ export default Service.extend(Evented, {
 
       if (!isArrayEqual(proposed, this._locale)) {
         this._locale = proposed;
-        this.trigger('localeChanged');
+        next(() => this.trigger('localeChanged'));
```
jasonmit: Should we cancel on destroy?

jasonmit: Can you help me understand this issue of needing to schedule on the next runloop.  `localeChanged` is relied on in the helpers to recompute on locale change.  Is that behavior broken?

https://github.com/ember-intl/ember-intl/blob/89009718cb36ebfce0b2783e9aff05702d9f6431/addon/helpers/-format-base.js#L22

buschtoens:
@jasonmit Thanks a lot for taking this over! ❤️

PR #653, opened 2018-10-24
feat: remove null translation values

### LINKED ISSUE #637: Filter empty translation keys to reduce file size
> ```
> # en-us.yml
> sample_translation:
>     translation_1: Lorem ipsum
>     translation_2: dolor sit amet
> ```
> ```
> # de-de.yml
> sample_translation:
>     translation_1: Lorem ipsum
>     translation_2: 
> ```

`translation_2` would be `null` for `de-de`.

As suggested in https://github.com/ember-intl/ember-intl/issues/598#issuecomment-413779400:

> Maybe it would make more sense to strip these `null` keys during build time, since they unnecessarily inflate the size of the translations bundle?

Which would prevent this error https://github.com/ember-intl/ember-intl/issues/598#issue-351179851:

> In this case while using German translations within the application, when we encounter `sample_translation.translation_2` an error ‚A message must be provided as a String or AST.’ will be thrown and app will break.

https://github.com/ember-intl/ember-intl/blob/c92422d416de65db3fb70f68c57fec9b782dc40c/lib/broccoli/translation-reducer.js#L167-L173

Before passing `translations` to `findMissingTranslations`, `null` values should be removed.

makepanic:
fixes #637

### DIFF
--- a/lib/broccoli/translation-reducer.js
+++ b/lib/broccoli/translation-reducer.js
@@ -20,6 +20,7 @@ const fs = require('fs');
 const propKeys = require('../utils/prop-keys');
 const forEachMessage = require('../utils/for-each-message');
 const validateMessage = require('../utils/validate-message');
+const stripNestedNulls = require('../utils/strip-nested-nulls');
 
 function readAsObject(filepath) {
   let data = fs.readFileSync(filepath);
@@ -114,6 +115,10 @@ class TranslationReducer extends CachingWriter {
       }
 
       let translation = readAsObject(translationPath);
+
+      if (this.options.stripEmptyTranslations === true) {
+        translation = stripNestedNulls(translation);
+      }
 
       if (!translation) {
         this._log(`cannot read path "${translationPath}"`);
--- a/tests-node/unit/utils/strip-nested-nulls-test.js
+++ b/tests-node/unit/utils/strip-nested-nulls-test.js
@@ -1,0 +1,21 @@
+/* eslint-env node */
+/* globals describe, it */
+
+'use strict';
+
+let expect = require('chai').expect;
+
+let stripNestedNulls = require('../../../lib/utils/strip-nested-nulls');
+
+describe('stripNestedNulls', function() {
+  [
+    [{ a: true, b: null }, { a: true }],
+    [{ a: { b: null } }, { a: {} }],
+    [{ a: { b: null, c: '' }, d: '' }, { a: { c: '' }, d: '' }],
+    [{ a: true, b: { c: true, d: { e: null, f: true } } }, { a: true, b: { c: true, d: { f: true } } }]
+  ].forEach(([object, expected]) => {
+    it(`${JSON.stringify(object)} -> ${JSON.stringify(expected)}`, function() {
+      expect(stripNestedNulls(object)).to.deep.equal(expected);
+    });
+  });
+});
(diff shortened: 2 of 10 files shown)


lib/broccoli/translation-reducer.js
```diff
@@ -114,6 +115,7 @@ class TranslationReducer extends CachingWriter {
       }
 
       let translation = readAsObject(translationPath);
+      translation = stripNestedNulls(translation);
```
makepanic: I'm not sure if this is best done here or outside the `readAsObject` method.

Also this could be refactored to `stripNestedNulls(readAsObject(translationPath))` (need pipeline operator :cry: )

jasonmit:
Technically this is a breaking change, should we add a boolean config option with it defaulting to `false`.  In the next major release we'll flip it on.

I could see people intentionally leaving keys blank.

tests-node/unit/utils/strip-nested-nulls-test.js
```diff
@@ -0,0 +1,20 @@
...
+
+'use strict';
+
+let expect = require('chai').expect;
+
+let stripNestedNulls = require('../../../lib/utils/strip-nested-nulls');
+
+describe('stripNestedNulls', function() {
+  [
+    [{ a: true, b: null }, { a: true }],
```
jasonmit: Can you add a test ensure empty strings do not get stripped out

makepanic:
> Technically this is a breaking change, should we add a boolean config option with it defaulting to `false`. In the next major release we'll flip it on.
> 
> I could see people intentionally leaving keys blank.

Any suggestions on a configuration name? I'd call it `stripEmptyTranslations` for now. Do we also have to add a markdown documentation page or is it enough to be documented in the blueprint?

tests-node/unit/utils/strip-nested-nulls-test.js
makepanic: done

buschtoens:
> Do we also have to add a markdown documentation page or is it enough to be documented in the blueprint?

I wouldn't know what to name it and what content to put on it.
@jasonmit What's your take?

buschtoens:
I've renamed the PR to `feat` to stay semver compliant, since we are introducing a new config flag. 😊

jasonmit:
> I wouldn't know what to name it and what content to put on it.
> @jasonmit What's your take?

I think for now config/env + js comments will suffice.  If you all disagree, I can be easily convinced we should add the markdown file :)

jasonmit:
@buschtoens the commit still says `fix`, should we rewrite history in this case?

buschtoens:
@jasonmit I always squash merge and edit the commit message in these cases. For more complex PRs I `git rebase -i` and reword or ask the author to do it.\

Since were not officially enforcing [conventional commit messages](https://www.conventionalcommits.org/) (yet 😛 ), I have no problem with keeping it as is. I write the release notes manually anyway. 😄 

If you like I can install semantic releases, which would auto-release, as soon as you merge something that should trigger a release.

jasonmit:
> If you like I can install semantic releases, which would auto-release, as soon as you merge something that should trigger a release.

Lets stick with manually release for now

buschtoens:
Released as `v3.5.0`. Thanks again, @makepanic! 🎉

PR #675, opened 2018-11-07
Scope broccoli-cldr-data to ember-intl

snewcomer:
Blocked by https://github.com/ember-intl/broccoli-cldr-data/pull/10

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,11 +31,11 @@
     "prettier": "prettier --single-quote --print-width 120 --write \"{blueprints,config,lib,app,addon,addon-test-support,tests,tests-node}/**/*.js\""
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


snewcomer:
This also updates cldr data to v34 - `cldr-core "^34.0.0"` from v29 so might be a minor or even part of a major release.  Lmk your thoughts.

package.json
```diff
@@ -31,11 +31,11 @@
     "prettier": "prettier --single-quote --print-width 120 --write \"{blueprints,config,lib,app,addon,addon-test-support,tests,tests-node}/**/*.js\""
   },
   "dependencies": {
+    "@ember-intl/broccoli-cldr-data": "^3.0.0",
```
jasonmit: Yeah this is a breaking change (as you mentioned).

We'll need to publish a new major on ember-intl.

Issue #601, opened 2018-08-23
Warn user on ICU arguments mismatch


makepanic:
I think ember-intl could theoretically check if a translation string has the same ICU argument in all translations, similar on how we already do it for the existence of translation keys in all languages.

e.g.
```
# de.yaml
foo: Hallo {who}

# en.yaml
foo: Hello {whos}
```

Would warn (or throw if `throwMissingTranslations` is enabled) that `foo` has different ICU argument names during build time, to avoid runtime errors when rendering the mismatching translation strings.

I'm unsure if there is a good use-case on omitting ICU arguments in some languages, maybe we can also provide something like `requiresTranslation: () => true` when filtering these errors.

jasonmit:
Definitely is possible and honestly I don't think it would be difficult to implement.

If anyone wants to take this on, https://github.com/yahoo/intl-messageformat-parser is basically what you'll need to parse and compare the `argumentElement`'s between translations.

snewcomer:
This seems like an interesting problem to solve.  I'd be happy to tackle when I get some free time.

makepanic:
@snewcomer were you able to work on this? If not I think I could work on this over the weekend.

PR #683, opened 2018-11-11
feat: add throwICUArgumentMismatch option to throw on ICU argument mismatch

makepanic:
This implements ICU argument checking to solve #601.

I have a couple open questions:

* is `throwICUArgumentMismatch` ok as a name?
* should this be added to the docs?
* is it ok if this PR introduces a [mapping an array with side-effects](https://github.com/ember-intl/ember-intl/compare/master...makepanic:issues/601?expand=1#diff-8218ca25365bb38750c0455c805b9fbeR110)?
* should this also support a configurable filter function to allow ICU mismatch in some cases (similar to `requiresTranslation`)?


tests-node/unit/broccoli/translation-reducer-test.js
```diff
@@ -7,6 +7,14 @@ const subject = require('../../../lib/broccoli/translation-reducer');
 
 describe('translation-reducer', function() {
   beforeEach(function() {
+    this.icuFixture = {
+      de: {
+        foo: 'Hallo {who}',
```
jasonmit: Can you write a test that uses parameters.

`You have {numPhotos, plural, =0 {no photos.} =1 {one photo.} other {# photos.}}`

I'd expect it to not fail if the parameters are different, only should match on first param (key).

index.js
```diff
@@ -25,6 +25,7 @@ const defaultConfig = {
   autoPolyfill: false,
   inputPath: 'translations',
   outputPath: 'translations',
+  throwICUArgumentMismatch: false,
```
jasonmit: `throwICUArgumentMismatch` -> `errorOnNamedArgumentMismatch`

(Context: I want to go back and rename `throwMissingTranslations` to `errorOnMissingTranslations`)

jasonmit:
> is it ok if this PR introduces a mapping an array with side-effects?

I'm fine with that.

> should this be added to the docs?

I think the ember-intl config file itself can act as its own doc via comments for now. 

> should this also support a configurable filter function to allow ICU mismatch in some cases (similar to requiresTranslation)?

I wouldn't make that a requirement for this to land.

makepanic:
I've updated the PR

tests-node/unit/broccoli/translation-reducer-test.js
```diff
@@ -7,6 +7,18 @@ const subject = require('../../../lib/broccoli/translation-reducer');
...
+        photos: 'You have {numPhotos, plural, =0 {no photos.} =1 {one photo.} other {# photos.}}',
```
jasonmit: Can you add a test that this is valid:

```
en: {
  photos: 'You have {numPhotos, plural, =0 {no photos.} =1 {one photo.} other {# photos.}}'
},
de: {
  photos: 'You have {numPhotos, plural, =0 {foo} =1 {bar} other {# baz}}'
}
```

lib/broccoli/translation-reducer.js
```diff
@@ -82,30 +82,80 @@ class TranslationReducer extends CachingWriter {
...
 
-  findMissingTranslations(translations) {
+  extractICUArguments(string) {
+    return parser
+      .parse(string)
+      .elements.filter(element => element.type === 'argumentElement')
+      .map(argumentElement => argumentElement.id);
+  }
+
+  lintTranslations(translations) {
```
jasonmit: Can we break this function up into smaller fns?

makepanic:
I've added the valid argument test and tried to break the "linting" into smaller methods.
I've also rebased onto current master.

jasonmit:
Thanks again @makepanic :)  It will be in the next 4.x beta

PR #707, opened 2018-11-27
fix(macro): locale change with implicit intl service

### LINKED ISSUE #636: translationMacro without intl service injection does not recompute on locale change
https://github.com/ember-intl/ember-intl/pull/614 made the `intl` service injection optional, but `intl.locale` is still statically listed as a dependent key:

https://github.com/ember-intl/ember-intl/blob/c92422d416de65db3fb70f68c57fec9b782dc40c/addon/macro.js#L72

When not injecting the service, this of course does not work. Instead we then need to manually watch `intl.locale` and invalidate the computed property.

buschtoens:
Fixes #636.

- [x] ~~For now this just adds a failing test.~~
- [x] ~~Works 🚀~~
- [x] _...Works for all versions of Ember_ 😅 

@jasonmit Thanks for that awesome idea!

### DIFF
--- a/addon/macro.js
+++ b/addon/macro.js
@@ -3,9 +3,10 @@
  * https://github.com/jamesarosen/ember-i18n/blob/master/addon/utils/macro.js
  */
 import { assert } from '@ember/debug';
-import { computed, get } from '@ember/object';
+import { get, defineProperty } from '@ember/object';
+import ComputedProperty from '@ember/object/computed';
+import { inject as service } from '@ember/service';
 import { assign } from '@ember/polyfills';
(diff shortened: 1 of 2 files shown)


jasonmit:
Nice work, thanks for moving this forwards!

buschtoens:
I'll likely refactor this slightly and extract a `IntlComputedProperty` base class that implicitly injects the `intl` service, so that we can create a range of macros that extend from it. I am eyeing towards https://github.com/ember-intl/ember-intl/issues/392#issuecomment-431570182 here.

PR #708, opened 2018-11-28
refactor: create IntlComputedProperty base class

buschtoens:
As written in https://github.com/ember-intl/ember-intl/pull/707#issuecomment-442390143:

> I'll likely refactor this slightly and extract a `IntlComputedProperty` base class that implicitly injects the `intl` service, so that we can create a range of macros that extend from it. I am eyeing towards [#392 (comment)](https://github.com/ember-intl/ember-intl/issues/392#issuecomment-431570182) here.

### DIFF
--- a/addon/intl-computed-property.js
+++ b/addon/intl-computed-property.js
@@ -8,10 +8,7 @@ export default class IntlComputedProperty extends ComputedProperty {
     super(function(propertyKey) {
       let intl = get(this, 'intl');
 
-      assert(
-        `ember-intl: Could not look up 'intl' service for the property '${propertyKey}' on '${this}'.`,
-        intl
-      );
+      assert(`ember-intl: Could not look up 'intl' service for the property '${propertyKey}' on '${this}'.`, intl);
 
       return fn.call(this, intl, propertyKey, this);
     });
--- a/addon/intl-computed-property.js
+++ b/addon/intl-computed-property.js
@@ -4,13 +4,16 @@ import { inject as service } from '@ember/service';
 import { assert } from '@ember/debug';
 
 export default class IntlComputedProperty extends ComputedProperty {
-  constructor(fn, ...dependentKeys) {
+  constructor(...dependentKeysAndGetterFn) {
+    const getterFn = dependentKeysAndGetterFn.pop();
+    const dependentKeys = dependentKeysAndGetterFn;
+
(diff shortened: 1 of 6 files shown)


addon/intl-computed-property.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor(fn, ...dependentKeys) {
+    super(function(propertyKey) {
+      let intl = get(this, 'intl');
+
+      assert(
+        `ember-intl: Could not look up 'intl' service for the property '${propertyKey}' on '${this}'.`,
+        intl
+      );
+
+      return fn.call(this, intl, propertyKey, this);
```
buschtoens: Passing `this` as the third arguments allows derived classes to use arrow functions and still have access to the object instance the computed property is installed on.

addon/intl-computed-property.js
```diff
@@ -0,0 +1,34 @@
+import { get, defineProperty } from '@ember/object';
+import ComputedProperty from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+import { assert } from '@ember/debug';
+
+export default class IntlComputedProperty extends ComputedProperty {
+  constructor(fn, ...dependentKeys) {
```
buschtoens: I thought about moving the `...dependenKeys` in front, to mimic the invocation of [`computed`](https://www.emberjs.com/api/ember/3.5/functions/@ember%2Fobject/computed), but this would require some extra argument juggling in here.

I'm fine with it though.

addon/macro.js
```diff
@@ -67,37 +68,22 @@ export function raw(value) {
...
+    const dependentKeys = Object.values(dynamicValues);
+
+    super(
+      (intl, propertyKey, ctx) =>
+        intl.t(
+          translationKey,
+          assign({}, staticValues, mapPropertiesByHash(ctx, dynamicValues))
+        ),
+      ...dependentKeys
+    );
```
buschtoens: So considering https://github.com/ember-intl/ember-intl/pull/708/files#r237071761 this could instead also be:

```js
super(...dependentKeys, (intl, propertyKey, ctx) =>
  intl.t(
    translationKey,
    assign({}, staticValues, mapPropertiesByHash(ctx, dynamicValues))
  )
);
```

Which is closer to:

```js
computed(...dependenKeys, getterFn)
```

Issue #695, opened 2018-11-19
Get current locale value (string or array?)


BHuys:
- [yes] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [yes] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version:  3.5.0
- Ember CLI Version: 3.5.0
- Ember Intl Version: latest
- Browser(s): chromium v68
- Node Version: 8.11.3

I need the current locale at different instances in my application, to get the current locale I use:
`this.get('intl').get('locale')`

This always returns an array with a single string value, is there a way to just return te value instead of an array with a single element? 

My value was set using 
`this.get('intl').setLocale('en')`

How do I get a return of "en" instead of ["en"] ?

I was searching the docs when I noticed this:

>Locale

> Set/get the current locale for your application. The value can either be a String or an Array of Strings.

But couldn't find more info about when which return is given.

Thanks in advance

Driezzz:
Why doesn't `this.get('intl.locale')` just return the locale string? Currently, everytime you want to just get the active locale, you need to write a computed property.

buschtoens:
Because `setLocale` accepts an array of locales in order to support fallbacks. This is the reason, why there is no "one" locale. There is only a primary locale and a number of fallback locales.

You could access `this.get('intl.locale.firstObject')`.

We could also think about adding a `primaryLocale` to the `intl` service for convenience.

Driezzz:
@buschtoens With `primaryLocale `, do you mean the currently Active locale or the primary locale in the locales[]. 

If it's the current Active locale, that would be very helpful!

buschtoens:
@Driezzz With `primaryLocale` I mean the first locale of the currently active locales, i.e. `this.get('intl.locale.firstObject')`.

The `locales` array in `config/ember-intl.js` is really just a set, not an ordered array. There is not primary locale. ☺️

AndreJoaquim:
I know this is already closed but it'd just like to say that it'd be really cool to have the `primaryLocale` alias to `intl.locales.firstObject` in the `intl` service.

As I'm migrating from `ember-i18n` to `ember-intl`, I'll have to replace `i18n.locale` with `intl.locale.firstObject` (which can be confusing for some people as `locale` is singular but represents an array) everywhere. With the `primaryLocale` everything would be better :)

buschtoens:
[...]

PR #726, opened 2018-12-13
Add primaryLocale to intl service

AndreJoaquim:
This PR add an alias to `locale.firstObject` to the `intl` service called `primaryLocale`. This alias improves code readability when accessing the primary locale and eases the transition from `ember-i18n` to `ember-intl`: `i18n.locale` -> `intl.primaryLocale`

Closes https://github.com/ember-intl/ember-intl/issues/695 for good!

I've added documentation regarding this new property and updated the `locale` documentation, as it's always an array.

### DIFF
--- a/docs/ember-service-api.md
+++ b/docs/ember-service-api.md
@@ -14,13 +14,15 @@ Ember.Object.extend({
 });
 ```
 
-Access the service from within the instance via: `this.get('intl')` or `this.intl`
+Access the service from within the instance via: `this.get('intl')` or just `this.intl`, if you have [ES5 getters enabled](https://www.emberjs.com/blog/2018/04/13/ember-3-1-released.html#toc_es5-getters-for-computed-properties-2-of-4).
 
 ## Properties
 
 **locale**
 
-Set/get the current locale for your application.  The value is an Array of Strings.  When providing an array, the `t` helper and `t` method will attempt to try all the locales in order when resolving a translation key.  This is useful if you want to always fallback to another locale when a translation may be missing.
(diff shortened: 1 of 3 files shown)


docs/ember-service-api.md
```diff
@@ -14,13 +14,21 @@ Ember.Object.extend({
...
 
-Access the service from within the instance via: `this.get('intl')`
+Access the service from within the instance via: `this.get('intl')` or `this.intl`
 
 ## Properties
 
 **locale**
 
-Set/get the current locale for your application.  The value can either be a String or an Array of Strings.  When providing an array, the `t` helper and `t` method will attempt to try all the locales in order when resolving a translation key.  This is useful if you want to always fallback to another locale when a translation may be missing.
+Set/get the current locale for your application.  The value is an Array of Strings.  When providing an array, the `t` helper and `t` method will attempt to try all the locales in order when resolving a translation key.  This is useful if you want to always fallback to another locale when a translation may be missing.
```
buschtoens:
```suggestion
Set/get the current locale for your application. The value you set it to can either be a string or an array of strings. When providing an array, the `t` helper and `t` method will attempt to try all the locales in order when resolving a translation key. This is useful if you want to always fallback to another locale when a translation may be missing.

When you get this property, it will always return an array of strings, even if you have set it to be just one single locale. If you are only interested in retrieving the single (or first) locale, use **`primaryLocale`**.
```

docs/ember-service-api.md
```diff
@@ -14,13 +14,21 @@ Ember.Object.extend({
 });
 ```
 
-Access the service from within the instance via: `this.get('intl')`
+Access the service from within the instance via: `this.get('intl')` or `this.intl`
```
buschtoens:
```suggestion
Access the service from within the instance via: `this.get('intl')` or just `this.intl`, if you have [ES5 getters enabled](https://www.emberjs.com/blog/2018/04/13/ember-3-1-released.html#toc_es5-getters-for-computed-properties-2-of-4).
```

Issue #730, opened 2018-12-17
ICU argument missing combined with falsey requiresTranslation case


PoslinskiNet:
Hi there,

I've encountered probably an edge case that could be covered (if I understand it correctly).

Let me explain the case:

1. We have a few languages.
2. Let's assume that our base language is `en-GB`
3. Other languages can have a translation, but if they don't they should use `en-us` as a fallback.
4. Everything works fine, especially when I change `ember-intl.js` a little:

```javascript
requiresTranslation(key, locale) {
  return locale.toLowerCase() === 'en-gb';
}
```

The only problem I have is the mismatch between arguments in the translations (ICU argument missing) - but in fact, I don't care about this argument anymore since I want to use fallback translation. It makes sense if there are some typos in arguments (and I think that was the point of this feature when I saw an connected issue: https://github.com/ember-intl/ember-intl/issues/601), but if there is no translation and it can be skipped (requiresTranslation resolves false for the given key/locale) - it should also skip ICU argument comparison. 

What do you think? Maybe there is some other way to achieve such a behaviour or I can prepare PR with the edge case covered?

Here is the code that is connected with the issue: https://github.com/ember-intl/ember-intl/blob/a47251122ca5d197dca50a430177460ffd2f75ec/lib/broccoli/translation-reducer/index.js#L132

### Environment

- Ember Version: 3.2
- Ember CLI Version: 3.2
- Ember Intl Version: 4.0.0-beta.4
- Browser(s): N/A
- Node Version: N/A

### Steps to Reproduce

```json
// en-gb.json
{
   "test": "variable: {var}",
   "something": "bla bla"
}

// en-us.json
{
   "something": "bla bla 2"
}
```

jasonmit:
@PoslinskiNet I'm able to reproduce this.

> Let's assume that our base language is en-GB

Are you referring to the `fallbackLocale` option?

> I can prepare PR with the edge case covered?

Would very much appreciate this.

PR #731, opened 2018-12-18
Don't report ICU linting errors if not needed

PoslinskiNet:
# Use case
Don't report ICU arguments lint errors if a translation is missing and is not obligatory.

It resolves the following issue: https://github.com/ember-intl/ember-intl/issues/730

### DIFF
--- a/tests-node/unit/broccoli/translation-reducer/index-test.js
+++ b/tests-node/unit/broccoli/translation-reducer/index-test.js
@@ -203,16 +203,12 @@ describe('translation-reducer', function() {
       ]);
     });
 
-    it('lintTranslations returns list of icu argument mismatch without those that are required', function () {
-      let subject = new TranslationReducer(this.input.path(), { fallbackLocale: 'en'});
+    it('lintTranslations returns list of icu argument mismatch without those that are required', function() {
+      let subject = new TranslationReducer(this.input.path(), { fallbackLocale: 'en' });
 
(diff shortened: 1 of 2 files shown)


tests-node/unit/broccoli/translation-reducer/index-test.js
```diff
@@ -203,6 +203,23 @@ describe('translation-reducer', function() {
       ]);
     });
 
+    it('lintTranslations returns list of icu argument mismatch without those that are required', function () {
+      let subject = new TranslationReducer(this.input.path(), { fallbackLocale: 'en'});
+
+      subject.options.requiresTranslation = (_, locale) => locale === 'en';
```
jasonmit: The implementation I was thinking of was avoid reporting on ICU Argument Mismatching if the translation came from the fallback locale.  I don't believe `requiresTranslation` should play a part in this implementation.

Let me know if you have thoughts on this.

PoslinskiNet: I agree with that and it's gonna be even simpler that way. Please take a look at recent changes then 👍

PoslinskiNet:
@jasonmit any thoughts on this after feedback commit 😃?

jasonmit:
@PoslinskiNet looks good, I'll version a patch release today :)

PoslinskiNet:
@jasonmit thanks for merging this one. Looking forward to version release, so we can start using this on the project right after new release is out.

PR #767, opened 2019-01-29
Update pushing example with extra await statement

vasind:
Issue:
On previous example, `promise` object got pushed and it is not working as expected.

Solution:
The `response` has to be resolved and then push it to the `addTranslations` API

### DIFF
--- a/docs/asynchronously-loading-translations.md
+++ b/docs/asynchronously-loading-translations.md
@@ -38,9 +38,8 @@ export default Ember.Route.extend({
   intl: Ember.inject.service(),
   
   async beforeModel() {
-    const response = await fetch('/translations/en-us.json');
-    const translations = await response.json();
-    this.get('intl').addTranslations('en-us', translations);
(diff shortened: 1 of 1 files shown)


docs/asynchronously-loading-translations.md
```diff
@@ -38,9 +38,9 @@ export default Ember.Route.extend({
   intl: Ember.inject.service(),
   
   async beforeModel() {
-    const translations = await fetch('/translations/en-us.json');
-    
-    this.get('intl').addTranslations('en-us', translations.json());
+    const response = await fetch('/translations/en-us.json');
+    const translations = await response.json();
+    this.get('intl').addTranslations('en-us', translations);
```
makepanic: I think you can also change it to

```js
const translations = await fetch('/translations/en-us.json');
    
this.get('intl').addTranslations('en-us', await translations.json());
```

which might be cleaner

vasind: @makepanic Done as per comment.

PR #774, opened 2019-02-02
bug: fixes treeFor hooks from fixing many times per build.

jasonmit:
Before this, all treeFor hooks would fire as many times as it's present in a addon + once for the host app.  This yields better build performance results for those that are in that scenario.


buschtoens:
Relevant RFC: https://github.com/ember-cli/rfcs/blob/master/complete/0090-addon-tree-caching.md

PR #815, opened 2019-03-04
refactor(macros): avoid using `ComputedProperty` directly

### LINKED ISSUE #785: ComputedProperty system will need to be updated for 3.9
Hey there, wanted to open this issue up for visibility! We've merged some major refactors upstream in Ember to set things up for the upcoming change to ship native decorators. The system is becoming _more_ private, but we've made sure that this library and `ember-concurrency` have an upgrade path in the meantime. In the future, I believe a true decorator is pretty much what you'll want anyways 😄 

The gist of the new system is that you actually export a spec compliant decorator, and `Ember.defineProperty` knows how to apply that decorator, you just need to mark it as a decorator manually. We can help out with the upgrade process, let us know if you have any questions! (and hopefully, we'll have a PR up soon)

buschtoens:
This fixes #785 by avoiding direct usage of `ComputedProperty`, which should be regarded as private API.

One drawback is that we now cannot eagerly inject the `intl` service when using the `t` (or `intl`) computed property macro. In hindsight, I think that this was a really bad idea anyway. But historically it made sense, since `t` expected the service to already be injected.

This new computed property macro lazily injects the `intl` service as a non-enumerable, semi-random property, when the CP is accessed for the first time. The injection is non-enumerable & semi-random to prevent userland code from implicitly depending on the injection being made for them.

Any already injected `intl` service will be ignored.

This is a _somewhat_ breaking change for users which relied on this implicit behavior, but since master is `v4` anyway, that's fine, I guess. 👍 

#### Breaking Changes

- Translation macros doe not implicitly inject the `intl` service any more.
  <sup>Well it still does, but it's supposed to be invisible to userland code.</sup>
- If the `intl` service is already injected, it will be ignored by the translation macros and they will use their own invisible service injection instead.

#### Possibly Breaking Changes

- The macro source files have been moved, but if you only imported from `ember-intl` directly, you will not notice.

#### Deprecations

- The export name for the `t` translation macro has changed from `translationMacro` to `t`. The old name is still supported, but will log a deprecation message.
  ```diff
  - import { translationMacro as t, raw } from 'ember-intl';
  + import { t, raw } from 'ember-intl';
  ```


addon/macros/intl.js
```diff
@@ -0,0 +1,24 @@
...
+
+export const __intlInjectionName = `intl-${Date.now().toString(36)}`;
+
+export default function intl(...dependentKeysAndGetterFn) {
+  const getterFn = dependentKeysAndGetterFn.pop();
+  const dependentKeys = dependentKeysAndGetterFn;
+
+  return computed(function(propertyKey) {
+    if (!get(this, __intlInjectionName)) {
+      defineProperty(this, __intlInjectionName, service('intl'));
```
pzuraq: It might be cleaner to do `this.set(__intlInjectionName, getOwner(this).lookup('service:intl'));`, this is a case where using the owner directly would actually probably be cheaper and a bit cleaner.

buschtoens: Oh yes, definitely!

buschtoens:
image


🎉

buschtoens:
@jasonmit Do you agree with the breaking changes and deprecations I've listed out above?

addon/macros/intl.js
```diff
@@ -0,0 +1,22 @@
+import { computed, get, defineProperty } from '@ember/object';
+import { getOwner } from '@ember/application';
+
+export const __intlInjectionName = `intl-${Date.now().toString(36)}`;
```
buschtoens: We have to use a string instead of a symbol here, so that we can add `${__intlInjectionName}.locale` as a dependent key in L21.

jasonmit:
Thanks for taking this up :)  LGTM!

I'll cut a new 4.0.0 beta later on today.

Issue #797, opened 2019-02-20
Defining translations inside engines


buschtoens:
I've noticed that `publicOnly` does not play well with engines. If enabled, the generated translation file is placed in the host app's `translations` dir and not in the engine's `translations` dir.

Disabling `publicOnly` also doesn't work, because the translation is not bundled into `engine.js` or `engine-vendor.js`, but seemingly gets merged into the host app's app tree:

https://github.com/ember-intl/ember-intl/blob/57691dd4af26d9722bd00a2ae01284d303e14b35/index.js#L111-L142

#### Input

`publicOnly: true`

``` bash
client
├── package.json       # name: "client"
├── app
│   └── ...
└── translations
    └── de-de-x-host.yml

client-packages/engines/
└── direct-sales
    ├── package.json   # name: "@clark/engine-direct-sales"
    ├── addon
    │   └── ...
    └── translations
        └── de-de-x-engine.yml
```

#### Expected Output

```
dist
├── asset-manifest.json
├── assets
│   ├── client.js
│   └── vendor.js
├── engines-dist
│   └── @clark
│       └── engine-direct-sales
│           ├── assets
│           │   ├── engine-vendor.js
│           │   └── engine.js
│           ├── config
│           │   └── environment.js
│           └── translations
│               └── de-de-x-engine.json
└── translations
    └── de-de-x-host.json
```

#### Actual Output

```
dist
├── asset-manifest.json
├── assets
│   ├── client.js
│   └── vendor.js
├── engines-dist
│   └── @clark
│       └── engine-direct-sales
│           ├── assets
│           │   ├── engine-vendor.js
│           │   └── engine.js
│           └── config
│               └── environment.js
└── translations
    ├── de-de-x-host.json
    └── de-de-x-engine.json
```

#### Problem

If the engine and host app would share the same locale name, e.g. just `de-de`, one would clobber the other. Moving this to the "root" also makes it harder for an engine to reliably load the translation, _I would assume_.

I think this is just an oversight, since ember-intl has rarely been used with ember-engines.

#### Solution

We special-handle engines and merge into `treeForAddon` instead of `treeForApp`. In general, we might want to allow addons that specify `ember-intl` as a dependency to configure this.

jasonmit:
> Disabling `publicOnly` also doesn't work, because the translation is not bundled into `engine.js` or `engine-vendor.js`, but seemingly gets merged into the host app's app tree

[...]

Issue #728, opened 2018-12-14
Can't find get started guide


xud6:
I can't find any guide / get start guide on how to use this addon.

buschtoens:
I was meaning to set up ember-cli-addon-docs for quite some time and migrate the `docs` directory over.

xud6:
Actually i didn't have much problems with docs directory as long as it's organized.
The biggest problem i have is where to start in the docs folder. It seems all of the docs in the docs directory assumes you know the basics of how the addon work. But i need docs that can help me get started
The ember i18n addon's doc (https://github.com/jamesarosen/ember-i18n/wiki), while not fancy, is way better for users with zero knowledge of the addon.

jasonmit:
@xud6 good feedback, thanks.  Also, please send any PRs my way that you think could improve the docs.

hakilebara:
I would like to work on a ember-cli-addon-docs migration

jasonmit:
Would love that @hakilebara!

hakilebara:
I propose the following structure:

- Getting Started
    - Overview
    - Installation
    - Quickstart guide
- Guide
    - Asynchronous loading translations
    - Compact decimal formatting
    - Ember service API
    - Intl.js polyfill
    - Supported locales
    - Missing translations
    - Testing
- Template Helpers
    - The `t` helper
    - format-date
    - format-number
    - format-message
    - format-relative
    - format-time
- Cookbook
    - log translations not found on Sentry
    - migrate from 2.0 to 3.0
    - migrate from 3.0 to 4.0
    - migrate from ember-i18n
- Advanced
    - addon support
    - ember-cp-validation

jasonmit:
That all seems reasonable, can you add a placeholder under Advanced for Engine support?  We'll need to document the behavior for the currently opened PR.

PR #846, opened 2019-04-11
Initial ember-engines support.

brettburley:
In looking into #797 in more detail, we found the current behavior of ember-intl in engines to be a bit unexpected (see https://github.com/ember-intl/ember-intl/issues/797#issuecomment-479195256 for more details). Looking for feedback on the proposed behavior changes to better support ember-engines.

This change makes a few modifications to the behavior in the build process. With these changes, there are now two sets of behavior, depending on how an engine depends on `ember-intl`: 

* If an ember-engine depends directly on `ember-intl`, it will perform its own `ember-intl` build and produce its own set of translation files. This build will respect the configuration provided in the `config/ember-intl.js` file in the engine. It still uses the treeForApp hook because ember-engines already transforms this output so that they’re namespaced for the engine. Additionally, if a host application of the engine depends itself on `ember-intl`, it will ignore the engine addon and not bundle the engine’s translations in the host app. This provides engine-level translation isolation, and the expectation is that each engine will use its own independent intl service.
* If an ember-engine does not depend directly on `ember-intl`, then it will be treated as a normal addon, and its host application will bundle its translations in its own translation build. This enables folks to use the host application’s intl service and pass it down to the child engine and use the host application’s intl service.

In order to introduce this behavior, we did have to introduce knowledge into the addon during the build process on whether it’s being invoked in an engine or not (since they have slightly different methods of fetching configuration data), and so that we can treat addons as either plain addons, or engines with their own, scoped ember-intl config/build. We also had to fix the `cacheKeyForTree` cache key so that the same assets wouldn’t be re-used for every engine that’s built.

Not entirely sure how best to write tests for this behavior (it looks like the existing build tests are relatively light, e.g. https://github.com/ember-intl/ember-intl/blob/master/tests-node/unit/index-test.js?), any suggestions welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 const buildTree = require('./lib/broccoli/build-translation-tree');
 const TranslationReducer = require('./lib/broccoli/translation-reducer');
 const isValidLocaleFormat = require('./lib/utils/is-valid-locale-format');
+const findEngine = require('./lib/utils/find-engine');
 
 const DEFAULT_CONFIG = {
   locales: null,
@@ -40,20 +41,25 @@ module.exports = {
   isLocalizationFramework: true,
 
   cacheKeyForTree(treeType) {
-    return calculateCacheKeyForTree(treeType, this);
+    return calculateCacheKeyForTree(treeType, this, this.package.root);
   },
 
-  included() {
+  included(parent) {
     this._super.included.apply(this, arguments);
 
-    this.app = this._findHost();
-    this.opts = this.createOptions(this.app.env, this.app.project);
-    this.isSilent = this.app.options.intl && this.app.options.intl.silent;
+    const host = this._findHost();
+    const engine = findEngine(parent);
+
+    this.package = engine || this.project;
+
+    this.appOptions = (engine || host).options || {};
+    this.opts = this.createOptions(host.env);
+    this.isSilent = host.options.intl && host.options.intl.silent;
     this.locales = this.findAvailableLocales();
   },
 
   generateTranslationTree(bundlerOptions) {
-    const translationTree = buildTree(this.project, this.opts.inputPath, this.treeGenerator);
+    const translationTree = buildTree(this.package, this.opts.inputPath, this.treeGenerator);
     const _bundlerOptions = bundlerOptions || {};
     const addon = this;
 
@@ -89,7 +95,7 @@ module.exports = {
 
   contentFor(name, config) {
     if (name === 'head' && !this.opts.disablePolyfill && this.opts.autoPolyfill) {
-      let assetPath = this.findAssetPath(this.app.options);
+      let assetPath = this.findAssetPath(this.appOptions);
       let locales = this.locales;
       let prefix = '';
 
@@ -145,12 +151,10 @@ module.exports = {
     let trees = [];
 
     if (!this.opts.disablePolyfill) {
-      let appOptions = this.app.options || {};
-
       trees.push(
         require('./lib/broccoli/intl-polyfill')({
           locales: this.locales,
-          destDir: (appOptions.app && appOptions.app.intl) || 'assets/intl'
+          destDir: this.findAssetPath(this.appOptions)
         })
       );
     }
@@ -175,11 +179,11 @@ module.exports = {
   },
 
   readConfig(environment, project) {
-    // NOTE: For ember-cli >= 2.6.0-beta.3, project.configPath() returns absolute path
-    // while older ember-cli versions return path relative to project root
-    let configPath = path.dirname(project.configPath());
+    let configPath = project.configPath ? path.dirname(project.configPath()) : path.join(project.root, 'config');
     let config = path.join(configPath, 'ember-intl.js');
 
+    // NOTE: For ember-cli >= 2.6.0-beta.3, project.configPath() returns absolute path
+    // while older ember-cli versions return path relative to project root
     if (!path.isAbsolute(config)) {
       config = path.join(project.root, config);
     }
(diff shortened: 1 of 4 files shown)


jasonmit:
Thanks for taking this up!  I'll need some time to review.  I'm comfortable landing any of these breaking changes in the beta channel - as I never officially supported engines.

The testing story around an engine integration I'm also unclear about - @buschtoens any ideas?

buschtoens:
Thanks so much for taking this on! I'm sorry that I did not get to review this yet, but will try to do so ASAP. I'm super interested in this!

jasonmit:
@brettburley when you get a chance, can you add an MD file to the docs section that details the expected behavior of Engines and perhaps instructions on setting it up.

Thanks!

Related: https://github.com/ember-intl/ember-intl/issues/728#issuecomment-485491605

index.js
```diff
@@ -18,6 +18,7 @@ const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 const buildTree = require('./lib/broccoli/build-translation-tree');
```
jasonmit: Issues I'm seeing right now testing this within my application:
* In my application I'm seeing translations duplicated in the payload - once in app.js and once in vendor.js.  I'd expect engine translations to be in vendor.js only.
* Engine's are unable to implement `treeForTranslations`

brettburley:
@jasonmit does your engine have a `dependencies` (i.e. _not_ `devDependencies`) on `ember-intl`, as well as a `keyword: ['ember-engine']` in the `package.json`? This change is using this to identify if a package is an engine or not, see https://github.com/ember-engines/ember-engines/issues/574 for reference.

jasonmit:
@brettburley it does have `"ember-intl"` under dependencies along with the `"ember-engine"` keyword set.  The strange thing is some of the translations are coming in via the hook and some are not - haven't quite figured out why but verified it by looking at the bundled output.

lib/broccoli/build-translation-tree.js
```diff
@@ -28,8 +29,14 @@ function buildTranslationTree(project, inputPath, treeGenerator) {
...
 }
 
+function isEngineWithIntl(addon) {
+  return isEngine(addon) && addon.pkg.dependencies['ember-intl'];
+}
+
 function processAddons(addons, translationTrees, treeGenerator) {
-  addons.forEach(addon => _processAddon(addon, translationTrees, treeGenerator));
+  addons
+    .filter(addon => !isEngineWithIntl(addon))
```
jasonmit: This line ended up being the reason for the custom treeForTranslations hook not firing.  I'll need to think about how we can preserve that behavior along with your changes.

brettburley: Is the use case for the engine to add additional paths for its own build, or to add translations to the host application (that embeds the engine)?

jasonmit: The hook is used to programmatically provide translations to the engine or app.  In the context of your changes, this would be scoped to the immediate engine and not the host app.

It turned out to be a trivial change to what you've already done: https://github.com/ember-intl/ember-intl/commit/8f30184e2607ce2de92d115603eacefa30edd233

Feel free to cherry-pick into your fork & branch.

jasonmit: There's still the issue where I'm finding the engine's translations being bundled as the host translations.  I'll try and poke around to see if I can figure that one out.

jasonmit:
I was able to isolate the other issue outside of my project.

```
git clone git@github.com:jasonmit/ember-intl-bug.git
cd ember-intl-bug
yarn
yarn start
```

Notice that the translation for `app` is not in the bundle however `engine` is unexpectedly there.
You can verify this by looking at `packages/x-app/dist/x-app.js` and grepping for `x-app/translations/en-us`.

If you remove `x-addon` from `x-engine`'s pkg.json, app translations will bundle correctly.

jasonmit:
@brettburley any ideas on the above?  I haven't had a chance to circle back to it.

scottkidder:
I'm hoping to be able to test this in our app soon. Thanks for the great work @brettburley @jasonmit ! :tada:

brettburley:
Finally took a look into the issue you produced @jasonmit (thanks for the reproduction!). Still not entirely sure on the right solution. It appears that what's happening is that ember-cli is reusing the `x-addon` build it does for the engine for the main app as well:
https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L677

If I redefine `x-addon` to not cache, it fixes the issue:
```
module.exports = {
  name: require('./package').name,
  cacheKeyForTree() {
    return null;
  }
};
```

It seems like either we can force the `ember-intl` addon instance that is on the main app to run (and don't let children packages that include it to run also) -- i.e. only allow 1 `ember-intl` addon instance per app/engine to run, or find some way of making the addon cache be invalidated? Not sure the best path forward or existing solutions for similar issues.

jasonmit:
@brettburley I also had a feeling it was around the cache key logic but never validated it.  That's unfortunate to see this being a deep root upstream bug.

Perhaps we detect if an addon is dependent on ember-intl and is shared across an engine and app.  If we detect it, we log out the known issue and offer the work around you pointed out.

It's the best thing I got right now and will continue to think about it/ask around.

netes:
Hey guys,
Thanks for this! I stumbled upon the problem with `ember-engines` and `ember-intl` recently and this seems to be the solution. Any update on this? :)

jasonmit:
@netes no update yet, blocked on the issue I mentioned above.

abhilashlr:
Great work @brettburley!

This PR seems to work fine as required in #797. However, the advantage of app overriding the keys of engine's or addon's translation seems to be lost. Perhaps the following works:
```
dist
├── asset-manifest.json
├── assets
│   ├── client.js
│   └── vendor.js
├── engines-dist
│       └── engine-a
│           ├── assets
│           │   ├── engine-vendor.js
│           │   └── engine.js
│           ├── config
│           │   └── environment.js
│           └── translations
│               └── de-de.json
└── translations
    └── de-de.json
```
but if de-de.json (of the engine) had few keys that needed to be overridden by de-de.json (from the app) - that does not seem to work.

jasonmit:
It's possible https://github.com/ember-intl/ember-intl/pull/1152 may fix the bug I mentioned above.  I'll need to rebase and verify in the future.

jasonmit:
Closing since this has fallen out of date and appears to be stale.

jasonmit:
I think we're in a better place if someone wants to try and rebase and bring it back to life - the addon build code is now signifcantly less complicated now that the polyfill logic has been removed.

PR #849, opened 2019-04-14
Added a very important bit of config code in the README (at least for ember version…

nucleoid:
… 2.18.2).  Otherwise some views fail to render at all.


jasonmit:
Thanks @nucleoid  :)

PR #867, opened 2019-05-10
fix: Always go through setLocale even with the default locale.

### LINKED ISSUE #866: Always update document language
Can we always update the document title here at L48?

https://github.com/ember-intl/ember-intl/blob/b9ea8fef81ff05458e182e215016f4d3a2054cbd/addon/services/intl.js#L44-L51

I have a use case that we set different font-family depends on `html[lang]` attribute. The problem is, on the first time the page gets loading, `this.updateDocumentLanguage` don't have a chance to run if you set `'en-us'` as the current locale. I believe we should always call `this.updateDocumentLanguage` when we set a new locale, or if we want to avoid to set the same value multiple times, we should give `this._locale` a `null` as the default value, make sense?

jasonmit:
fixes #866

### DIFF
--- a/addon/services/intl.js
+++ b/addon/services/intl.js
@@ -25,7 +25,7 @@ import getDOM from '../utils/get-dom';
 
 export default Service.extend(Evented, {
   /** @private **/
-  _locale: ['en-us'],
+  _locale: null,
 
   /** @private **/
   _adapter: null,
@@ -46,8 +46,6 @@ export default Service.extend(Evented, {
         cancel(this._timer);
         this._timer = next(() => this.trigger('localeChanged'));
         this.updateDocumentLanguage(this._locale);
-
-        return this._locale;
       }
 
       return this._locale;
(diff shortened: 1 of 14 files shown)


addon/services/intl.js
```diff
@@ -93,14 +88,17 @@ export default Service.extend(Evented, {
...
 
     if (typeof Intl === 'undefined') {
       warn(`[ember-intl] Intl API was not found.\nSee: ${links.polyfill}`, false, {
         id: 'ember-intl-undefined-intljs'
       });
     }
 
+    const initialLocale = get(this, 'locale') || ['en-us'];
+
+    this.setLocale(initialLocale);
```
jasonmit: I plan to do more work in the future around this method.

* renaming
* moving locale setter into setLocale (again)
* deprecate & remove locale setter
* consider asserting provided array (for consistency and can make better downstream assumptions)

PR #937, opened 2019-07-29
Remove extra 'documentation' in readme

k-fish:
It looks like there are two documentation sections, but one links to `/docs` which 404's. Unless I'm missing something, I'm guessing this section is no longer required.


jasonmit:
Thank you

Issue #228, opened 2015-09-11
Warn on nested translation collision


jasonmit:
https://github.com/yahoo/ember-intl/pull/227 introduced the ability to organize translations.

Might me worth it to warn on collision where one of the translation files overwrites another.  Also needs clearer documentation, perhaps in a wiki.

/cc @null-null-null

kraltom:
Might be worth it to prefix keys in nested translation files by directory name. E.g. if file 
`translations/some/directory/en-us.yaml` would contain `hello: "hello"` string, the build would output file `en-us.json` with `{ "some.directory.hello": "hello" }` contents.

PR #945, opened 2019-08-12
Add namespaces to nested translations

AriasBros:
> If a file is under "translations/foo/bar/en-us.json" and we have the "baz" key in that file we will need to access to the translation using the "foo.bar.baz" key instead of "baz". This feature needs be activated setting the config "wrapTranslationsWithNamespace" to true.

The issue #228 is related with this pull request.

Hello, in our current project we have several translations organize in subdirectories. For example, we have a `validation` directory with the next translations:

```
validation:
  accepted: The :attribute must be accepted.
  birthday: The :attribute field is an invalid birthday date.
  digits: The :attribute must be :digits digits.
  digits_between: The :attribute must be between :min and :max digits.
  email: The :attribute must be a valid email address.
  integer: The :attribute must be an integer.

  max:
    array: The :attribute may not have more than :max items.
    number: The :attribute may not be greater than :max.
    string: The :attribute may not be greater than :max characters.

  min:
    array: The :attribute must have at least :min items.
    number: The :attribute must be at least :min.
    string: The :attribute must be at least :min characters.

  not_regex: The :attribute format is invalid.
  number: The :attribute must be a number.
  regex: The :attribute format is invalid.
  required: The :attribute field is required.
```

The idea of this pull request is to have an option in the config file (`wrapTranslationsWithNamespace`) to tell to intl that we want add a namespace/prefix to the nested translations. Converting the previous translation file to:

```
accepted: The :attribute must be accepted.
birthday: The :attribute field is an invalid birthday date.
digits: The :attribute must be :digits digits.
digits_between: The :attribute must be between :min and :max digits.
email: The :attribute must be a valid email address.
integer: The :attribute must be an integer.

max:
  array: The :attribute may not have more than :max items.
  number: The :attribute may not be greater than :max.
  string: The :attribute may not be greater than :max characters.

min:
  array: The :attribute must have at least :min items.
  number: The :attribute must be at least :min.
  string: The :attribute must be at least :min characters.

not_regex: The :attribute format is invalid.
number: The :attribute must be a number.
regex: The :attribute format is invalid.
required: The :attribute field is required.
```
And  using the same key to access to the translations: `validation.max.array`, for example.

### DIFF
--- a/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
+++ b/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
@@ -19,29 +19,28 @@ const path = require('path');
  * @private
  */
 function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
-  filepath = path.dirname(filepath);
-  filepath = filepath.replace(inputPath, '');
+  let dirname = path.dirname(filepath).replace(inputPath, '');
 
-  if (filepath) {
+  if (dirname) {
     let prefix = path.sep;
 
     for (let addon of addonNames) {
       addon = path.sep + addon;
 
-      if (filepath.startsWith(addon)) {
+      if (dirname.startsWith(addon)) {
         prefix = addon;
         break;
       }
     }
 
-    filepath = filepath.replace(prefix, '');
-    const filepathParts = filepath
+    dirname = dirname.replace(prefix, '');
+    const dirnameParts = dirname
       .split(path.sep)
       .filter(part => part)
       .reverse();
 
-    if (filepathParts.length > 0) {
-      for (const pathPart of filepathParts) {
+    if (dirnameParts.length > 0) {
+      for (const pathPart of dirnameParts) {
         object = {
           [pathPart]: object
         };
--- a/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
+++ b/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
@@ -42,7 +42,7 @@ function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
     if (dirnameParts.length > 0) {
       for (const pathPart of dirnameParts) {
         object = {
-          [pathPart]: object
+          [pathPart.replace(/ /g, '_')]: object
         };
       }
     }
--- a/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
+++ b/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
@@ -35,7 +35,10 @@ function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
     }
 
     filepath = filepath.replace(prefix, '');
-    const filepathParts = filepath.split(path.sep).filter(part => part).reverse();
+    const filepathParts = filepath
+      .split(path.sep)
+      .filter(part => part)
+      .reverse();
 
     if (filepathParts.length > 0) {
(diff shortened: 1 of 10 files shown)


jasonmit:
I really appreciate you taking this up!  I'll need a few days to review the changes (busy time for me at the moment).

AriasBros:
Sure @jasonmit, no problem. Let me know any change I must do when you can.

lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
```diff
@@ -0,0 +1,55 @@
...
+ * @method wrapWithNamespaceIfNeeded
+ * @param {Object} object
+ * @param {String} filepath
+ * @param {String} inputPath
+ * @param {String[]} addonNames Names of the addons with translations
+ * @return {Object} Returns the input object
+ * @private
+ */
+function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
+  filepath = path.dirname(filepath);
```
jasonmit: nit: can we use another variable instead of reassigning the passed in argument

tests-node/unit/broccoli/translation-reducer/wrap-with-namespace-if-needed.js
```diff
@@ -0,0 +1,78 @@
+/* eslint-env node */
+'use strict';
+
+let expect = require('chai').expect;
+
+let wrapWithNamespaceIfNeeded = require('../../../../lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed');
+
+describe('wrapWithNamespaceIfNeeded', function() {
```
jasonmit: Can we add a test for folders with spaces

lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
AriasBros: Done!
https://github.com/ember-intl/ember-intl/pull/945/commits/04b5dfcd4279acecace102624ade373ea9696bc0

tests-node/unit/broccoli/translation-reducer/wrap-with-namespace-if-needed.js
AriasBros: Good catch. What should be the better approach to this case?

Should we throw an exception when we find a folder with white spaces?
Should we parse "foo bar" to "foo-bar"?
Should we parse "foo bar" to "foo_bar"?
Should we do nothing, left "foo bar" as "foo bar"? (This is the current behavior)

jasonmit: Lets go with underscores `foo_bar`

AriasBros: Done! https://github.com/ember-intl/ember-intl/pull/945/commits/9aa7bc4a8b0e3126f0f1242055eddcd878c7d9d5

jasonmit:
Thanks for being so responsive through all of this @AriasBros.

I'll cut a release soon :)

AriasBros:
> Thanks for being so responsive through all of this @AriasBros.
> 
> I'll cut a release soon :)

Thanks to you @jasonmit for take the time for the review. 
By the way, I can add some mention to this feature in the documentation if you would like.

Maybe here? (it is the only place where nested translations are mentioned) https://github.com/ember-intl/ember-intl/blob/master/tests/dummy/app/pods/docs/guide/translating-text/template.md

jasonmit:
> By the way, I can add some mention to this feature in the documentation if you would like.

Yes please!  Once you do that I'll cut a new minor release.

Issue #934, opened 2019-07-18
Support for compact style in format-number helper


k-dauda:
- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

It would be nice to have support for a [compact](https://ember-intl.github.io/ember-intl/docs/guide/compact-decimal-formatting) style in the [format-number](https://ember-intl.github.io/ember-intl/docs/helpers/format-number) helper, so we can do something like below in templates:

`{{format-number 19634 style="compact"}}` to display 19.6K

jasonmit:
Official support for this API is moving forward as part of the spec:
https://github.com/tc39/proposal-unified-intl-numberformat/blob/436ed1607c4b446124cf18e25f7eae6d14ed2857/README.md#intlnumberformat-unified-api-proposal

This is what stopped me from implementing this as part of the 4.x API - I did not want to conflict with the incoming native API.

jasonmit:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat#Syntax

Added to the public Intl.NumberFormat API under `notation` property.  Currently only works in Chrome and Node 13 but will enable it in my open PR.

PR #959, opened 2019-08-25
Add helper format-number-compact

AriasBros:
#934

- [x] Add new helper format-number-compact.
`{{format-number-compact 19634}}` will display `20K`
- [x] Add `style` option to select the style from formats file.
`{{format-number-compact 19634 style="oneDigit"}}` will display `19.6K`
- [x] Add documentation for the new helper

### DIFF
--- a/addon/helpers/format-number-compact.js
+++ b/addon/helpers/format-number-compact.js
@@ -9,7 +9,7 @@ import { get } from '@ember/object';
 
 export default BaseHelper.extend({
   format(value, options) {
-    const style = get(options, "style");
+    const style = get(options, 'style');
     const opts = { value };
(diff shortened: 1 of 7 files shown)


jasonmit:
The thing holding this up is I'm unsure if `@ember-intl/intl-messageformat` API should be involved here and instead could use the `cldr-compact-number` API directly.

This will couple us to to intl-messageformat further when I'm anticipating migrating towards https://github.com/messageformat/messageformat for a number of reasons.

jasonmit:
Closing in favor of enabling the incoming native API:
https://github.com/ember-intl/ember-intl/pulls?q=is%3Apr+is%3Aclosed

I'll soon be removing the previous number-compact logic before releasing 5.0.

PR #978, opened 2019-09-17
fix addons without a root breaking the build

robclancy:
Not all addons have a root, we have an addon that just does tasks during a build. It has no root in this case from what I can see.

When I look at this addon compared to others it is returning a `function` on the `index.js` vs all other addons returning an object.

eg. addon with this in `index.js` breaks:

```js
module.exports = function (defaults) {
  return {
    name: 'broken',
  };
};
```

eg. addon with this in `index.js` works as expected:

```js
module.exports = {
  name: 'working',
};
```

### DIFF
--- a/lib/broccoli/build-translation-tree.js
+++ b/lib/broccoli/build-translation-tree.js
@@ -38,6 +38,10 @@ function processAddons(addons, addonNames, translationTrees, treeGenerator) {
 }
 
 function _processAddon(addon, addonNames, translationTrees, treeGenerator) {
+  if (! addon.root) {
+    return;
+  }
+  
   const addonTranslationPath = path.join(addon.root, 'translations');
   let addonGeneratedTree;
 


Turbo87:
why do you think that exporting a function from `index.js` should work? https://github.com/ember-cli/ember-cli/blob/master/blueprints/in-repo-addon/files/__root__/__name__/index.js is just exporting a plain object too 🤔

robclancy:
Because it does work? What kind of question is that? Blueprints provide a
common way of doing things not the only way...

robclancy:
In just going to change it I guess. I don't remember why it needed this way
when it was made years ago. You can close this.

Turbo87:
just because it works doesn't mean it's supposed to work. there is a difference between officially supported APIs and things that just happen to work. and please note that this was a genuine question. I don't know if this is actually supposed to work or not because I could not find any documentation on this pattern and only know that the default blueprints do not generate this pattern so it seemed likely that it's not actually supported.

jasonmit:
I don't feel it's worth officially supporting.  If this becomes supported API in ember-cli, I'll happily reopen.

PR #1004, opened 2019-10-04
docs: add return value type to api docs

### LINKED ISSUE #974: intl service API documentation missing
There's a guide section on service API, but it's not explicit.

For example, I wasn't able to find the return value type of `setLocale()` on the `intl` service.

makepanic:
We could also add the return value to the method description. This change uses TS like return value declarations.

I've also updated `addTranslations` to not return anything as [`addTranslation`](https://github.com/ember-intl/ember-intl/blob/master/addon/models/translation.js#L51) seems to return nothing.

fixes #974

### DIFF
--- a/addon-test-support/add-translations.js
+++ b/addon-test-support/add-translations.js
@@ -19,5 +19,5 @@ export default makeIntlHelper((intl, localeName, translations) => {
     localeName = pickLastLocale(get(intl, 'locale'));
   }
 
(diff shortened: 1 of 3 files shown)


jasonmit:
Much appreciated, thank you!

jasonmit:
It will be published in the next release of 4.x

lolmaus:
Thx! 

As documentation still contains missing sections, should this issue stay open?

PR #1048, opened 2019-11-19
feat: Adds `notation` option to Intl.NumberFormat

jasonmit:
`notation` is an experimental API but should be exposed as a possible option for those that want to start exploring it.

API currently works in Chrome and in Node 12 (partial) and 13 (full)

* [x] Add chrome test

Related to #934

### DIFF
--- a/tests/unit/helpers/format-number-test.js
+++ b/tests/unit/helpers/format-number-test.js
@@ -176,4 +176,17 @@ module('format-number', function(hooks) {
 
     assert.equal(this.element.textContent.trim(), '$3.00€8.00¥10.00');
   });
+
+  test('should support notation option', async function(assert) {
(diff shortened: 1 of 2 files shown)


tests/unit/helpers/format-number-test.js
```diff
@@ -176,4 +176,17 @@ module('format-number', function(hooks) {
 
     assert.equal(this.element.textContent.trim(), '$3.00€8.00¥10.00');
   });
+
+  test('should support notation option', async function(assert) {
+    if (!('notation' in new Intl.NumberFormat().resolvedOptions())) {
+      // test running in unsupported browser
```
jasonmit: Right now only supported in Chrome and Node 12 (partial) and 13 (full support)

PR #1056, opened 2019-11-23
Calculating caching key for tree with additionalCacheKeyParts

abhilashlr:
https://github.com/ember-cli/calculate-cache-key-for-tree/blob/master/lib/calculate-cache-key-for-tree.js#L18

Thanks to PR #846 from @brettburley

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ module.exports = {
   },
 
   cacheKeyForTree(treeType) {
-    return calculateCacheKeyForTree(treeType, this);
+    return calculateCacheKeyForTree(treeType, this, this.project.root);
   },
 
   included() {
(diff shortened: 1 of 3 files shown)


abhilashlr:
Might be a lengthy comment on why this is required. Let's assume, we have a host application and an engine-a.

## Scenario 1
host has ember-intl@4.2.0
engine-a has ember-intl@4.2.0
Resultant translations => `dist/translations/(host+engine).json`

## Scenario 2
host has ember-intl@4.2.0
engine-a has ember-intl@forked-hash
Resultant translations => `dist/translations/(host+engine).json` and `engines-dist/engine-a/translations/(host+engine).json`

## Scenario 3
host does not have ember-intl
engine-a has ember-intl@4.2.0
Resultant translations => `engines-dist/translations/(host+engine).json`

Scenario 2 and 3 were the same working principles until ember-intl@3.5 as well. The issue started popping up when `cacheKeyForTree` was introduced. Which is also being discussed as part of #846.

jasonmit:
Any thoughts on a testing story?  Otherwise, I'll likely continue to not advertise that engines is something that is fully supported.  I may need to re-introduce a mono repo so that we can have complex test scenarios beyond just the single dummy app.

jasonmit:
Released under v4.2.1

abhilashlr:
Thanks @jasonmit!

PR #1069, opened 2019-12-05
#1035: fix: Breaking test cases for nested translations

### LINKED ISSUE #1035: Nesting of translations breaks Testing
Since [nesting (similar to this)](https://www.i18next.com/translation-function/nesting) is not supported by ember-intl yet, the alternative is to define a translation json as:
```
{
  "reuseText": "{count, plural, =1 {Name} other {Names}}",
  "secondKey: "{reuseText} has Hello world"
}
```
and in the template we define the usage as `{{t "secondKey" reuseText=(t "reuseText" count=2)}}`. This isn't going to throw an error. Instead, the `reuseText` key is translated to proper text and applied as a param to the `secondKey`.

But the same template when running in its test case throws an error similar to [this](https://github.com/ember-intl/ember-intl/issues/616).

After debugging, the test cases scenario, I figured out that the `stringify` method makes an additional escape character for `double quotes: "`. And the same value is now applied to the first translation as a value to `reuseText`. But since the intl.message.parser cannot translate a key that contains a `\`, the error of #616 comes up - but only in test cases. The reason being, ember-intl's test-support helper overrides the [missing-message translation util](https://github.com/ember-intl/ember-intl/blob/master/addon-test-support/setup-intl.js#L26).

1. One possible solution is to fix this to replace even the escaped double quotes with a simple doubt quotes in [here](https://github.com/ember-intl/ember-intl/blob/master/addon-test-support/-private/serialize-translation.js#L23).

2. Or, if nesting translations is itself a part of the ember-intl addon, the solution gets simpler where we could define the translation something like the following similar to i18next:
```
{
  "reuseText": "{count, plural, =1 {Name} other {Names}}",
  "secondKey: "$t(reuseText) has Hello world"
}
```

3. Or, is it a mistake to do a translation and then apply that string value to another key?

4. The other possible solution is to rewrite all the strings which are similar to `reuseText` as it's value manually in all the places which are again a tedious manual process like the following:
```
{
  "reuseText": "{count, plural, =1 {Name} other {Names}}",
  "secondKey: "{count, plural, =1 {Name} other {Names}} has Hello world"
}
```
5. Or, is there another totally different solution to this?

abhilashlr:
Fixes #1035

### DIFF
--- a/addon-test-support/-private/serialize-translation.js
+++ b/addon-test-support/-private/serialize-translation.js
@@ -23,7 +23,8 @@ const stringifyDeterministically = obj => JSON.stringify(obj, Object.keys(obj).s
 const replaceInterpolators = subject =>
   String(subject)
     .replace(/\{/g, '(')
-    .replace(/\}/g, ')');
+    .replace(/\}/g, ')')
+    .replace(/\\"/g, '"');
 
 /**
  * A list of internal options that should not be serialized.


abhilashlr:
Sure @buschtoens I will get that done. Thanks :)

jasonmit:
ping @buschtoens

abhilashlr:
> ping @buschtoens

I think it's waiting for me to add test cases. I will try to get something around it over the weekend @jasonmit

abhilashlr:
@jasonmit @buschtoens I'm sorry about being delayed in making the requested changes. I've added the test case for the same and updated the doc block with the Github issue URL. Please have an 👀

abhilashlr:
I've addressed the same in this PR: https://github.com/ember-intl/ember-intl/pull/1252.
@jasonmit @buschtoens

PR #1078, opened 2019-12-12
fix(blueprints): Installation of ember-intl in addon use correct app,config paths

GreatWizard:
## Fix

### Installation of ember-intl in addon use correct app,config paths

The aim of this fix is to installed in the correct app and configs directories depending if the ember project is an app or an addon.

Here is an example of running blueprints in this addon:

```sh
~/workspace/ember-intl  fix/blueprints-addon ✔                                                                                                                                                                                                                           0m
▶ ember g ember-intl
installing ember-intl
? Overwrite tests/dummy/app/formats.js? No, skip
? Overwrite tests/dummy/config/ember-intl.js? No, skip
  skip tests/dummy/app/formats.js
  skip tests/dummy/config/ember-intl.js
  create translations/en-us.yaml
[ember-intl] Don't forget to setup your application-wide locale.  Documentation: https://github.com/ember-intl/ember-intl#setting-locale
```


jasonmit:
Much appreciated!

Issue #733, opened 2018-12-20
Convert to TypeScript


buschtoens:
#600 was an attempt to convert ember-intl to native classes, but it ended up getting reverted for reason I do not recall any more.

I've been using [`ember-cli-typescript@2.0.0-rc.1`](https://github.com/typed-ember/ember-cli-typescript/tree/v2.0.0-rc.1) quite successfully with [`@ember-intl/decorators`](https://github.com/ember-intl/decorators) and other projects and suggest we convert ember-intl over to TypeScript as well.

@jasonmit Do you remember, why you needed to revert #600?

jasonmit:
@buschtoens I unfortunately can't recall exactly what it was other than it didn't work on older builds of Ember.

buschtoens:
Perfect! I think all of this got resolved with [`ember-native-class-polyfill`](https://github.com/pzuraq/ember-native-class-polyfill), which supports Ember 3.4 & 3.5. The polyfilled behavior is native in 3.6.

We currently test down to 2.12. We'd need to update our support matrix then. _Maybe_ we can get the polyfill to run for 2.18, which receives security patches until Feb 2019, [according to the LTS schedule](https://www.emberjs.com/builds/), but I am not sure, if that's worth the effort.

(Some version usage statistics would be priceless here. 😍 )

In any case, I think we should get `v4` out of the door first. Maybe we ship some more features in minor releases and then we can do a `v5` TypeScript refactor afterwards.

jasonmit:
> which supports Ember 3.4 & 3.5

I think we would want to support further back than that for now.

gossi:
Maybe ship an `index.d.ts` at first as "lightweight" solution, so people with Typescript don't get those squiggly lines in their IDE and then conversion afterwards?

PS. Maybe ember observer has some usage statistics

jacobq:
Question: are there type definitions for `ember-intl` somewhere? (What do I need to do to add them to my project?) I am seeing warnings from the TypeScript compiler like this:

```
tests/integration/components/my-component-test.ts:7:30 - error TS2307: Cannot find module 'ember-intl/test-support'.

7 import { setupIntl, t } from 'ember-intl/test-support';
```

gossi:
I made some: https://github.com/gossi/spomoda/blob/master/web/types/ember-intl/index.d.ts
Use them locally and extend them with the ones you need for test-support (I haven't covered them yet)

erkie:
[...]

PR #1106, opened 2020-01-16
Typescript

kturney:
This PR converts the js in the `addon` folder to typescript.

Sorry for submitting this PR without discussion first.
It started out as an "I wonder if I can..." after reading https://github.com/ember-intl/ember-intl/issues/733#issuecomment-544306133.
I'm also submitting it in the spirit of https://github.com/typed-ember/ember-cli-typescript/issues/48.

If we like this, I can also set about converting the `addon-test-support` code.


kturney:
It seems tests are failing because travis is using node 8 and a dep wants node 10.
I can update the node version for travis in this PR, if desired.

jasonmit:
> It seems tests are failing because travis is using node 8 and a dep wants node 10.

I'd like to drop node 8 as part of 5.0.0-stable so feel free to bump in the Travis config and make the commit with BREAKING.

As for the rest of the PR, I'll need some time to go over it.  Thanks a ton for taking this on!

addon/-private/formatters/-base.ts
```diff
@@ -1,25 +1,39 @@
...
 const EMPTY_OBJECT = {};
 
-/**
- * @private
- * @hide
- */
-export default class FormatterBase {
-  get options() {
-    return emberArray();
+export interface FormatterOptions {
```
jasonmit: Could be typed per each formatter since they have slightly different shapes/options

addon/-private/formatters/-base.ts
```diff
@@ -1,25 +1,39 @@
...
+  formats: any;
+}
+
+export type FormatResult = string | { toString(): string };
+
+interface NativeFormatter {
+  format(value: any, options?: object): FormatResult;
+}
+
+export default interface Formatter<T> {
```
jasonmit: If you fix the above, this would become `Formatter<T, FormatterOptionsType>`

addon/-private/normalize-locale.ts
```diff
@@ -0,0 +1,15 @@
+import { assert } from '@ember/debug';
+
+function assertIsString(val: any): asserts val is string {
```
jasonmit: `asserts val is string` - Is this a typescript feature?  If so, pretty awesome!

jasonmit:
Took a first pass at it, looks great so far!  Left some comments.

addon/models/translation.ts
```diff
@@ -0,0 +1,65 @@
...
+    // translations doesn't get set in constructor on ember <= 3.4
+    if (!this.translations) {
+      this.translations = new Map();
+    }
+  }
+
+  /**
+   * Adds a translation hash
+   */
+  addTranslations(translations: any) {
```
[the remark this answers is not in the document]
kturney: this translations param is a JSON blob, I believe, but maybe something like the following would suffice?
```
interface Translations {
  [key: string]: Translations | string
}
```

addon/models/translation.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+}
+
+// export as named and default so the default adapter can augment the class types because default exports cannot be augmented
+// see http://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
+export class TranslationModel extends EmberObject {
+  translations: Map<string, string> = new Map();
+
+  init() {
```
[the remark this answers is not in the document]
kturney: It's my understanding that the declaration (`translations: Map<string, string> = new Map();`) should have initialized translations in the constructor. For some reason in older ember versions that was not the case, so I initialize in `init` if it's missing. I can investigate wether manually initializing in the constructor instead of `init` works in the older versions.

addon/-private/formatters/-base.ts
kturney: I like it. Will do.

addon/-private/normalize-locale.ts
kturney: a brand new feature! https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

addon/models/translation.ts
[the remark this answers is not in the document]
kturney: The `ember-lts-2.16` and `ember-lts-2.18` scenarios fail when initializing in the constructor instead of `init`.

addon/models/translation.ts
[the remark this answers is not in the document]
jasonmit: That works!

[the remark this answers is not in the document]
chriskrycho: If it's a JSON blob, you probably want it to be `[key: string]: Translations | string | undefined`.

Given how common this particular kind of thing is, you *may* want to define a `Dict` type like this:

```ts
interface Dict<T> {
  [key: string]: T | undefined;
}
```

Then in scenarios like this you'd type it as:

```ts
export class TranslationModel extends EmberObject {
  // ...
  addTranslations(translations: Dict<Translations | string>) {
```

kturney:
I rebased to get travis running again.
https://travis-ci.community/t/travis-ci-stopped-tracking-building-a-pr/4159/3

jasonmit:
CI should now be fixed if you rebase

fabmiz:
@buschtoens are there reasons this is not getting merged in?

buschtoens:
Time only, I guess. 😅

I've moved this to my weekend to-do list now. I hope I get some time for it. Thank you for pinging!

.eslintrc.js
```diff
@@ -57,6 +64,16 @@ module.exports = {
...
+    },
+
+    // all TypeScript files
+    {
+      files: ['**/*.ts'],
+      rules: {
+        // These are covered by tsc
+        'no-undef': 'off',
+        'no-unused-vars': 'off'
+      }
```
buschtoens: This is no critique against this PR in any way and more of a note to myself and other maintainers: I personally have a great distaste for these `overrides`-based eslint configs.

At our company we use multiple `root` eslint configs per package with great success. This works especially well, when you have browser TypeScript and Node JavaScript files in the same pacakge.

We might want to use a setup like the one suggested in [`@clark/eslint-config-ember`](https://github.com/ClarkSource/eslint-config/tree/master/packages/eslint-config-ember) in the future.

addon-test-support/-private/make-intl-helper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @return {function}
+ * @hide
+ */
+// export default function makeIntlHelper<T, A1>(fn: (intl: IntlService, arg1: A1) => T): (arg1: A1) => T;
+export default function makeIntlHelper<T extends keyof IntlFunctionProps>(
+  fn: T
+): (...args: Parameters<IntlFunctionProps[T]>) => ReturnType<IntlFunctionProps[T]>;
+export default function makeIntlHelper<T, A1, A2>(
+  fn: (intl: IntlService, arg1: A1, arg2: A2) => T
+): (arg1: A1, arg2: A2) => T;
```
buschtoens: Does this work?

```ts
export default function makeIntlHelper<T, A extends unknown[]>(
  fn: (intl: IntlService, ...args: A) => T
): (...args: A) => T;
```

addon-test-support/-private/pick-last-locale.ts
```diff
@@ -11,4 +8,10 @@ import last from 'lodash.last';
...
  * @hide
  */
-export default locale => last(castArray(locale));
+export default function pickLastLocale(locale: string | string[]) {
+  if (Array.isArray(locale)) {
+    return locale[locale.length - 1];
+  }
+
+  return locale;
+}
```
buschtoens: Some might say an arrow function would be easier to read:

```ts
export default (locale: string | string[]) =>
  Array.isArray(locale) ? locale[locale.length - 1] : locale;
```

But fine with me. 👍

addon-test-support/-private/serialize-translation.ts
```diff
@@ -9,7 +9,7 @@ import omit from 'lodash.omit';
  * @param {object} obj
  * @return {string}
  */
-const stringifyDeterministically = obj => JSON.stringify(obj, Object.keys(obj).sort());
+const stringifyDeterministically = (obj: Record<string, any>) => JSON.stringify(obj, Object.keys(obj).sort());
```
buschtoens:
```suggestion
const stringifyDeterministically = (obj: Record<string, unknown>) => JSON.stringify(obj, Object.keys(obj).sort());
```

addon-test-support/-private/serialize-translation.ts
```diff
@@ -20,7 +20,7 @@ const stringifyDeterministically = obj => JSON.stringify(obj, Object.keys(obj).s
  * @param {string} subject
  * @return {string}
  */
-const replaceInterpolators = subject =>
+const replaceInterpolators = (subject: any) =>
```
buschtoens:
```suggestion
const replaceInterpolators = (subject: unknown) =>
```

addon-test-support/-private/serialize-translation.ts
```diff
@@ -40,7 +40,7 @@ const INTERNAL_OPTIONS = 'resilient default htmlSafe'.split(' ');
  * @param {object} options
  * @return {string}
  */
-const stringifyOptions = (options = {}) =>
+const stringifyOptions = (options: Record<string, any> = {}) =>
```
buschtoens:
```suggestion
const stringifyOptions = (options: Record<string, unknown> = {}) =>
```

addon-test-support/-private/serialize-translation.ts
```diff
@@ -53,7 +53,8 @@ const stringifyOptions = (options = {}) =>
  * @return {string}
  * @hide
  */
-export const serializeTranslation = (key, options) => `t:${key}:${stringifyOptions(options)}`;
+export const serializeTranslation = (key: string, options?: Record<string, any>) =>
```
buschtoens:
```suggestion
export const serializeTranslation = (key: string, options?: Record<string, unknown>) =>
```

addon-test-support/-private/serialize-translation.ts
```diff
@@ -68,4 +69,5 @@ export const serializeTranslation = (key, options) => `t:${key}:${stringifyOptio
  * @return {string}
  * @hide
  */
-export const missingMessage = (key, locales, options) => serializeTranslation(key, options);
+export const missingMessage = (key: string, _: string[], options?: Record<string, any>) =>
```
buschtoens:
```suggestion
export const missingMessage = (key: string, _locales: string[], options?: Record<string, unknown>) =>
```

addon-test-support/setup-intl.ts
```diff
@@ -1,5 +1,16 @@
 import addTranslations from './add-translations';
 import { missingMessage } from './-private/serialize-translation';
+import { Translations } from '../addon/models/translation';
+import IntlService from '../app/services/intl';
+
+interface Hooks {
+  beforeEach: (fn: Function) => any;
+}
```
buschtoens: I think `ember-qunit` already exposes `NestedHooks`.

addon-test-support/setup-intl.ts
```diff
@@ -1,5 +1,16 @@
...
+
+interface ApplicationContext {
+  owner: any;
+  intl?: IntlService;
+}
```
buschtoens:
```ts
import { TestContext } from "ember-test-helpers";

export interface IntlTestContext extends TestContext {
  intl: IntlService;
}
```

This would allow consumers to do this in their tests:

```ts
import { setupIntl, IntlTestContext } from 'ember-intl/test-support';

module('foo', function() {
  setupIntl(this);

  test('foo', (this: IntlTestContext, assert) {
    this.intl.setLocale('en-US');
    // ...
  })
});
```

addon-test-support/setup-intl.ts
```diff
@@ -16,24 +27,25 @@ import { missingMessage } from './-private/serialize-translation';
...
+    hooks.beforeEach(function(this: ApplicationContext) {
+      this.intl!.setLocale(locale as string | string[]);
     });
+  } else if (typeof locale === 'object') {
+    hooks.beforeEach(() => addTranslations(locale, undefined));
   }
 
   if (translations) {
-    hooks.beforeEach(() => addTranslations(translations));
+    hooks.beforeEach(() => addTranslations(translations, undefined));
```
buschtoens: Are the explicit `undefined` arguments necessary?

I would expect the `translations?` to work correctly. Maybe it works with the suggested `...args` change to `makeIntlHelper`? 🤔 

```ts
export default makeIntlHelper((intl: IntlService, localeName: string | Translations, translations?: Translations) => { ... });
```

types/global.d.ts
```diff
@@ -0,0 +1,6 @@
+// Types for compiled templates
+declare module 'ember-intl/templates/*' {
+  import { TemplateFactory } from 'htmlbars-inline-precompile';
+  const tmpl: TemplateFactory;
+  export default tmpl;
+}
```
buschtoens: I think we don't need these types, as `ember-intl` has no templates.

addon-test-support/setup-intl.ts
kturney: `ember-qunit` defines `NestedHooks`, but from what I could tell it isn't exported

kturney:
Thanks for the feedback! I've
- rebased on master
- replaced most uses of `any` with `unknown`
- bumped typescripty deps

jasonmit:
I'm still interesting in landing TS support but this has fallen a bit behind.  Let me know if you want to re-attempt this and I'll try and hold off on any large changes.

PR #1126, opened 2020-01-30
chore(eslint): consolidate config within the root config

kturney:


kturney:
That must be one of the most well put together and well layered eslint config setups I've seen!
I too didn't really care for the config consolidation when it was added to the blueprints, but I generally try to stick with the default setup for ember projects.

buschtoens:
Thanks for the kind words and thanks for the PR! 🤗

PR #1152, opened 2020-02-24
Change tree cache to support engines tested with dummy apps

### LINKED ISSUE #1151: Incorrect behaviour on engine translations in dummy app
When adding `ember-intl` to a standalone Engine (engine within its own repository), translation engine files are no generated. Adding this Engine to a real App generates translation engine files correctly.

### Environment

- Ember Version: 3.1.6.0
- Ember CLI Version: 3.16.0
- Ember Intl Version: 4.2.3
- Browser(s): Chrome 79
- Node Version: 10.15.0

### Standalone engine

Clone and build the following repository:

```
$ git clone https://github.com/BBVAEngineering/intl-engine && cd intl-engine
$ npm install
$ ember build
$ ls dist/engines-dist/intl-engine/translations
```

**Expected**: Engine translation files are generated.

**Reality**: Engine translation files are **not** generated.

### App with engine

Clone and build the following repository:

```
$ git clone https://github.com/BBVAEngineering/intl-app && cd intl-app
$ npm install
$ ember build
$ ls dist/engines-dist/intl-engine/translations
```

Engine translation files are generated correctly.

adrigzr:
This fixes #1151 .

### How to test

Clone and build the following repository:

```
$ git clone --branch bugfix-engine-dummy-app https://github.com/BBVAEngineering/intl-engine && cd intl-engine
$ npm install
$ ember build
$ ls dist/engines-dist/intl-engine/translations
```

Engine translation files are generated correctly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,9 @@ module.exports = {
   },
 
   cacheKeyForTree(treeType) {
-    return calculateCacheKeyForTree(treeType, this, this.package.root);
+    const paths = this.package.paths || this.package.treePaths;
+
+    return calculateCacheKeyForTree(treeType, this, paths ? paths[treeType] : this.package.root);
   },
 
   included(parent) {


index.js
```diff
@@ -50,7 +50,9 @@ module.exports = {
   },
 
   cacheKeyForTree(treeType) {
-    return calculateCacheKeyForTree(treeType, this, this.package.root);
+    const paths = this.package.paths || this.package.treePaths;
```
jasonmit: Can you explain these changes a bit so I understand what's going on.

adrigzr: Sure. 

When you boot the dummy app, the `package.root` for the engine instance is the same as the dummy app instance: the engine `root` folder.

With this change, the tree cache key changes by the tree folder for the instance. So, for example, for the public tree, the engine cache key will be `root/public` and the dummy app cache key will be `root/tests/dummy/public`.

When used inside a real app, in the same example as before, the engine cache key for public tree will be `root/node_modules/engine/public` and for the app `root/public`.

jasonmit: Thanks!  Curious, is `paths` ever falsey?

adrigzr: Yep, When instance is an engine there is a `paths` property and when it is the app it has a `treePaths` property 🤷‍♂ 

Also, some addons has no property at all, so I use the same `root` property.

jasonmit:
I'll cherry pick to 4.x-stable and release a patch tomorrow, thanks!

PR #1180, opened 2020-03-24
Skip ICU arguments warning when requireTranslation returns false

sandstrom:
This will avoid warnings such as 

> [ember-intl] "my.translation.key" ICU argument missing: "de": "color"

When setting `config/ember-intl.js` to something like this:

```js
requiresTranslation(_key, locale) {
      return locale !== 'de'; // german translations aren't required
    },
```

Let me know what you think! @jasonmit

### DIFF
--- a/lib/broccoli/translation-reducer/linter/find-missing-icu-arguments.js
+++ b/lib/broccoli/translation-reducer/linter/find-missing-icu-arguments.js
@@ -1,8 +1,8 @@
 function findMissingIcuArguments(key, allIcuArguments, locales, icuArguments, requiresTranslation) {
   const keyIcuArguments = Array.from(allIcuArguments[key]);
   return locales
-    .map(locale => {
-      let filteredArgs = keyIcuArguments.filter(arg => {
+    .map((locale) => {
+      let filteredArgs = keyIcuArguments.filter((arg) => {
         let missingProp = !Object.prototype.hasOwnProperty.call(icuArguments[locale], key);
         return missingProp || !icuArguments[locale][key].includes(arg);
       });
--- a/lib/broccoli/translation-reducer/linter/find-missing-icu-arguments.js
+++ b/lib/broccoli/translation-reducer/linter/find-missing-icu-arguments.js
@@ -1,4 +1,4 @@
-function findMissingIcuArguments(key, allIcuArguments, locales, icuArguments) {
+function findMissingIcuArguments(key, allIcuArguments, locales, icuArguments, requiresTranslation) {
   const keyIcuArguments = Array.from(allIcuArguments[key]);
   return locales
     .map(locale => {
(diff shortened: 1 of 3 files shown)


jasonmit:
I feel this conflates the meaning  of required translation a bit.  I don't have a great solution yet but will continue to think about the problem.

sandstrom:
My reasoning is that if you don't require a particular translation, it would make sense to squelch the warning for the mismatch in ICU arguments, where the mismatch is between {no arguments since string isn't even present} and {string is present and has same arguments in all present locales}. In other words, the warning doesn't really do what you'd expect it to (in this case).

The way I see it there are a few different options:

- What I proposed in this PR, ignore the message if we don't require this particular locale/key. I.e. it doesn't make sense to warn for ICU arguments mismatch if the key isn't even present.
- Add a new hook, something like `requiresIdenticalIcuArguments`
- Always ignore/skip the ICU mismatch warning if the key isn't present in a locale, in other words when doing the ICU mismatch compare, consider only keys that are non-null, i.e. actual strings.

Perhaps the last one is even better than what I've proposed in this PR.

I'm happy with different solutions too though.

@jasonmit Let me know what you think!

jasonmit:
Reopening as I continue to think about this.

sandstrom:
@jasonmit Let me know if you want me to work on this (either with the current direction, or some new direction that would help squelch these warning messages)

sandstrom:
@jasonmit If a user as squelched "missing translation" warnings via `requiresTranslation` for a particular key + locale, does it make sense to still warn about a mismatch between the arguments in the locale where the key exist, and the non-existing arguments for the locale where the key doesn't exist?

Because if it doesn't, I think what I've proposed here should work pretty much without any changes.

Let me know if I should elaborate or explain more!

sandstrom:
@wopian does the solution proposed here sound good to you?

wopian:
If the solution is only squelching ICU warnings for keys not present in the translated locale then yes.

| Warning | Translation Required | Translation Not Required
| ------- | -------------------- | ------------------------
| Missing key | ✔️ | ❌ 
| Mismatched ICU arguments on keys present | ✔️ | ✔️
| Mismatched ICU arguments on keys not present | ❌\* | ❌

\* I'd argue warning the user of mismatched ICU arguments on keys not present are already covered by the missing key warning.

jasonmit:
@sandstrom can you rebase

sandstrom:
@jasonmit Fixed!

> If the solution is only squelching ICU warnings for keys not present in the translated locale then yes.

@wopian Yes, that's what this PR would introduce.

sandstrom:
@wopian @jasonmit I've added a second PR, with another way of solving this. Perhaps you like that approach better?

https://github.com/ember-intl/ember-intl/pull/1224

jasonmit:
Closed in favor of https://github.com/ember-intl/ember-intl/pull/1224

Thanks!

Issue #1185, opened 2020-03-25
Update formatjs libs


jasonmit:
* [x] migrate from `intl-relativetimeformat` to `@formatjs/intl-relativetimeformat`
Note: There are many breaking changes in this upgrade but necessary to align with the native API.
Breaking changes: API is more or less entirely different now.  Different options and behavior.  No longer supports passing in Date() instances.  Drops use of `now` and `interval`
* [x] update to latest parser and compiler
Breaking changes: HTML tags need to be wrapped in single quotes
* [x] update documentation
Breaking: `Intl.PluralRules` is now natively used so will need to figure out polyfill or documentation story
* [x] fix translation linter to work with new messageformat ast

PR #1186, opened 2020-03-25
feat(BREAKING): upgrades parser, compiler, migrates to native intl.relativetime

jasonmit:
* [x] make translation linter to work with latest messageformat ast

Related to #1185


tests/dummy/snippets/format-relative.hbs
```diff
@@ -1,3 +1,3 @@
 {{format-relative yesterday}}
 {{format-relative instant}}
-{{format-relative now interval=1000}}
+{{format-relative now}}
```
jasonmit: Needs unit

PR #1190, opened 2020-03-25
feat(BREAKING): removes config.locales option.

jasonmit:
This option is obsolete now that ember-intl no longer handles polyfilling.

The reason we removed auto polyfilling is because there are many strategies and they often vary between projects.

If you're interested in understanding the runtime requirements, read:
https://github.com/ember-intl/ember-intl/blob/master/tests/dummy/app/pods/docs/getting-started/runtime-requirements/template.md


sandstrom:
@jasonmit Awesome! 🎉

PR #1225, opened 2020-04-08
Make this addon work with CSP

sandstrom:
See https://github.com/ef4/ember-auto-import/issues/275 for more details

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,7 @@ module.exports = {
 
   options: {
     autoImport: {
+      forbidEval: true,
       webpack: {
         node: {
           global: true,


jasonmit:
I believe that webpack config is global in it's all merged together - so this setting can creep into other addon/app builds.  I'm going to wait for @ef4 to respond before merging.

Which libs bundled via webpack are creating the issue?

ef4:
Yes, this setting doesn't belong in addons.

Apps are allowed to control it directly themselves, and this works even if the app doesn't directly depend on ember-auto-import. Apps can say:

```js
let app = new EmberApp(defaults, {
  autoImport: {
    forbidEval: true
  } 
});
```

And that will control ember-auto-import within all their addons.

sandstrom:
@ef4 Isn't it a problem that if an addon we are using switch to ember-auto-import it will break our app?

I've opened an issue in your repo as well, let's continue the discussion there:
https://github.com/ef4/ember-auto-import/issues/275

jasonmit:
@sandstrom I would be fine with a build time warning if we can some how detect if CSP is enabled and !forbidEval.

PR #1230, opened 2020-04-10
Move webpack configuration into dummy application

bobisjan:
This configuration is affecting our application in FastBoot with other dependencies, which require Node.js environment without mocking like this.

Based on https://github.com/ember-intl/ember-intl/commit/f0419f29e51cbbb0337203de9cd09dba3998b87e I'm not sure if it was affecting end-users? The issue I have found is with importing `testdouble` in tests, hence moving into dummy application configuration. I've tested removing this configuration in our application with `ember-intl#4.2.3` and it was working well both in FastBoot and browser.

@jasonmit, can you confirm, that the original issue was with `testdouble` only?

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -10,6 +10,14 @@ module.exports = function (defaults) {
     options.vendorFiles['jquery.js'] = null;
   }
 
+  options.autoImport = {
+    webpack: {
+      node: {
+        global: true,
+      },
+    },
+  };
+
   let app = new EmberAddon(defaults, options);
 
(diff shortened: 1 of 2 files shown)


jasonmit:
@bobisjan yes I believe it was just testdouble.  I'll backport to 4.x branch and cut a patch for you later today.

bobisjan:
@jasonmit thanks 😍, happy to update to latest 4.x release.

jasonmit:
@bobisjan no need - just released 4.3.1 with your changes.  Thanks again!

PR #1324, opened 2020-05-22
Reduce dependabot noise

sandstrom:
Running it too often creates noise in the commit log. How about running it less often?

image


jasonmit:
I like the change, thank you!

Issue #1061, opened 2019-11-27
(testing) Convert to monorepo for many "dummy" test apps


jasonmit:
I need to bring back the monorepo strategy mainly to more easily support testing multiple dummy app strategies.

* Future support for engine tests
* The addon story
  * Translation collisions between the host app
  * Multiple addons
  * `wrapTranslationsWithNamespace` behavior

houfeng0923:
i found that wrapTranslationsWithNamespace does not work  in v5.1.1

jasonmit:
@houfeng0923 I haven't verified this but the dummy app already has `wrapTranslationsWithNamespace` enabled so should be catching any bugs in a smoke test.  I'll circle back but feel free to look into it if it's urgent.

buschtoens:
We can then also pull out the `ember-cli-addon-docs` into a standalone app. IMO using the test dummy app for docs is an anti-pattern.

PR #1339, opened 2020-05-29
upgrade broccoli-funnel

houfeng0923:
using `wrapTranslationsWithNamespace`,  I have a problem in https://github.com/ember-intl/ember-intl/issues/1061#issuecomment-635263823

after check, i found the reason is the version of broccoli-funnel .
the `relativePath` bug now is published in 3.0.3 fix version ( https://github.com/broccolijs/broccoli-funnel/pull/123 )

so , we can upgrade it now .

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1126,7 +1126,7 @@
   dependencies:
     "@formatjs/intl-utils" "^3.0.1"
 
-"@formatjs/intl-utils@^3.0.0", "@formatjs/intl-utils@^3.0.1":
+"@formatjs/intl-utils@^3.0.1":
   version "3.0.1"
   resolved "https://registry.yarnpkg.com/@formatjs/intl-utils/-/intl-utils-3.0.1.tgz#b9a7f79033b54bff1d03e7f05ba0e0627ba49451"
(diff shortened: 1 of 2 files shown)


houfeng0923:
o, what ?

```
[ember-intl] "subdirectory.smoke.subdirectory" was not found in "es-es", "fr-fr"
cleaning up
cleaning up...
Build Error (broccoli-persistent-filter:Babel > [Babel: @ember-data/store]) in -private/system/core-store.ts
/home/travis/build/ember-intl/ember-intl/-private/system/core-store.ts: Property typeName of TSTypeReference expected node to be of a type ["TSEntityName"] but instead got "MemberExpression"
```

yarn.lock
```diff
@@ -3324,6 +3324,22 @@ broccoli-funnel@^3.0.0, broccoli-funnel@^3.0.2:
     path-posix "^1.0.0"
     walk-sync "^2.0.2"
 
+broccoli-funnel@^3.0.3:
+  version "3.0.3"
+  resolved "https://registry.npm.taobao.org/broccoli-funnel/download/broccoli-funnel-3.0.3.tgz#26fd42632471f67a91f4770d1987118087219937"
```
jasonmit, line 3329: This resolve path should be pointing at yarnpkg.com

houfeng0923: ok, i will update it . i use other mirror in global .

Issue #1071, opened 2019-12-06
intl.locale property should be auto-tracked


sly7-7:
- [ x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [ x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.14
- Ember CLI Version: 3.14
- Ember Intl Version: 4.2.2
- Browser(s): Chrome
- Node Version:

### Steps to Reproduce

Basically I try to remove the observers from my app, and in a helper, I have something like

```
import { observes } from '@ember-decorators/object';
import { inject as service } from '@ember/service';
import Helper from '@ember/component/helper';

export default class FormatListHelper extends Helper {
  @service intl;

  @observes('intl.locale')
  localeChanged() {
    this.recompute();
  }

  compute([elements]) {
    if (elements.length === 1) {
      return elements[0];
    }
    return `${elements.slice(0, -1).join(', ')} ${this.intl.t('and')} ${elements.slice(-1)}`;
  }
}
```

after a discussion with @rwjblue I removed the observer, and if I understood well, just rewriting as this should `just work`:

```
import { inject as service } from "@ember/service";
import Helper from "@ember/component/helper";

export default class FormatListHelper extends Helper {
  @service intl;

  compute([elements]) {
    if (elements.length === 1) {
      return elements[0];
    }

    return `${elements.slice(0, -1).join(", ")} ${this.intl.t(
      "and"
    )} ${elements.slice(-1)}`;
  }
}
```

But for now, I have to access the `intl.locale` property in order to make it work:

```
import { inject as service } from "@ember/service";
import Helper from "@ember/component/helper";

export default class FormatListHelper extends Helper {
  @service intl;

  compute([elements]) {
    if (elements.length === 1) {
      return elements[0];
    }

    this.intl.locale; // used to force auto-tracking of `ember-intl`'s locale computed

    return `${elements.slice(0, -1).join(", ")} ${this.intl.t(
      "and"
    )} ${elements.slice(-1)}`;
  }
}
```

sly7-7:
also, cc/ @marcoow, just for information

rwjblue:
[...]

PR #1340, opened 2020-05-29
Ensure the current locale entangles correctly with autotracking

richard-viney:
Addresses #1071

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "ember-cli-inject-live-reload": "^2.0.2",
     "ember-cli-qunit": "^4.4.0",
     "ember-cli-testdouble-qunit": "^2.1.1",
+    "ember-compatibility-helpers": "^1.2.0",
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.1",
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,9 @@ jobs:
       env: EMBER_TRY_SCENARIO=ember-lts-2.16
     - env: EMBER_TRY_SCENARIO=ember-lts-2.18
     - env: EMBER_TRY_SCENARIO=ember-lts-3.4
+    - env: EMBER_TRY_SCENARIO=ember-lts-3.8
+    - env: EMBER_TRY_SCENARIO=ember-lts-3.12
+    - env: EMBER_TRY_SCENARIO=ember-lts-3.16
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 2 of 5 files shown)


jasonmit:
Strange CI failure, doesn't look related to your changes.  Will kick off a rebuild and will investigate if it continues.  Thanks for taking this up, looks great!

richard-viney:
This may be related to the ember-cli-babel 7.20.1 release today which has caused issues for people: https://github.com/babel/ember-cli-babel/pull/338.

They both involve `MemberExpression`, which does seem a little bit suspicious.

richard-viney:
Pretty certain this CI failure is nothing to do with this PR, it's caused by a regression in `ember-cli-babel@3.20.1` that triggers a build failure inside `ember-data`, possibly to do with its use of `type = import('...')...` syntax. As above, see emberjs/data#7201.

At this stage I don't have a solution other than avoiding `ember-cli-babel@7.20.1`, waiting for a fix, or dropping the `ember-data` dependency from `ember-intl` (but it's wanted by the addon docs so isn't really an option).

richard-viney:
Note that the test for this change is expected to fail when run with Ember < 3.13, but I'm not sure how best to handle this yet.

Also, it would be good to update the list of LTS versions the tests are run against to include the recent LTS versions: Ember 3.8, 3.12, and 3.16.

fivetanley:
@richard-viney the fix is out! bumping to ember-cli-babel 7.20.2 should fix it (or making sure that the babel-plugin-ember-modules-api-polyfill gets updated in the lockfile). more info:  https://github.com/babel/ember-cli-babel/pull/338#issuecomment-635982708

richard-viney:
The `ember-cli-babel` issue has been fixed :+1:, and. I've made the following updates to this branch:

* Test on recent LTS versions: 3.8, 3.12, 3.16.
* Only run the new autotracking test on Ember 3.13 and above.

I saw a separate failure on CI here that I'll attempt to replicate locally if it happens again.

jasonmit:
Ya more CI issues that are most definitely upstream, will continue to wait while they work themselves out.

package.json
```diff
@@ -77,6 +77,7 @@
     "ember-cli-inject-live-reload": "^2.0.2",
     "ember-cli-qunit": "^4.4.0",
     "ember-cli-testdouble-qunit": "^2.1.1",
+    "ember-compatibility-helpers": "^1.2.0",
```
jasonmit, line 80: Is this needed?  I don't see a change to the lockfile

richard-viney: `ember-compatibility-helpers` is already in the lockfile because a bunch of other packages already depend on it. I think this is why `yarn.lock` has no diff on this branch.

We're now importing from it directly for the new test, hence why it's been added to `package.json`.

jasonmit:
Thanks for taking this up, I'll version shortly

jasonmit:
Released as part of 5.2.0, thanks again!

Issue #1337, opened 2020-05-26
can glob configurable?


houfeng0923:
hi, sometime for our project, we just need one lang for build .
so , can we build one of the lang by glob filtering ?

https://github.com/ember-intl/ember-intl/blob/86fb010016f36d3b20283a125998f46f77377b94/lib/broccoli/build-translation-tree.js#L16

jasonmit:
@houfeng0923 what file format were you interested in adding?

houfeng0923:
i have  `translations/en.json` and `translations/ja.json` in app.
and i only need ja or en for one building .

jasonmit:
So you're building an app bundle per locale?  Seems like a reasonable thing we can support.

houfeng0923:
ye,
I've thought about this approach, and it would be better if we didn't adjust the directory structure.
maybe, add an optional filter parameter ?
if it is possible , i can create a fork to try .

jasonmit:
Sounds good, looking forward to the change!

houfeng0923:
hi, @jasonmit
i pull a MR,
use `includeLocales` or `excludeLocales` , it can filter to build necessary resources in inputPath and other translation addons.
and i try to update docs , pls give me your advice for it .

PR #1341, opened 2020-05-29
feat: filter locales options

houfeng0923:
hi, @jasonmit , i see the `locales` option is OBSOLETE_OPTIONS in v5 .
at first, i think we can keep using it just for this filter feature,
util i see loading translations asynchronously which used old `locales` .

so, i add two optional options do this: `includeLocales` `excludeLocales`.

related issue: https://github.com/ember-intl/ember-intl/issues/1337

### DIFF
--- a/tests-node/unit/broccoli/translation-reducer/index-test.js
+++ b/tests-node/unit/broccoli/translation-reducer/index-test.js
@@ -135,6 +135,103 @@ describe('translation-reducer', function () {
     });
   });
 
+  describe('includeLocales', function () {
+    it('should only include specified locales', function () {
+      const subject = new TranslationReducer(this.input.path(), {
+        includeLocales: ['en-us'],
+        errorOnMissingTranslations: true,
+      });
+
+      this.input.write({
+        'de-de.json': `{ "greet": "hallo" }`,
+        'en-us.json': `{ "greet": "hello" }`,
+      });
+
+      return build(createBuilder(subject), (fs) => {
+        expect(fs).to.deep.equal({
+          'en-us.json': `{"greet":"hello"}`,
+        });
+      });
+    });
+
+    it('should be case-insensitive for includeLocales', function () {
+      const subject = new TranslationReducer(this.input.path(), {
+        includeLocales: ['en-us', 'zh-CN'],
+        errorOnMissingTranslations: true,
+      });
+
+      this.input.write({
+        'de-DE.json': `{ "greet": "hallo" }`,
+        'en-US.json': `{ "greet": "hello" }`,
+        'zh-CN.json': `{ "greet": "你好" }`,
+      });
+
+      return build(createBuilder(subject), (fs) => {
+        expect(fs).to.deep.equal({
+          'en-us.json': `{"greet":"hello"}`,
+          'zh-cn.json': `{"greet":"你好"}`,
+        });
+      });
+    });
+  });
+
+  describe('excludeLocales', function () {
+    it('should exclude specified locales', function () {
+      const subject = new TranslationReducer(this.input.path(), {
+        excludeLocales: ['en-us'],
+        errorOnMissingTranslations: true,
+      });
+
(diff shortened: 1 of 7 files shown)


tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
```diff
@@ -55,3 +55,20 @@ or
 
 `{{format-number numCustomers notation="compact"}}`
 
+## **New Features**
+
+### **Filter Locales**
```
jasonmit, line 60: Not something that should go in the migration docs.  Can you find a new home for it in the docs - maybe a new section for config options (I likely should have added one but doesn't look like I have)?

I can fill in the other config options before versioning.

houfeng0923: Ok, i will update docs

tests/dummy/app/pods/docs/cookbook/addon-configs/template.md
```diff
@@ -0,0 +1,18 @@
...
+# Addon Configs
+
+## Filter Locales
+
+If you need to distribute applications separately in different languages, now we can use `includeLocales` or `excludeLocales` configuration options.
+
+```
+// config/ember-intl.js
+{
+  includeLocales: ['en-us', 'zh-cn'],
```
jasonmit, line 11: could/should this be case insensitive?

lib/broccoli/filter-translation-tree.js
```diff
@@ -0,0 +1,22 @@
+const funnel = require('broccoli-funnel');
+
+function filterTranslationTree(tree, includeLocales, excludeLocales) {
```
jasonmit, line 3: Can we add tests for this

tests/dummy/app/pods/docs/cookbook/addon-configs/template.md
houfeng0923: em, you are right . `normalize-locale`

houfeng0923: hi,@jasonmit , i see the config in broccoli-funnel, include /exclude glob string no support case-insensitive (not use minimatch `nocase` )
so , this case , can we use it like below ？
```
includeLocales: ['en-*', 'zh-*'] 
```

or i need refactor this pr , not a new filter tree , but modify translation-reducer .

lib/broccoli/filter-translation-tree.js
jasonmit: https://github.com/ember-intl/ember-intl/blob/master/tests-node/unit/broccoli/translation-reducer/index-test.js

houfeng0923:
@jasonmit, i review `broccoli-caching-writer` lib, and push a new commit for better filter and cache .
and the problem of case insensitive is resolved now.

tests-node/unit/broccoli/translation-reducer/index-test.js
```diff
@@ -36,15 +36,15 @@ describe('translation-reducer', function () {
...
+        'en-us.json': `{ "greet": "你好" }`,
       });
 
       return build(createBuilder(subject), (fs) => {
         expect(fs).to.deep.equal({
           custom: {
             output: {
               path: {
-                'en-us.json': `{"greet":"hi"}`,
+                'en-us.json': `{"greet":"你好"}`,
```
jasonmit, line 47: Can we revert this and line 39.  Other than that, looks good to go!  Thanks

houfeng0923: 😆，sorry， i mad a mistake when replace txt

jasonmit:
Thanks I need to update the docs then will version!

jasonmit:
Versioned as part of 5.3.0, thanks!

PR #1359, opened 2020-06-11
fix Symbol in ie

houfeng0923:
ie not support native Symbol. and we are sensitive to package size, so can we make this const (`MISSING_TRANSLATION`) to String instead.


jasonmit:
Fixed in 5.2.2

PR #1363, opened 2020-06-20
fix invalid import

NullVoxPopuli:
there is no default import in intl-messageformat-parser :shrug: 
```js
// cat node_modules/intl-messageformat-parser/lib/index.js
import { pegParse } from './parser';
import { normalizeHashtagInPlural } from './normalize';
export * from './types';
export * from './parser';
export * from './skeleton';
export function parse(input, opts) {
    var els = pegParse(input, opts);
    if (!opts || opts.normalizeHashtagInPlural !== false) {
        normalizeHashtagInPlural(els);
    }
    return els;
}

// $ cat node_modules/intl-messageformat-parser/dist/index.js
"use strict";
function __export(m) {
    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
}
Object.defineProperty(exports, "__esModule", { value: true });
var parser_1 = require("./parser");
var normalize_1 = require("./normalize");
__export(require("./types"));
__export(require("./parser"));
__export(require("./skeleton"));
function parse(input, opts) {
    var els = parser_1.pegParse(input, opts);
    if (!opts || opts.normalizeHashtagInPlural !== false) {
        normalize_1.normalizeHashtagInPlural(els);
    }
    return els;
}
exports.parse = parse;
```

### DIFF
--- a/addon/-private/utils/parse.js
+++ b/addon/-private/utils/parse.js
@@ -1,11 +1,11 @@
-import IntlMessageFormat from 'intl-messageformat-parser';
+import { parse } from 'intl-messageformat-parser';
 
 /**
  * @private
  * @hide
  */
 export default function parseString(string) {
-  return IntlMessageFormat.parse(string, {
+  return parse(string, {
     normalizeHashtagInPlural: false,
   });
 }


jasonmit:
Strange that this ever worked, LGTM - will version shortly.

NullVoxPopuli:
embroider caught it. 🤷 
maybe there is something the default build system does to try to be more lenient? idk

jasonmit:
Versioned in 5.2.2, thanks a lot!

PR #1365, opened 2020-06-27
fix: Schedules DOM `lang` update after localeChanged is dispatched

### LINKED ISSUE #1239: HTML lang tag set too early

### Environment

- Ember Version: 3.17.0 in twiddle (but I encountered it on 3.16.6 myself)
- Ember CLI Version: N/A (3.16.1 myself)
- Ember Intl Version: 4.3.1
- Browser(s): Chrome
- Node Version: N/A for twiddle (12.14.1 myself)

### Steps to Reproduce
See this twiddle: https://ember-twiddle.com/39138ab576de35bb8b09dc8fc40ade45?openFiles=twiddle%5C.json%2C

When clicking the button you toggle between FR and EN locales. With FR having a red date background. The date format is obviously different between them. To better show the problem I added an extra delay to make sure it is noticed.
When you switch to FR the lang tag is set and thus the css is applied, but the translations are not done yet. So you see the FR css with the EN 'content'. 

I think this is because the translations are done on the next ember run? I guess that is because the click is handled in the 'action' section of the run loop. And the locale change will only trigger rerender on the next run loop.
So maybe the lang attribute should also be set on the next run. And maybe even on a different run loop part? https://api.emberjs.com/ember/3.17/functions/@ember%2Frunloop/schedule with afterRender maybe?

---edit:
If I move the update inside the already existing run.next it fixes my render problem locally
```js
        this._timer = next(() => {
          this.trigger('localeChanged');
          this.updateDocumentLanguage(this._locale);
        });
```

jlami:


jasonmit:
Fixed in 4.4.1 and 5.3.1 - thanks!

PR #1399, opened 2020-08-06
fix: corrects public lookup() behavior to return original string instead of intl-messageformat ast.

### LINKED ISSUE #1395: `IntlService#lookup()` return value changed in v5
https://ember-intl.github.io/ember-intl/docs/guide/service-api#lookup-translationkey-string-optionallocale-string-array-string-optionaloptions-object-string-undefined

We're currently upgrading our app from v4 to v5 and I found that the return value of `IntlService#lookup()` has changed.

https://github.com/ember-intl/ember-intl/blob/322d325733146b71d77c5aa6d1724ed514f0b07c/addon/services/intl.js#L116-L136

It used to return the raw translation string, but now it returns an array like:

```js
[{ type: 0, value: "Some translation" }]
```

Was this change intentional? I don't think it is.

jasonmit:
fixes #1395

### DIFF
--- a/addon/-private/store/container.js
+++ b/addon/-private/store/container.js
@@ -15,38 +15,60 @@ export default EmberObject.extend({
     this._translations = new Map();
   },
 
-  lookupTranslationObject(localeName, autogenerate = true) {
-    if (this._translations.has(localeName) || !autogenerate) {
-      return this._translations.get(localeName);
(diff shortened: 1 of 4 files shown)


buschtoens:
Awesome, thank you! ✨

PR #1401, opened 2020-08-06
fix(formatter): options take precedence over named format

buschtoens:
This seems to be a behavioral change between v4 / v5. We noticed because in one place we invoke `formatNumber()` like so and our tests broke:

```ts
this.intl.formatNumber(12345, { format: 'EUR', minimumFractionDigits: 0 }); 
```

```ts
// formats.js
export default {
  // ...
  number: {
    EUR: {
      style: 'currency',
      currency: 'EUR',
      minimumFractionDigits: 2,
      maximumFractionDigits: 2
    },
    // ...
  }
};
```

In v4 this returned `12345 €`. In v5 it returns `12345,00 €`.

This is because the `namedFormatsOptions` take precedence over the explicitly specified `formatterOptions`. I would assume this change was accidental, as I would expect explicitly specified options to always override named format options.

### DIFF
--- a/addon/-private/formatters/-base.js
+++ b/addon/-private/formatters/-base.js
@@ -55,8 +55,8 @@ export default class FormatterBase {
       const namedFormatsOptions = this.getNamedFormat(formatOptions.format);
 
       formatterOptions = {
-        ...formatterOptions,
         ...namedFormatsOptions,
+        ...formatterOptions,
       };
     }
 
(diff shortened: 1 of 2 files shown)


jasonmit:
Thanks for catching that.  npm is having some issues right now with scoped packages: https://status.npmjs.org/

Otherwise, PR looks goods - merge whenever.

Issue #1400, opened 2020-08-06
Rewrite service api docs


jasonmit:
The docs aren't all that helpful, arguments should be explained more.  Should reveal interfaces for objects passed as arguments to the service methods.

yuri-semenenko:
Where are the old docs?

jasonmit:
@yuri-semenenko same place they were: https://ember-intl.github.io/ember-intl/versions/v4.4.0/

MrChocolatine:
I agree.

I was about to ask you to create the documentation for `formatDate()`:
https://ember-intl.github.io/ember-intl/versions/master/docs/guide/service-api#formatdate-value-date-number-string-optionaloptions-object-string

Thank you.

jasonmit:
@MrChocolatine here are the options `formatTime`/`formatDate` supports:
https://github.com/ember-intl/ember-intl/blob/master/addon/-private/formatters/format-date.js#L9-L24

You can read about each option:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#Description

buschtoens:
I'd like to base this on #1412 which migrates to TypeScript. We could then use [TypeDoc](https://typedoc.org/).

jasonmit:
@buschtoens I do like our document format, know of any ways to integrate TypeDoc into it some how?  My thinking is, TypeDoc would be used for covering the interfaces for the public APIs and the rest of the docs would remain the same.  Let me know if you have any ideas.

buschtoens:
Unfortunately there's no beaten path for integrating TypeDoc into `ember-cli-addon-docs`. There is `ember-cli-addon-docs-esdoc` for which I added very primitive TS support in https://github.com/ember-learn/ember-cli-addon-docs-esdoc/pull/7. It might already be sufficient, but I doubt it.

We can also try keeping both the old API docs and add the new TypeDoc at a separate route. I'll try looking into it. I need to fix the docs deployment anyway still. 😄

ijlee2:
I'll close this issue as there isn't a concrete, achievable plan to replace `ember-cli-addon-docs` for the docs app. Until we have more active contributors and maintainers, I think we can consider `ember-cli-addon-docs` to be sufficient (good enough).

PR #1412, opened 2020-08-15
refactor: migrate to TypeScript

buschtoens:
Previous attempts of migrating to TypeScript were abandoned, because they would have broken compatibility with pre-3.4 Ember.js versions, as they rely on the [`ember-native-class-polyfill`](https://github.com/pzuraq/ember-native-class-polyfill).

This PR only migrates code that _does not_ make use of the `EmberObject` system. For code that does use `EmberObject`, we provide `.d.ts` files instead.

The types are definitely not _awesome_ yet, but I think this is a solid foundation to iterate on. It will also allow us to better make use of DocBlock comments, which is great for in-IDE docs. We could evaluate using TypeDoc instead of the standard `ember-cli-addon-docs` API extractor for the web docs as well.


addon/-private/formatters/format-date.ts
```diff
@@ -21,33 +21,33 @@ const DATE_TIME_OPTIONS = [
...
+    value: ConstructorParameters<typeof Date>[0],
+    formatOptions: Intl.DateTimeFormatOptions & BaseOptions
+  ): string {
     const formatterOptions = this.readOptions(formatOptions);
 
     this.validateFormatterOptions(locale, formatterOptions);
     const formatterInstance = this.createNativeFormatter(locale, formatterOptions);
 
-    return formatterInstance.format(new Date(value), formatOptions);
+    return formatterInstance.format(new Date(value));
```
buschtoens, line 51: According to MDN, there is no second parameter:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format#Syntax

addon/-private/formatters/format-number.ts
```diff
@@ -27,33 +27,29 @@ const NUMBER_OPTIONS = [
...
-
-  format(locale, value, formatOptions) {
+  format(locale: string | string[], value: number, formatOptions: Intl.NumberFormatOptions & BaseOptions): string {
     const formatterOptions = this.readOptions(formatOptions);
 
     this.validateFormatterOptions(locale, formatterOptions);
     const formatterInstance = this.createNativeFormatter(locale, formatterOptions);
 
-    return formatterInstance.format(value, formatOptions);
+    return formatterInstance.format(value);
```
buschtoens, line 53: According to MDN, there is no second parameter:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format#Syntax

addon/-private/formatters/format-time.ts
```diff
@@ -9,6 +9,8 @@ import FormatDateFormatter from './format-date';
  * @private
  * @hide
  */
+// @ts-ignore We could use a proper intermediary abstract class, but it's
+// unnecessarily complicated.
 export default class FormatTime extends FormatDateFormatter {
-  static type = 'time';
+  static readonly type = 'time';
```
buschtoens, line 15: Redefining `type` is not allowed, as it is `readonly`. If we want to avoid `@ts-ignore` here, we could either remove `readonly`, which would remove some safety, or add another intermediary abstract `DateTimeFormat` base class.

In the long run, it would probably be more sensible to just remove the distinction between `time` and `date` formats, as they both use the same formatter anyway.

jasonmit: Ya I can see in the future time and date being merged together.  The main benefit is the separation of the named formats and consistency between [react-intl](https://formatjs.io/docs/react-intl/components#formattedtime) and ember-intl.

jasonmit:
Did a quick pass, will do another one tonight and likely approve.  Thanks again, you're amazing 🙏

buschtoens:
Thank you 🙈 

Some of the types are really whack / overly complex, but this often happens, when you start out with a vanilla JS code base and want to retrofit TypeScript on top of it. 😅 
I didn't want to do any significant refactors, to keep the risk as low as possible, considering the scope of this PR. 🙈 

In follow-up PRs I would like to:

- Migrate the dummy app and tests to TypeScript, so that we can dog-food our own types. I'll probably have time for this tomorrow, so that we can do it before the next release, to prove the overall type validity in CI.
- Add DocBlock comments for all public API and try to get it into the docs (#1400).

jasonmit:
> Some of the types are really whack / overly complex, but this often happens, when you start out with a vanilla JS code base and want to retrofit TypeScript on top of it. 😅

Agree, we'll clean things up as we go - no need to do it in one sweep.

Looking forward to the test migration and doc PR 🙏

addon/-private/formatters/format-message.ts
```diff
@@ -46,18 +52,23 @@ function escapeOptions(object) {
 /**
  * @private
  * @hide
+ * @TODO Should not extend from abstract base class.
```
jasonmit, line 55: Agree, can be a future PR that I can pick up once this lands

addon/services/intl.d.ts
```diff
@@ -0,0 +1,95 @@
+import Evented from '@ember/object/evented';
```
jasonmit, line 1: > In order for your addon's users to benefit from type information from your addon, you need to put .d.ts declaration files at the location on disk where the compiler expects to find them. This addon provides two commands to help with that: ember ts:precompile and ember ts:clean. The default ember-cli-typescript blueprint will configure your package.json to run these commands in the prepublishOnly and postpublish phases respectively, but you can also run them by hand to verify that the output looks as you expect.

Even if we convert all our public modules to ts that we would need to still maintain declaration files?  Am I reading this correct?

jasonmit:
> In order for your addon's users to benefit from type information from your addon, you need to put .d.ts declaration files at the location on disk where the compiler expects to find them. This addon provides two commands to help with that: ember ts:precompile and ember ts:clean. The default ember-cli-typescript blueprint will configure your package.json to run these commands in the prepublishOnly and postpublish phases respectively, but you can also run them by hand to verify that the output looks as you expect.

Going through ember-cli-typescript to familiarize myself and I'm not seeing the pkg.json changes that the blueprint emits.  Should we include them this way we get declaration file generation?

package.json
```diff
@@ -24,13 +24,17 @@
...
+    "lint:types": "tsc --noEmit",
     "lint-staged": "lint-staged",
     "tests-node": "mocha \"tests-node/**/*.js\"",
     "test": "yarn run tests-node && ember test",
     "test-node": "yarn tests-node",
     "testall": "yarn run tests-node && ember try:each",
-    "prettier": "prettier --single-quote --print-width 120 --write \"{blueprints,config,lib,app,addon,addon-test-support,tests,tests-node}/**/*.js\""
+    "prettier": "prettier --single-quote --print-width 120 --write \"{blueprints,config,lib,app,addon,addon-test-support,tests,tests-node}/**/*.js\"",
+    "prepack": "ember ts:precompile",
+    "postpack": "ember ts:clean"
```
buschtoens, line 37: @jasonmit this is where the magic happens :)

prepack / postpack work more reliably than pre/post-publish and will be used by ember-cli-typescript@4 by default.

jasonmit: Weird - I didn't see this in the github diff, sorry!

buschtoens:
I've released this as `v5.5.0-beta.1` and will give it a whirl in our apps at work.

Then before we release this as `latest` stable, I would like to migrate our tests to TypeScript first.

chriskrycho:
This is awesome! I strongly recommend, as a follow-on, adding [`expect-type`](https://github.com/mmkal/ts/tree/master/packages/expect-type) and using it to write type tests, and integrating those into CI. That will help you know that you're maintaining the public API contract of your types. See [ember-modifier](https://github.com/ember-modifier/ember-modifier) and [ember-concurrency](https://github.com/machty/ember-concurrency/) for good examples of that, and feel free to reach out to us in `#e-typescript` on Discord if you hit any issues!

buschtoens:
Thank you for the great suggestion! We already discovered and fixed quite a few mismatches by converting all tests to TypeScript. But I will definitely look into integrating this as well.

chriskrycho:
Yeah, the key win you'll get from integrating `expect-type` or similar is that you won't accidentally have a type widen to `any`, for example, or narrow in a way that still happens to pass the tests but nonetheless broke your public API. For a deeper dive on the ways that you can end up with breaking changes in published types, see [RFC: Defining Semantic Versioning for Published Addon Types – Defining Breaking Changes](https://github.com/typed-ember/ember-cli-typescript/blob/rfc-type-stability/rfcs/0000-type-stability.md#defining-breaking-changes) over in the ember-cli-typescript repo (we should hopefully be landing that RFC in the next week or two, as I finally have time to finish revising it!).

PR #1418, opened 2020-08-17
refactor(test-support): migrate to TypeScript

buschtoens:
I completely forgot the `addon-test-support` in #1412. Installing the beta in our @ClarkSource monorepo and removing our custom types made that very apparent. 😅


addon-test-support/-private/pick-last-locale.ts
```diff
@@ -1,4 +1,7 @@
+// @ts-ignore We don't won't to bring along extra baggage, when installed in a
```
buschtoens, line 1:
```suggestion
// @ts-ignore We don't want to bring along extra baggage, when installed in a
```

addon-test-support/-private/serialize-translation.ts
```diff
@@ -1,3 +1,5 @@
+// @ts-ignore We don't won't to bring along extra baggage, when installed in a
```
buschtoens, line 1:
```suggestion
// @ts-ignore We don't want to bring along extra baggage, when installed in a
```

addon-test-support/add-translations.ts
```diff
@@ -0,0 +1,35 @@
+import { get } from '@ember/object';
+import makeIntlHelper from './-private/make-intl-helper';
+import pickLastLocale from './-private/pick-last-locale';
+import type IntlService from 'ember-intl/services/intl';
+import type { Translations } from 'ember-intl/-private/store/translation';
+
+// ! Because TypeScript seems to short-circuit overloaded functions when passed
+// as generics, including these overloads, would not work with `makeIntlHelper`.
```
buschtoens, line 8:
```suggestion
// as generics, including these overloads would not work with `makeIntlHelper`.
```

Issue #1396, opened 2020-08-04
Using variables in HTML attributes


buschtoens:
v4 did not require special escaping of HTML tags, e.g. you could have a translation string like:

```yml
legal: Please agree to our <a href="{url}">Terms & Conditions</a>
```

```hbs
{{t "legal" url="/legal/terms-and-conditions" htmlSafe=true}}
```

According to [Format HTML Message](https://ember-intl.github.io/ember-intl/versions/master/docs/helpers/t#format-html-message) and the [v4 -> v5 migration guide](https://ember-intl.github.io/ember-intl/versions/master/docs/guide/migration-4-0-to-5-0#translations) HTML tags now need to be escaped in single quotes (`'`), like so:

```
'<strong>'{name}'</strong>'
```

This does not seem to work, if the escaped HTML sequence contains variables itself:

```yml
legal: Please agree to our '<a href="{url}">'Terms & Conditions'</a>'
```

In v5 this would output:

```html
Please agree to our <a href="{url}">Terms & Conditions</a>
```

Instead of the expected:

```html
Please agree to our <a href="/legal/terms-and-conditions">Terms & Conditions</a>
```

jasonmit:
Does this work:

`legal: Please agree to our '<'a href="{url}"'>'Terms & Conditions'</a>'`

Seems related to https://github.com/ember-intl/ember-intl/issues/1336

buschtoens:
Works perfectly! Thanks for the tech support. 😅 
I'll update the docs!

jasonmit:
Just a heads up, soon the new translation escaping behavior in 5.x will soon no longer be necessary.

The v4 behavior will be restored in a backwards compatible way.  For those that have already went through the process of escaping everything, you will not have to revert back - both formats will continue to be supported.

https://github.com/ember-intl/ember-intl/pull/1420

Issue #1350, opened 2020-06-03
Upgrade to 5.1.1 - Issue when rendering tags with attributes


MrChocolatine:
- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.12
- Ember CLI Version: 3.12
- Ember Intl Version: 5.1.1
- Browser(s): Chromium Version 81.0.4044.129
- Node Version: 12.16.3

### Steps to Reproduce

My issue seems to be related to #1336 but the problem is not exactly the same, also I tried the provided solution and it didn't work in my case.

I am currently upgrading `ember-intl` from `4.2.3` to `5.1.1` and your new HTML escaping seems buggy and/or badly documented.

Before the upgrade, this was working perfectly :
```
-- HBS file
{{t "general.forms.required_fields" htmlSafe=true}}

-- YAML file
general:
  forms:
    required_fields: Fields marked <strong class="primary">*</strong> are required.
```

Render :
![01 ember-intl-ok](https://user-images.githubusercontent.com/47531779/83636216-7cc5ae80-a59d-11ea-9ab8-ac3ff3c9ea79.png)

---

According to your [your upgrade guide](https://ember-intl.github.io/ember-intl/docs/guide/migration-4-0-to-5-0) we now need to escape HTML tags, fine, let's do it the recommended way ([also here](https://ember-intl.github.io/ember-intl/docs/helpers/t#format-html-message)):
```
-- HBS file
{{t "general.forms.required_fields" htmlSafe=true}}

-- YAML file
general:
  forms:
    required_fields: Fields marked '<strong class="primary">'*'</strong>' are required.
```

Render :
![02 ember-intl-not ok](https://user-images.githubusercontent.com/47531779/83636415-db8b2800-a59d-11ea-8e23-8de9a5928ae4.png)

Hum... ?

---

Let's try again without involving too many quotes:
```
-- HBS file
{{t "general.forms.required_fields" htmlSafe=true}}

-- YAML file
general:
  forms:
    required_fields: Fields marked '<strong>'*'</strong>' are required.
```

Render :
![03 ember-intl-not ok](https://user-images.githubusercontent.com/47531779/83636590-1e4d0000-a59e-11ea-996f-c8afa5223896.png)

So I still have the same issue.

---

Let's try again :
```
[...]

Issue #1336, opened 2020-05-26
Upgrading to 5.x breaks dynamic values within HTML tags


gzurbach:
- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.4
- Ember CLI Version: 3.4
- Ember Intl Version: 5.1.1
- Browser(s): Chrome 81
- Node Version: 10.16

### Steps to Reproduce

We're in the process of upgrading `ember-intl` to v5. We have numerous URLs in our app that are passed as parameters in strings. The example below used to work properly until the upgrade:

```hbs
{{t "register.terms" termsUrl="https://company.com/en/terms" htmlSafe=true}}
```

```json
{
  "register": {
    "terms": "By signing up, you agree to the '<a href=\"{termsUrl}\">'terms of service'</a>' and ..."
  }
}
```

Unfortunately in v5 `{termsUrl}` is not detected anymore, likely because of the new escaping. The `<a>` tag literally contains `<a href="{termsUrl}">`. Is there a solution that would allow us to keep passing URLs as parameters within HTML tags?

Thanks in advance for your insights, and for this great library!

jasonmit:
Thanks for opening the issue, below should resolve your issue.  I should update the docs to include this example since it's definitely not immediately clear.

```json
{
  "register": {
    "terms": "By signing up, you agree to the '<'a href=\"{termsUrl}\"'>'terms of service'</a>' and ..."
  }
}
````

gzurbach:
@jasonmit Yep that works! 👍  

I'm pretty sure I tried to do this last night but I was probably to tired. Thanks a a bunch for your help, our upgrade is now complete.

jasonmit:
Just a heads up, soon the new translation escaping behavior in 5.x will soon no longer be necessary.

The v4 behavior will be restored in a backwards compatible way.  For those that have already went through the process of escaping everything, you will not have to revert back - both formats will continue to be supported.

https://github.com/ember-intl/ember-intl/pull/1420

Issue #1321, opened 2020-05-21
5.X breaking changes 


dbendaou:
Hello, thanks for the work you've done on ember-intl ! 

Changes that arrived in 5.x are quite huge.
I would like to understand why this change was motivated ? Especially the one about changing format of HTML in translations. 
Furthermore, our copywriters have to learn a new way of adding translations.

Can this change be made as optional with a param given to ember-intl ?

Thanks

jasonmit:
Change was made lower level in the message compiler (intl-messageformat) that I do not control.  They added it to support rich text support in messages, given the nature of Ember templates we cannot really harness that feature (i.e., ability to inject a link component in a translation).

It's a breaking change for a reason though, you're free to upgrade at your leisure or not at all.  Obviously a lot went into this release and this library is now in a much better place as a result.

jasonmit:
https://github.com/formatjs/formatjs/blob/master/packages/intl-messageformat/CHANGELOG.md#breaking-changes

Feel free to open a discussion with them to see if they want to make this configurable.

jasonmit:
Just a heads up, soon the new translation escaping behavior in 5.x will soon no longer be necessary.

The v4 behavior will be restored in a backwards compatible way.  For those that have already went through the process of escaping everything, you will not have to revert back - both formats will continue to be supported.

https://github.com/ember-intl/ember-intl/pull/1420

dbendaou:
Good news! Thanks for the heads up!

PR #1420, opened 2020-08-18
feat: Avoids complex escaping of HTML tags/attributes in translations

### LINKED ISSUE #1419: [Feature Request] Re-enable tags parsed as strings - avoiding text escaping
It looks as though the parser has landed the option [`ignoreTag`](https://formatjs.io/docs/intl-messageformat/#intlmessageformat-constructor).

This will allow ember-intl to go back to the translation format that we were used to in versions prior to 5.0.  It's also backwards compatible for us to turn this on by default 5.x since we never took advantage of the `TagElement` in the compiler.

cc @Turbo87 I saw today this impacted you and was preventing adoption.

jasonmit:
Migrating to v5 was painful for some because the translation syntax changed a bit around escaping tags in translations.  The compiler and parser have since been updated to support the translation syntax that we've been used to in every version up to 5.0.0.

This is backwards compatible, so anyone who put in the effort to migrate their translations to the new syntax does not need to migrate back.  Both styles will now be supported.

Resolves #1419

Related to #1350, #1396, #1336, #1321, #1289

### DIFF
--- a/addon/-private/formatters/format-message.ts
+++ b/addon/-private/formatters/format-message.ts
@@ -58,7 +58,9 @@ export default class FormatMessage extends Formatter<Options> {
   // @ts-ignore This class does not match the abstract base class.
   createNativeFormatter = memoize(
     (ast: TranslationAST, locales: string | string[], formatConfig?: Partial<Formats>) => {
-      return new IntlMessageFormat(ast as any, locales, formatConfig);
+      return new IntlMessageFormat(ast as any, locales, formatConfig, {
+        ignoreTag: true,
+      });
     }
   );
 
--- a/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
+++ b/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
@@ -6,6 +6,10 @@ The `locales`, `disablePolyfill`, and `autoPolyfill` configuration options in `c
 
 ## **Breaking Changes**
 
+### **Translations**	
+
+* Improved ICU-spec compliance, special characters are now escaped via a single quote `'` instead of a slash `\`
+
 ### **Node runtime**
 
 We now support down to Node 10, dropping support for Node 8.
--- a/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
+++ b/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
@@ -8,7 +8,7 @@ The `locales`, `disablePolyfill`, and `autoPolyfill` configuration options in `c
 
 ### **Translations**	
 
-* Improved ICU-spec compliance, special characters are now escaped via a single quote `'` instead of a slash `\`
+Improved ICU-spec compliance, special characters are now escaped via a single quote `'` instead of a slash `\`
 
 ### **Node runtime**
 
--- a/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
+++ b/tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
@@ -34,24 +34,6 @@ Changes:
 * `interval` was removed from the format-relative helper
 * `now` was removed from the format-relative helper
 
-### **Translations**
-
-* Special character escaping in translations is now done via a single quote, `'` instead of the previous slash `\`.
-
-_This is not to be confused with string escaping in YAML or JSON where you'll need to use `\` to escape quotes to keep a string together.  This change is specifically about escaping special characters and HTML tags within your ICU message._
-* Additionally, all HTML tags now need to be escaped.
-
-`<strong>{name}</strong>`
-
-becomes
-
-`'<strong>'{name}'</strong>'`
-
-If you use placeholders _inside_ of an HTML tag for attributes, escape only the
-angle brackets, like so:
-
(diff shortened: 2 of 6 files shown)


addon/-private/formatters/format-message.ts
```diff
@@ -58,7 +58,9 @@ export default class FormatMessage extends Formatter<Options> {
   // @ts-ignore This class does not match the abstract base class.
   createNativeFormatter = memoize(
     (ast: TranslationAST, locales: string | string[], formatConfig?: Partial<Formats>) => {
-      return new IntlMessageFormat(ast as any, locales, formatConfig);
+      return new IntlMessageFormat(ast as any, locales, formatConfig, {
```
jasonmit, line 61: @buschtoens curious if you know why `any` is needed since `TranslationAST` is typed to `MessageFormat[]` which is the expected type.

tests/dummy/app/pods/docs/guide/migration-4-0-to-5-0/template.md
```diff
@@ -34,24 +34,6 @@ Changes:
 * `interval` was removed from the format-relative helper
 * `now` was removed from the format-relative helper
 
-### **Translations**
-
-* Special character escaping in translations is now done via a single quote, `'` instead of the previous slash `\`.
```
jasonmit, line 39: I need to restore this one sentence.  Escaping still changed to be more ICU spec compliant - it's just no longer necessary for HTML tags and attributes.

Turbo87:
nice, thank you!

Alonski:
Amazing 👏 !!! ❤️

addon/-private/formatters/format-message.ts
Alonski: So I took a look at this and it seems like these types aren't identical:
```ts
import type { MessageFormatElement } from 'intl-messageformat-parser';
import type { MessageFormatElement } from 'intl-messageformat/intl-messageformat';
// The above is not an actual exported type.
```
They look identical from a quick look but might be an issue with them somewhere.

jasonmit: Thanks for investigating @Alonski!

buschtoens: Correct! I had that one hit me as well in #1412! But I also didn't look into it deeper yet.

tests/unit/helpers/format-message-(html)-test.js
```diff
@@ -25,14 +25,23 @@ module('format-message (html)', function (hooks) {
...
       '<strong>Welcome <em>Alexander</em>!</strong>'
     );
+
+    assert.equal(
+      this.intl.formatMessage(`<strong>Welcome {name}!</strong>`, {
+        htmlSafe: true,
+        name: htmlSafe('<em>Alexander</em>'),
+      }).string,
+      '<strong>Welcome <em>Alexander</em>!</strong>'
+    );
```
buschtoens, line 44: Should we for all these also assert for `isHTMLSafe(str)` or is this sufficiently covered in other tests already?

buschtoens: I would probably rather use `String(...)` or `.toString()` over `.string`, but I don't see this as an issue.

https://github.com/emberjs/ember.js/blob/9ad9b7b2ff80e1100f6b278171412f4b558b111b/packages/%40ember/-internals/glimmer/lib/utils/string.ts#L5-L19

I also made it a habit (and use a lint rule for it in my projects) to always use `assert.strictEqual()` over `assert.equal()`, but again, not important.

addon/-private/formatters/format-message.ts
buschtoens: We can fix this in a follow-up PR. :)

buschtoens:
I'll go ahead and merge this, as I found some issues with the types related to `format-message`. Basically, TS strips the `// @ts-ignore` when generating `.d.ts` files, which means that the type errors stemming from the incorrect extension of the abstract base class are not suppressed, when `ember-intl` is used as a dependency.

tests/unit/helpers/format-message-(html)-test.js
```diff
@@ -52,24 +61,41 @@ module('format-message (html)', function (hooks) {
...
+      this.intl.formatMessage(`'<'a href="{link}"'>'text'</a>'`, {
+        htmlSafe: true,
+        link: 'http://formatjs.io',
+      }).string,
+      '<a href="http://formatjs.io">text</a>'
+    );
 
-    assert.equal(output, '<a href="http://formatjs.io">text</a>');
+    assert.equal(
+      this.intl.formatMessage(`<a href="{link}"'>text</a>`, {
```
buschtoens, line 75: Prompted by #1443 I wanted to give this another read and just noticed this now.
I think the `'` is incorrect and might be masquerading a bug.

```suggestion
      this.intl.formatMessage(`<a href="{link}">text</a>`, {
```

jasonmit: Ya looks like it shouldn't be there.  Without it, the test is failing?

buschtoens: Unfortunately (?) the test does not fail. :/

I started #1444.

yoavfranco:
Hey @jasonmit 
Thanks for this!
It might be worth it to update the migration guide https://ember-intl.github.io/ember-intl/docs/guide/migration-4-0-to-5-0
Since it still lists this as a breaking change 
WDYT?

PR #1422, opened 2020-08-18
fix: remove public `// @ts-ignore`

buschtoens:
As mentioned in https://github.com/ember-intl/ember-intl/pull/1420#issuecomment-675496087, TypeScript strips all `// @ts-ignore` comments when generating the `.d.ts` files. If such a comment is used on a public interface to suppress an error, e.g. class methods, then the comment will be missing when this package is installed as a dependency and causes the host project to fail TypeScript type checking.

This PR applies some minor refactors, which were suggested in the original #1412 PR already, to avoid the usage of `// @ts-ignore` on public interfaces.


addon/-private/formatters/format-message.ts
```diff
@@ -52,19 +52,29 @@ function escapeOptions<T extends Record<string, unknown>>(object?: T) {
...
+
+    // NOTE: a fn since we lazily grab the formatter from the config
+    // as it can change at runtime by calling intl.set('formats', {...});
+    this.readFormatConfig = config.readFormatConfig;
+  }
+
   createNativeFormatter = memoize(
     (ast: TranslationAST, locales: string | string[], formatConfig?: Partial<Formats>) => {
-      return new IntlMessageFormat(ast as any, locales, formatConfig, {
+      return new IntlMessageFormat(ast, locales, formatConfig, {
```
buschtoens, line 72: https://github.com/ember-intl/ember-intl/pull/1420#discussion_r471896783

Weirdly, the issue is not occuring any more. Maybe there was a dependency version mismatch?

buschtoens: I upgraded the intl-related deps in 1214587e493fdf08f4570e7b7cc3cc11ff2f1ad0 just be _really_ sure 😅

PR #1430, opened 2020-08-20
Mark `htmlSafe` param as optional

vitch:
I ran into a couple of issues when calling `formatMessage` and not passing `htmlSafe` as part of the `data` object.

 1. The typings expect `htmlSafe` to be part of the `options` (which is also where you pass the data you are interpolating into the string)
 2. The return type is `SafeString | string` which requires casting to `string` in my case (where I don't pass `htmlSafe: true`)

This PR marks `htmlSafe` as optional and so fixes 1. I started playing around with function overloads to try and get the correct return type depending on the passed value of `htmlSafe` but didn't get that working yet.

Let me know your thoughts on the above and if you're interested I can try to look into a solution for 2.

### DIFF
--- a/addon/-private/formatters/format-message.ts
+++ b/addon/-private/formatters/format-message.ts
@@ -78,7 +78,7 @@ export default class FormatMessage {
   format(
     locale: string | string[],
     maybeAst: string | TranslationAST,
-    options?: Partial<Record<string, unknown>> & { htmlSafe: boolean }
+    options?: Partial<Record<string, unknown>> & { htmlSafe?: boolean }
   ) {
     let ast = maybeAst as TranslationAST;
 


buschtoens:
I'll look at the SafeString return value issue. :+1:

vitch:
Thanks!

vitch:
I did wonder whether the `options` should be separate to the `data` that you are passing to get merged into the template. That way if you ever needed to have `{htmlSafe}` in your message string it would work. And the types would probably be simpler. But that would need to be a backward incompatible change I guess :(

buschtoens:
Yeah, and it still would leave the ambiguity in the `{{t}}` helper, as that one only accepts a single named parameters object.

buschtoens:
@vitch Fixing in #1431 🥳

vitch:
Thank you ❤️ 

I tried something very similar last night but it was late and I guess I missed a bit somewhere :)

PR #1435, opened 2020-08-25
chore: removes use of deprecated getWithDefault

### LINKED ISSUE #1434: Remove use of deprecated getWithDefault() API
`getWithDefault()` has been deprecated in Ember 3.21.

`ember-intl` uses this function in one place that would be good to update in order to avoid deprecation warnings:

https://github.com/ember-intl/ember-intl/blob/0696e4dbe0a43f247f239cc5fe22a7646ceb97a4/addon/-private/helpers/-format-base.js#L34

jasonmit:
Fixes #1434

### DIFF
--- a/addon/-private/helpers/-format-base.js
+++ b/addon/-private/helpers/-format-base.js
@@ -30,7 +30,7 @@ const AbstractHelper = Helper.extend({
 
   compute([value], options) {
     if (isEmpty(value)) {
-      if (options.allowEmpty || this.allowEmpty) {
+      if (options.allowEmpty ?? this.allowEmpty) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


addon/-private/helpers/-format-base.js
```diff
@@ -31,7 +30,7 @@ const AbstractHelper = Helper.extend({
 
   compute([value], options) {
     if (isEmpty(value)) {
-      if (getWithDefault(options, 'allowEmpty', this.allowEmpty)) {
+      if (options.allowEmpty || this.allowEmpty) {
```
buschtoens, line 33:
```suggestion
      if (options.allowEmpty ?? this.allowEmpty) {
```

Otherwise `options.allowEmpty = false` would not work.

buschtoens (approved):

PR #1442, opened 2020-09-04
chore(deps): upgrade to `typescript@4`

buschtoens:
`ember-cli-typescript@4` got officially released and `typescript@4` is out now as well. We may want to / need to hold back on `typescript@4` or ship different `typesVersions` though. Maybe we should release `typescript@3` first and then add `typescript@4` support in the following version.


buschtoens:
We tried upgrading our code base and it turns out that the current `ember-intl@5.5.0` is incompatible with `typescript@4`, which really motivates me getting this over the finish line, finally. 😅 

The issue is that some of these changes are probably incompatible with `typescript@3`. So either we accept that and drop support (is that a major release then?) or find a way to ship backwards-compatible / extra types.

PR #1444, opened 2020-09-08
fix(translations-reducer): support legacy HTML escape format

### LINKED ISSUE #1443: Unencoded HTML tags still raise error in 5.5.0-beta.7
Hi.  I'm trying to upgrade from ember-intl@3.5 to ember-intl@5.5.0-beta.7.  I'm using the beta version so I don't have to escape any HTML tags in our translated text.  However, even on the beta version I'm still seeing errors for non-escaped HTML tags.  I've reproduced in a brand new empty project using:

* ember-cli: 3.19.0
* node: 12.18.3
* os: darwin x64

My `en-us.yml` file has two entries:

```yml
price_banner: The {product} costs {price, number, USD}
link: <a href="{href}">Click Here</a>
```

My application template looks like this:

```handelbars
{{t 'link' href="www.google.com" htmlSafe=true}}

{{outlet}}
```

When I build my application I get this error:

```bash
Build Error (TranslationReducer)

An error occurred (Expected "#", "'", "\n", "{", argumentElement, double apostrophes, end of input, or tagElement but "<" found.) when extracting ICU arguments for '<a href="www.google.com">Click Here</a>'
```

Based on the tests added in #1420 I think this should work.  Any advice as to what I'm doing wrong would be greatly appreciated.

buschtoens:
Attempts to fix #1443.

### DIFF
--- a/addon/-private/utils/parse.ts
+++ b/addon/-private/utils/parse.ts
@@ -5,6 +5,7 @@ import { parse } from 'intl-messageformat-parser';
  * @hide
  */
 export default function parseString(string: string) {
+  // ! Sync with `lib/message-validator/validate-message.js`
   return parse(string, {
     normalizeHashtagInPlural: false,
     ignoreTag: true,
--- a/addon/-private/utils/parse.ts
+++ b/addon/-private/utils/parse.ts
@@ -5,7 +5,7 @@ import { parse } from 'intl-messageformat-parser';
  * @hide
  */
 export default function parseString(string: string) {
-  // ! Sync with `lib/message-validator/validate-message.js`
+  // ! Sync with `lib/parse-options.js`
   return parse(string, {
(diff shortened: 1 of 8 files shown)


jasonmit:
I think the issue in #1443 is around the [linter's parser](https://github.com/ember-intl/ember-intl/blob/f00a4a65e565025609b8e2f35fc097b05aba670a/lib/broccoli/translation-reducer/linter/extract-icu-arguments.js#L11) likely needs these same options passed in:
https://github.com/ember-intl/ember-intl/blob/master/addon/-private/utils/parse.ts#L8-L11

Haven't confirmed but that's what I think is probably the reasoning.

buschtoens:
Ah yes, of course. 🤦 
I was completely oblivious to the fact, that this is a _build_ error, not runtime error. Thanks for the pointer! Will try to fix.

PR #1449, opened 2020-09-16
feat(helpers): re-expose `-format-base` helper publicly

buschtoens:
This was made private by c63225c1db5c6914bc9658cd7b317390352c0411 in v5.0.0.

I don't see an immediate reason why. Do you recall, @jasonmit?

In Discord @Turbo87 noted that it is quite useful to create custom formatting helpers and in fact I found myself using it too for a `{{format-money}}` helper.

Should we make it public again?

Some apps that updated _may_ import from the `-private` path now (even though they shouldn't). This PR would break them. We could avoid that by adding a re-export there, but I'm not sure whether that's worth it / sensible.


Turbo87:
instead of moving it out of the `-private` folder we could also just reexport the private thing from the `addon/helpers/` folder. might be easier :)

buschtoens:
Agree, in case we want to keep the old `-private` file for naughty apps. If we don't, then I think just moving is a bit less complicated.

jasonmit (approved):

jasonmit:
I am fine with making it public if there's value in it.

buschtoens:
Great! Do you think we should keep the `-private` file and just re-export it in a public place, like @Turbo87 suggested, or remove it?

jasonmit:
@buschtoens I'm fine with moving the file vs re-exporting

PR #1463, opened 2020-11-05
feat(format): add `base` and `defaults`

### LINKED ISSUE #1462: Default formatter options
Hello 👋 

We are currently migrating from using `ember-moment` usage of `moment-format` to the `ember-intl` usage of `format-date`. Previously we could globally set our desired date format via config.

```
moment: {
      outputFormat: 'lll'
}
```

Is there any way to set a default format, or do we have to pass the option at each invocation of the `{{format-date}}`?

buschtoens:
Closes #1462.

Adds `base` and `defaults` format options.

```js
export default {
  // Format options that are always implied. Explicitly specified options when
  // invoking a formatter override these base options.
  base: {
    time: {
      hour: 'numeric',
      minute: 'numeric'
    },
    // date: { },
    // number: { },
    // relative: { },
  },

  // Format options that are implicitly used, if neither a named `format` nor
  // any other know options were specified when invoking a formatter. These
  // are combined with and override the `base` options.
  defaults: {
    time: {
      hour: 'numeric',
      minute: 'numeric'
    },
    // date: { },
    // number: { },
    // relative: { },
  },

  time: {
    hhmmss: {
      hour: 'numeric',
      minute: 'numeric',
      second: 'numeric',
    },
  },
  date: {
    // ...
  },
  relative: {
    // ...
  },
  number: {
    // ...
  },
};
```

### DIFF
--- a/addon/-private/formatters/-base.ts
+++ b/addon/-private/formatters/-base.ts
@@ -22,6 +22,8 @@ export interface BaseOptions {
   format?: string;
 }
 
+type NamedFormats = Omit<Formats, 'base' | 'defaults'>;
+
 /**
  * @private
  * @hide
@@ -31,7 +33,7 @@ export default abstract class FormatterBase<KnownOptions extends {}> {
   protected readonly config: FormatterConfig;
   protected readonly readFormatConfig: () => Formats;
 
-  static type: keyof Formats;
+  static type: keyof NamedFormats;
 
   protected abstract readonly createNativeFormatter: (
     locales: string | string[],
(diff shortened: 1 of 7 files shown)


buschtoens:
@jasonmit What do you think about this API? Too over-engineered? Should we drop either `base` or `defaults`? Or do you like it?

jasonmit:
I can't immediately tell the difference between base and defaults - likely will be confusing for users as well.  I think we should drop one in favor of simplifying the API.

More important, all the formats into `intl-messageformat` since they're used within the translation syntax.  I don't believe `intl-messageformat` would have knowledge of base | default.

I'd expect these two to yield the same results when there's a base/default set:
`{{format-time timestamp}}`
`{{format-message '{timestamp, time}'}}`

If that's already the case (I don't know as I didn't look into the implementation too closely) then just write some tests so that we don't break that behavior in the future.

jasonmit (changes requested):
I love the overall feature but struggled to understand the difference between base and defaults - likely will be confusing for users.  I think we should drop one in favor of simplifying the API (maybe `base`?).

We should also make sure/write a test for supporting using the default key from within a translation string:
`"The time is {timestamp, time, default}"`

I believe we pass the whole formats object to `intl-messageformat` but since `default` is top-level instead of under the individual format options objects (time, date, relative, number) you may need to do a bit of work to achieve that.  To keep things simple, we can move defaults under `time`, `date`, , etc. and release a new major.

elonmulder:
@buschtoens @jasonmit Are there any plans on moving forward with this PR?

ijlee2:
Will close the issue, as the underlying assumptions might have changed since November 2020.

PR #1509, opened 2021-01-19
fix: Add test-support to typesVersions for TS v3

Alonski:
This was missing and caused TS to not recognize test-support functions:
image


buschtoens (approved):
_bueno!_

Issue #1566, opened 2021-05-31
t helper does not accept a hash object directly


SemaLuna:
- [ ] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
  - Not on the latest, but I believe this still applies to the latest version (looking at the code)
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.20.5
- Ember CLI Version: 3.15.2
- Ember Intl Version: 5.5.0
- Browser(s): Any
- Node Version: 12.18.4

### Steps to Reproduce

If I have a translation with multiple placeholders, 

`"myTranslation": "You can either {option1} or {option2}"`

I need to pass them individually, like so:

`{{t myTranslationPath option1='run' option2='hide'}}`

It would be ideal if we could pass the options hash directly, like so:

`{{t myTranslationPath optionsHash}}`

where the `optionsHash` is 

`{option1="run", option2="hide"}`

However - because of how helpers work, any non-named arguments are passed as an array, and those are not picked up by the 't' helper (just get an error saying that I did not pass the placeholders).

In the above example, it's just two arguments - but imagine the case where we have variable translation strings, and thus variable arguments - it's extra legwork that should not be necessary.

Not sure if this is a bug or considered an enhancement (or if there's a good workaround that I am missing).

nolaneo:
I have also run into this in the past. My workaround was to just make a `t2` helper which extends the original but allows for a context hash.

Note this is for version 4.2.0 but the approach would still work in 5.X.X

```js
// This helper simply allows us to spread the inputs
// {{t2 'translation-key' contextObject otherOptionKey='yes'}}

import T from 'my-app-name/helpers/t';
import { isEmpty } from '@ember/utils';

export default T.extend({
  compute([value, context], options) {
    if (isEmpty(value)) {
      if (options.allowEmpty ?? this.allowEmpty) {
        return;
      }

      if (typeof value === 'undefined') {
        throw new Error(`${this} helper requires value attribute.`);
      }
    }

    return this.format(value, options, context);
  },

  format(key, options, context = {}) {
    return this._super(key, { ...context, ...options });
  },
});

```

buschtoens:
Great idea!

Ideally we could solve this using [Spreadable Arguments (aka "splarguments")](https://github.com/emberjs/rfcs/pull/593), however the RFC is not even finished yet.

[...]

PR #1633, opened 2022-01-17
feat(helpers): positional options that override named options

jesdavpet:
This change extends the base abstract helper's `compute` method to accept a positional `options` argument, as per @buschtoens [suggestion](https://github.com/ember-intl/ember-intl/issues/1566#issuecomment-854176540) in the discussion on [issue #1566](https://github.com/ember-intl/ember-intl/issues/1566).

After this change, options may be passed as a positional options argument object:
```handlebars
{{t "say.hello" (hash place="world")}} => "Hello world."  
```

Before this change, options could only be passed as named attributes:
```handlebars
{{t "say.hello" place="world"}} => "Hello world."
```

In the case of mixed use of both positional and named arguments, the named arguments will override properties of the positional argument object where there are duplicate entries:
```handlebars
{{t "say.hello" (hash place="people") place="world"}} => "Hello world."
```

_NOTE: Using `hash` is just to keep the examples brief, the real benefit of this change IMO is the flexibility of passing a dynamic options object down from the component class instead of hard coding named argument keys at the template layer._

### DIFF
--- a/tests/dummy/app/pods/docs/helpers/t/template.md
+++ b/tests/dummy/app/pods/docs/helpers/t/template.md
@@ -47,8 +47,6 @@ _Positional argument:_
 ```
 
 In the case where both named and positional arguments are used, they'll be merged together and named arguments will take precedence over the properties of a positional argument where there are duplicates.
-
-
 ## Format HTML Message
 
 To enable rendering HTML within translations, pass an `htmlSafe` attribute to the `t` helper.
--- a/tests/dummy/app/pods/docs/helpers/t/template.md
+++ b/tests/dummy/app/pods/docs/helpers/t/template.md
@@ -1,10 +1,10 @@
 {{locale-switcher}}
-# The `t` helper
+# The `t` Helper
 
 The `t` helper accepts a translation key and returns a translated string.
 To provide values to the dynamic segment of the translation, pass an object hash.
 
-## ICU message syntax
+## ICU Message Syntax
 
 Compiles a [ICU message syntax](https://formatjs.io/docs/core-concepts/icu-syntax) strings with its hash values passed.
 
--- a/tests/dummy/app/pods/docs/helpers/t/template.md
+++ b/tests/dummy/app/pods/docs/helpers/t/template.md
@@ -30,23 +30,22 @@ Compiles a [ICU message syntax](https://formatjs.io/docs/core-concepts/icu-synta
 }
 ```
 
-## Named VS Positional Arguments
+## Named and Positional Arguments
 
-Options may be passed to the `t` helper as either named or positional arguments.
-
-The following examples are equivalent:
+Options may be passed to the `t` helper as either named or positional arguments. The following examples are equivalent:
 
 _Named arguments:_
 ```hbs
 {{t 'photos.banner' numPhotos=count}}
 ```
 
-_Positional argument:_
+_Positional arguments:_
 ```hbs
 {{t 'photos.banner' (hash numPhotos=count)}}
 ```
 
-In the case where both named and positional arguments are used, they'll be merged together and named arguments will take precedence over the properties of a positional argument where there are duplicates.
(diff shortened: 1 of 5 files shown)


buschtoens (approved):
I love it. Thank you. ❤️

I think your PR description is already really great documentation. Would be awesome, if you could add that to the docs. 🤩

buschtoens:
You might e.g. add them here: https://github.com/ember-intl/ember-intl/blob/main/tests/dummy/app/pods/docs/helpers/t/template.md

jesdavpet:
Excellent, thanks for pointing out where to update the docs @buschtoens ! 🙌 

I've updated the [docs for the `t` helper](https://github.com/ember-intl/ember-intl/blob/e0967ac66824bc29050d18d0f4bebf529d69a658/tests/dummy/app/pods/docs/helpers/t/template.md#named-vs-positional-arguments) as of: https://github.com/ember-intl/ember-intl/pull/1633/commits/e0967ac66824bc29050d18d0f4bebf529d69a658

tests/dummy/app/pods/docs/helpers/t/template.md
```diff
@@ -30,6 +30,25 @@ Compiles a [ICU message syntax](https://formatjs.io/docs/core-concepts/icu-synta
 }
 ```
 
+## Named VS Positional Arguments
```
buschtoens, line 33:
```suggestion
## Named vs. Positional Arguments
```

I'm not a native English speaker, but AFAIK the short form of `versus` would still be spelled as `vs.` in title case. What do you think?

tests/dummy/app/pods/docs/helpers/t/template.md
```diff
@@ -30,6 +30,25 @@ Compiles a [ICU message syntax](https://formatjs.io/docs/core-concepts/icu-synta
...
+```
+
+_Positional argument:_
+```hbs
+{{t 'photos.banner' (hash numPhotos=count)}}
+```
+
+In the case where both named and positional arguments are used, they'll be merged together and named arguments will take precedence over the properties of a positional argument where there are duplicates.
+
+
```
buschtoens, line 51:
```suggestion

```

tests/dummy/app/pods/docs/helpers/t/template.md
jesdavpet: Good catch @buschtoens Buschtoens, actually I think neither is correct because this heading is title cased.

🤔 💭 Now that you bring it up, I'm considering this a bit more:
- using Latin terms isn't really a best practice in English technical writing, and
- also the word "versus" implies opposition, but named and positional arguments aren't opposed, or mutually exclusive in this case they can be used together

I will simplify this wording to avoid using "versus" entirely.

jesdavpet: 😅 Writing docs is hard, thanks for the feedback @buschtoens !!!

1. I simplified the heading and final sentence of this section in https://github.com/ember-intl/ember-intl/pull/1633/commits/cd089913af14f51991e84d08313c269bad3c101e

2. I noticed that a couple of the other headings in the docs for the `t` helper were not title cased, which was inconsistent with the style of the rest of the docs, so I fixed that in https://github.com/ember-intl/ember-intl/pull/1633/commits/97ade66d33ad2fa12dc96e6b502d90e2a012c72b

buschtoens (approved):
Thanks for this PR and being so helpful!

There was one formatting issue, which I fixed via 75e10ad.

buschtoens:
The failing CI checks (`ember-release`, `ember-beta`, `ember-canary`, TypeScript 3) are already failing on `main` and unrelated to this PR. Therefore I think we can go ahead with merging this.

Please note that there currently also is confusion about what version actually is the "right" version to use / publish as: #1631

I would suggest, that this change is published as `v6.0.0-beta.3` and tagged as `beta` for the time being.

buschtoens:
🚀 Published as https://github.com/ember-intl/ember-intl/releases/tag/v6.0.0-beta.3.

Issue #1560, opened 2021-05-06
This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'.


NullVoxPopuli:
- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 3.25.x
- Ember CLI Version: 3.25.x
- Ember Intl Version: 5.6.2
- Browser(s): n/a
- Node Version: n/a
- TypeScript: 4.2.4

- Using composite TS configs
  - not even sure which project in the monorepo is causing this, because this is the sole error

### Steps to Reproduce

Not sure yet, but here is what I'm getting:

```
 tsc --build --clean && tsc --build
node_modules/ember-intl/addon/-private/formatters/-base.ts:8:1 - error TS1371: This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'.

8 import { Formats } from '../../types';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.

```

NullVoxPopuli:
Here is the referenced code: https://github.com/ember-intl/ember-intl/blob/main/addon/-private/formatters/-base.ts#L7

and the error makes sense. The import should be `import type { Formats } ...`, but I don't yet know why it even matters since its in node_modules, and `importsNotUsedAsValues` is a project-level setting?

buschtoens:
I've just merged @hadiwina's PR #1635. Thanks for the contribution.

And thanks @NullVoxPopuli for getting my attention in the Discord.

Unfortunately `main` is already headed for `v6.x`. I'd like to backport the fix to `v5.x`, however there's currently a bit of confusion around the two latest versions `v5.7.1` / `v5.7.2`, as further discussed here: https://github.com/ember-intl/ember-intl/issues/1631#issuecomment-1017900383

I'm not sure right now, where to base the backport on. Once this is resolved, I'll cut another `v5.x` release.

hadiwina:
@buschtoens I found another potential similar issue [here](https://github.com/ember-intl/ember-intl/blob/a24c34f97837f4303abcc2a9e37e0652398120e7/addon-test-support/setup-intl.ts#L7). I think that this is also supposed to be `import type { Formats, Translations }`?

[...]

PR #1635, opened 2022-01-20
fix(types): fix `Formats` declaration as type

hadiwina:
This small change will fix the type error on #1560. `Formats` was previously not declared as `type` even though it is being used as a type declaration.


buschtoens (approved):

buschtoens:
Thank you!

hadiwina:
Thanks for the speedy review @buschtoens! 🙏

PR #1648, opened 2022-03-08
fix: `wrapTranslationsWithNamespace` not working on Windows

### LINKED ISSUE #1451: Nested folder structure doesn't seem to work on Windows

## Description

Hello. I tried to organize translations using nested folders. I can see the translations when I run the app locally on Mac, but another developer who works on Windows has been seeing "missing translation" error.

image

When I removed nested folder structure (i.e. have 1 file, `translations/en-us.yaml`), both of us could see the translations. I wondered if there may be an OS issue with joining directory paths that `ember-intl` isn't accounting for.

https://github.com/ember-intl/ember-intl/issues/159 seemed to be a similar past issue that arose due to OS differences.

## Environment

- Ember Version: 3.21.3
- Ember CLI Version: 3.18.0
- Ember Intl Version: 5.5.0
- Browser(s): Chrome (on Mac), Edge (on Windows)
- Node Version: 10.22.1

## Steps to Reproduce

I haven't tested this with the other developer, but believe the steps may be:

1. Create a new Ember app and install `ember-intl`.
1. Set `wrapTranslationsWithNamespace` to `true` in `config/ember-intl.js`.
1. Create a component: `ember g component parent/child`.
1. In the component's template, add a translation `{{t "components.parent.child.some-text"}}`.
1. Create the translation for `some-text` in `translations/components/parent/child/en-us.yaml` file.
1. Call the `<Parent::Child>` component in `app/templates/application.hbs`.

When I removed nested folder structure, I set `wrapTranslationsWithNamespace` back to `false`.

tfloxolodeiro:
Fixes #1451

The problem arises from the format of the `filepath`, `inputPath` and the `addonNames`, which are in the format `/tmp/broccoli_debug-output_path-l4iBcmcT/en-us.json`. The problem with this is that in Windows the `path.sep` is `\\`, and not `/`, like in POSIX. Because of this the paths are then incorrectly parsed.

To fix this I use the `normalize` method of the `path` module, that normalizes the path according to the OS it is running on. With this the previous path is being converted, in Windows, to `\\tmp\\broccoli_debug-output_path-l4iBcmcT\\en-us.json`, which is then correctly parsed.

### DIFF
--- a/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
+++ b/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
@@ -16,12 +16,16 @@ const enums = require('../../enums');
  * @private
  */
 function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
-  let dirname = path.dirname(filepath).replace(inputPath, '');
+  const normalizedFilePath = path.normalize(filepath);
+  const normalizedInputPath = path.normalize(inputPath);
+  const normalizedAddonNames = addonNames.map(path.normalize);
+
+  let dirname = path.dirname(normalizedFilePath).replace(normalizedInputPath, '');
 
   if (dirname) {
     let prefix = path.sep;
 
-    for (let addon of addonNames) {
+    for (let addon of normalizedAddonNames) {
       let addonPrefix = `${path.sep}${enums.addonNamespace}${path.sep}${addon}`;
 
       if (dirname.startsWith(addonPrefix)) {


ijlee2:
Hi, @tfloxolodeiro. Can you rebase your branch so that I can try running CI?

```sh
git remote add upstream git@github.com:ember-intl/ember-intl.git

git checkout normalizing-wrap-paths
git pull --rebase upstream main

git push -f origin normalizing-wrap-paths
```

I tried the commands above but couldn't force-push. (You may need to allow an admin to edit your pull request.)

ijlee2:
Closed this PR in favor of #1725

PR #1660, opened 2022-04-13
Replace `intl-messageformat-parser` requires

charlesfries:
Fixes oversights from #1653

There is also an incorrect import in `addon/-private/utils/parse.ts` but TypeScript is complaining about the `normalizeHashtagInPlural` property not existing in the new package. Not entirely sure what that property does. `intl-messageformat-parser` documentation says this:

```js
/**
 * Whether to convert `#` in plural rule options
 * to `{var, number}`
 * Default is true
 */
normalizeHashtagInPlural?: boolean;
```

https://github.com/ember-intl/ember-intl/blob/main/addon/-private/utils/parse.ts#L10


runspired:
opened https://github.com/formatjs/formatjs/issues/3563 for discussion

PR #1674, opened 2022-06-09
Updated test-related packages to their latest version. Imported TestContext from @ember/test-helpers.

### LINKED ISSUE #1640: TestContext should be imported from @ember/test-helpers

### Environment

- Ember Version: 3.15.0
- Ember CLI Version: 3.15.0
- Ember Intl Version: 5.7.2
- Browser(s): -
- Node Version: 12.21.0

### Steps to Reproduce

`TestContext` should be imported from `@ember/test-helpers` instead of the older `ember-test-helpers`. Otherwise we get this error:

image

As per official documentation:

> Breaking: The TestContext for the callbacks used in module, beforeEach, afterEach, test, etc. has moved from ember-test-helpers to @ember/test-helpers. Having nearly everything be in @ember/test-helpers but not TestContext has been a persistent source of confusion over the years, and was just a bit of historical baggage left over from before @ember/test-helpers existed. 

Source: https://blog.emberjs.com/ember-v4-typescript-support-update/

ijlee2:
## Background

I was working on upgrading `@types/ember-qunit` for a production project. After upgrading `@types/ember-qunit` to `v5.0.0`, I ran into a TypeScript error because `ember-intl` had imported `TestContext` from `ember-test-helpers` (old syntax) instead of `@ember/test-helpers` (new).

```sh
$ tsc --noEmit
../node_modules/ember-intl/test-support/setup-intl.d.ts:2:53 - error TS2307: Cannot find module 'ember-test-helpers' or its corresponding type declarations.

import type { TestContext as BaseTestContext } from 'ember-test-helpers';
```

I used [`patch-package`](https://www.npmjs.com/package/patch-package) to temporarily patch `ember-intl`.

## What the pull request does 

In addition to keeping the development dependencies up-to-date, we can close #1640 and avoid patching the addon in the consuming project.

## References

- [Migration guide: Upgrading `ember-qunit` from `v4` to `v5`](https://github.com/emberjs/ember-qunit/blob/v5.1.5/docs/migration.md)
- [Pinning `engines.node` to `v14`](https://github.com/vercel/vercel/pull/7782)


package.json
```diff
@@ -91,7 +92,6 @@
     "ember-cli-deploy-git": "^1.3.4",
     "ember-cli-htmlbars": "^5.3.2",
     "ember-cli-inject-live-reload": "^2.0.2",
-    "ember-cli-qunit": "^4.4.0",
```
ijlee2, line 94: [`ember-cli-qunit`](https://github.com/ember-cli/ember-cli-qunit) has been deprecated. I replaced it with [`ember-qunit`](https://github.com/emberjs/ember-qunit).

package.json
```diff
@@ -127,7 +129,7 @@
     }
   },
   "engines": {
-    "node": ">= 10.*"
+    "node": ">= 14.*"
```
ijlee2, line 132: When `ember-intl@v6.0.0` is released, we may want to denote that Node 10 and 12 supports are dropped?

ijlee2, line 132: Running `ember s` or `ember t` resulted in an error (this happened even in the `main` branch at the time of writing):

```sh
building... [Babel: ember-test-helpers > applyPatches]node:internal/validators:164
      throw new ERR_INVALID_ARG_TYPE(name, 'Object', value);
      ^

TypeError [ERR_INVALID_ARG_TYPE]: The "options" argument must be of type object. Received an instance of Array
    at ChildProcess.target.send (node:internal/child_process:733:7)
    at Array.forEach (<anonymous>)
    at dispatchQueuedRequests (/ember-intl/node_modules/@ember/test-helpers/node_modules/workerpool/lib/WorkerHandler.js:174:21)
    at ChildProcess.<anonymous> (/ember-intl/node_modules/@ember/test-helpers/node_modules/workerpool/lib/WorkerHandler.js:129:7)
    at ChildProcess.emit (node:events:527:28)
    at ChildProcess.emit (node:domain:475:12)
    at emit (node:internal/child_process:938:14)
    at processTicksAndRejections (node:internal/process/task_queues:84:21) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

According to https://github.com/vercel/vercel/pull/7782, we'll want to pin Node to `14`.

addon-test-support/setup-intl.ts
```diff
@@ -1,7 +1,7 @@
 import addTranslations from './add-translations';
 import { missingMessage } from './-private/serialize-translation';
-import type { TestContext as BaseTestContext } from 'ember-test-helpers';
 import { settled } from '@ember/test-helpers';
+import type { TestContext as BaseTestContext } from '@ember/test-helpers';
```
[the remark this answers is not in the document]
buschtoens: Can you explain why?

Could we add `@ember/test-helpers` as a peer dependency instead?

package.json
buschtoens: I think we should say so in the release notes, but it wouldn't necessarily fail the install, unless [`engine-strict`](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict) is set.

Given that `10` and `12` have long reached their EOL, I think it's fine to drop them.

addon-test-support/setup-intl.ts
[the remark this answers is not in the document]
ijlee2: I think I misunderstood the differences between `ember-test-helpers` and `@ember/test-helpers` (if there are any). If so, we probably don't need to specify `@ember/test-helpers` as a peer dependency.

Initially, I was thinking, depending on the following 2 factors:

- Is `ember-test-helpers` or `@ember/test-helpers` installed for the user's app
- Which version of `ember-intl` is installed? One that imports `TestContext` from `ember-test-helpers`, or one that imports `TestContext` from `@ember/test-helpers`?
 
the user might still run into the error `Cannot find module '@ember/test-helpers' or its corresponding type declarations.`, just like I had run into the error `Cannot find module 'ember-test-helpers' or its corresponding type declarations.`.

ijlee2:
@buschtoens Thanks for quickly getting to reviewing this pull request.

If you run the CI workflow again, I think lint and test will pass now. I feel bad that the pull request contains unrelated commits. If you prefer, I can create a separate pull request for updating Node and for updating files in the `lib` and `tests-node` folder.

lib/broccoli/translation-reducer/linter/extract-icu-arguments.js
```diff
@@ -1,4 +1,4 @@
-const parser = require('intl-messageformat-parser');
+const { parse, TYPE } = require('@formatjs/icu-messageformat-parser');
```
ijlee2, line 1: There seems to have been a syntax change.

https://formatjs.io/docs/icu-messageformat-parser/#usage

PR #1687, opened 2022-06-12
Allow to override format* methods on service, fixes #1684

### LINKED ISSUE #1684: Allow to override format methods in service

### Steps to Reproduce

1. Add own service `app/services/intl.js`
2. Subclass service from `ember-intl`
3. Override any of the `format*` methods

```javascript
import EmberIntl from 'ember-intl/services/intl';

export default class IntlService extends EmberIntl {

  formatDate(date, options) {
    // do something custom
    return super.formatDate(...arguments)
  }
}
```

4. `service.formatDate()` calls original [method-as-property](https://github.com/ember-intl/ember-intl/blob/main/addon/services/intl.js#L83) instead of the custom method 💥

I'm not sure if this is intentional or `format*` methods should be defined as prototype methods on class to allow/fix standard/expected behaviour?

bobisjan:


ijlee2:
Hi. I'll close the PR, as the project has changed since June 2022 and I'm currently not looking to add features.

PR #1694, opened 2022-06-16
Fixed continuous deployment

ijlee2:
## Description

By trial and error, I discovered two issues that would prevent correctly deploying the app.

1. We need to pass the environment variable `ADDON_DOCS_UPDATE_LATEST=true` to `ember deploy production` if we want code changes to the demo application to take place. As a result, I think it makes more sense to pass the variable when a pull request is merged to `main`, rather than [when a tag is released](https://github.com/ember-intl/ember-intl/pull/1425).

2. `ember-cli-addon-docs` used to depend on `ember-get-config@v1.0.4`, which causes a runtime error in a brand-new, modern Ember app (`ember-source@v4.4.1`, `ember-auto-import@v2`).

    

    Two pull requests that remove the dependency have been merged:

    - https://github.com/ember-learn/ember-cli-addon-docs/pull/1137
    - https://github.com/ember-learn/ember-cli-addon-docs/pull/1143

    However, the addon hasn't released a new version that would allow these pull requests to take effect. As a temporary measure, I pinned `ember-get-config` to `v0.5.0` (in order to revert the pull request `#1084` in https://github.com/ember-learn/ember-cli-addon-docs/releases/tag/v4.2.2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,6 +132,9 @@
       "optional": true
     }
   },
+  "resolutions": {
+    "ember-get-config": "0.5.0"
+  },
   "engines": {
     "node": "14.* || >= 16"
   },
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -146,7 +146,7 @@ jobs:
     needs: [lint, test-addon, test-floating-dependencies, try-scenarios]
     timeout-minutes: 10
     # Only run on pushes to the main branch or a tag (i.e. ignore pull requests and cron)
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -132,6 +132,9 @@
       "optional": true
     }
   },
+  "resolutions": {
+    "ember-get-config": "0.5.0"
+  },
```
ijlee2, line 137: Please remove this resolution after `ember-cli-addon-docs` makes a new release, in which `ember-get-config` is no longer a dependency.

PR #1696, opened 2022-06-17
set importsNotUsedAsValues

NullVoxPopuli:
in typescript projects, whenever typescript decides it needs to verify types of libraries, if `importsNotUsedAsValues` is set in the parent project, ember-intl will error because it has had inconsistent imports.

Using type imports helps communicate which imports are removed at build time.


ijlee2 (approved):
Looks great, thank you!

PR #1682, opened 2022-06-09
Updated ember-source to v4.4.1

ijlee2:
## Description

Updating the demo app to run the latest Ember isn't trivial, because its dependency `ember-cli-addon-docs` is outdated. We can see from #1634 that an attempt to update everything at once may lead to languished development. (The pull request did thankfully provide me a few ideas.)

In this pull request, we'll update `ember-source` to the latest (`v4.4.1`), then delete the demo application so that we can easily start over. In subsequent pull requests, I will reintroduce the documentations using the latest version of `ember-cli-addon-docs`.

PR #1697, opened 2022-06-22
Enabled eslint-plugin-qunit

ijlee2:
## Description

In #1682, `eslint-plugin-qunit` had been installed as a part of the upgrade to Ember 4.4. I had to disable the addon, however, because many tests didn't follow a good standard.

In this pull request, I focused on replacing the assertion methods (e.g. from `assert.equal` to `assert.strictEqual`) so that the tests meet the standard requirements of `eslint-plugin-qunit`.


testem.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  test_page: 'tests/index.html?hidepassed',
+  test_page: 'tests/index.html?hidepassed&nolint',
```
ijlee2, line 2: I preemptively added the `nolint` flag so that `yarn test` focuses on testing files (and `yarn lint` on linting files).

The `nolint` flag isn't needed for `eslint`-related packages. It is needed, however, should we later install `ember-cli-dependency-lint`.

PR #1706, opened 2022-07-13
Add new test-helper: getIntlService

NullVoxPopuli:
the current way to use the intl service in tests is to use `this.intl` on the test context via `setupIntl` -- this provides a nice convenience, but can be intrusive in typecsript codebases as it requires that the `this` type for each test use the custom `TestContext` provided by ember-intl.

Because modifying `this` in typecsript can be a bit annoying (and bug/error prone), I've found it best to avoid using `this` altogether in typescript tests -- which happens to align nicely with where the framework is going with Polaris, and `<template>` syntax in tests (with things being able to be directly accessed via scope, rather than proxied through `this`).

This PR adds a new test-helper, `getIntlService`, which returns the `IntlService` with the correct type.

```ts
   import { setupIntl, getIntlService } from 'ember-intl/test-support';
  
   module('My Module', function (hooks) {
     setupRenderingTest(hooks);
     setupIntl(hooks);
  
     test('my test', async function (assert) {
       let intlService = getIntlService();
  
       intlService.t('my-translation-key')
     });
   });
``` 
This allows folks to have a correctly typed version of the intl service as well as avoid `this` :partying_face:

### DIFF
--- a/addon-test-support/get-intl-service.ts
+++ b/addon-test-support/get-intl-service.ts
@@ -1,0 +1,46 @@
+import type EngineInstance from '@ember/engine/instance';
+import { getContext, TestContext } from '@ember/test-helpers';
+import type IntlService from 'ember-intl/services/intl';
+
+/**
+ * Helper that provides the instance of the intl service for use during tests.
+ *
+ * @example
+ * ```js
+ * import { module, test } from 'qunit';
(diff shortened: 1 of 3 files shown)


ijlee2:
Hi. I'll close the PR, as the project has changed since July 2022 and I'm currently not looking to add features.

PR #1715, opened 2022-08-17
fix(types): allow `exactOptionalPropertyTypes`

jamescdavis:
v5.x is not currently compatible with apps that have [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) enabled. This turns that option on here and updates the types of `RelativeTimeFormatOptions` to be compatible.

The reason why this is required is that the `& keyof KnownOptions` restriction here:
https://github.com/ember-intl/ember-intl/blob/daad6f592fdd7e52f01df058e43c0a4f847149f6/addon/-private/formatters/-base.ts#L61
forces the return type to include `| undefined` for each option key.

@ijlee2 @buschtoens

### DIFF
--- a/addon/-private/formatters/-base.ts
+++ b/addon/-private/formatters/-base.ts
@@ -47,7 +47,7 @@ export default abstract class FormatterBase<KnownOptions extends {}> {
   }
 
   get options(): readonly (keyof KnownOptions)[] {
-    return ([] as unknown[]) as readonly (keyof KnownOptions)[];
+    return [] as unknown[] as readonly (keyof KnownOptions)[];
   }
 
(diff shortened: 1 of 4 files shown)


buschtoens (approved):
Thank you!

buschtoens:
The changes themself are fine, but our CI setup is broken. I'll merge and try to allocate some time to fixing it.

jamescdavis:
LMK when you are able to release this. 🙂 We're trying to set up ember-intl on several apps at @salsify that have `exactOptionalPropertyTypes` enabled.

PR #1719, opened 2022-09-08
Add Glint types

jamescdavis:
This adds Glint types for all ember-intl helpers and sets up Glint to run in CI to check invocations in tests.

The `format-date` unit tests include a couple of invocations with a value of `null`, which is invalid according to the types. I've marked those with `@glint-expect-error` for now.

CC: @dfreeman @chriskrycho

### DIFF
--- a/addon/helpers/-format-base.d.ts
+++ b/addon/helpers/-format-base.d.ts
@@ -4,7 +4,7 @@ import IntlService from '../services/intl';
 
 export interface BaseHelperSignature<Value, Options extends Record<string, unknown>> {
   Args: {
-    Positional: [Value] | [Value, Options] | [];
+    Positional: [Value?, Options?];
     Named?: Options & { allowEmpty?: boolean };
   };
   Return: string;
--- a/addon/helpers/-format-base.d.ts
+++ b/addon/helpers/-format-base.d.ts
@@ -2,14 +2,21 @@ import Helper from '@ember/component/helper';
 
 import IntlService from '../services/intl';
 
-// eslint-disable-next-line @typescript-eslint/ban-types
-export default abstract class AbstractHelper<V, O extends {} | undefined> extends Helper {
+export interface BaseHelperSignature<Value, Options extends Record<string, unknown>> {
+  Args: {
+    Positional: [Value] | [Value, Options] | [];
+    Named?: Options & { allowEmpty?: boolean };
+  };
+  Return: string;
+}
+
+export default abstract class AbstractHelper<Signature extends BaseHelperSignature> extends Helper<Signature> {
   readonly intl: IntlService;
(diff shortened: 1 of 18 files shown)


jamescdavis:
Oops, my `@glint-expect-error` directives messed up tests. I'll fix.

addon/helpers/-format-base.d.ts
```diff
@@ -2,14 +2,21 @@ import Helper from '@ember/component/helper';
...
 
-// eslint-disable-next-line @typescript-eslint/ban-types
-export default abstract class AbstractHelper<V, O extends {} | undefined> extends Helper {
+export interface BaseHelperSignature<Value, Options extends Record<string, unknown>> {
+  Args: {
+    Positional: [Value] | [Value, Options] | [];
+    Named?: Options & { allowEmpty?: boolean };
+  };
+  Return: string;
+}
```
ijlee2, line 11: Could someone review these lines of code in particular?

cc: @buschtoens

ijlee2 (approved):
I checked all edited files except `addon/helpers/-format-base.d.ts`. Looks good to me.

addon/helpers/-format-base.d.ts
```diff
@@ -2,14 +2,21 @@ import Helper from '@ember/component/helper';
 
 import IntlService from '../services/intl';
 
-// eslint-disable-next-line @typescript-eslint/ban-types
-export default abstract class AbstractHelper<V, O extends {} | undefined> extends Helper {
+export interface BaseHelperSignature<Value, Options extends Record<string, unknown>> {
+  Args: {
+    Positional: [Value] | [Value, Options] | [];
```
dfreeman, line 7: Would `[Value?, Options?]` work here?

tests/unit/helpers/format-date-test.ts
```diff
@@ -26,7 +26,9 @@ module('format-date', function (hooks) {
 
   test('should render empty string for a null value', async function (assert) {
     assert.expect(1);
-    await render(hbs`{{format-date null}}`);
+    await render(hbs`
+      {{~! @glint-expect-error ~}}
+      {{format-date null}}`);
```
dfreeman, line 31: If there are tests that explicitly verify the behavior of `format-date` with a `null` input, should the types allow for that?

jamescdavis: The underlying library's (`@formatjs/intl`) types don't allow `null`.

addon/helpers/-format-base.d.ts
jamescdavis: Yeah, should be equivalent!

ijlee2 (approved):

PR #1727, opened 2022-10-18
test: backport await settled in test setupIntl() to fix race condition with fake timers

achambers:
Apply changes from https://github.com/ember-intl/ember-intl/pull/1616 to v5.x branch


achambers:
Wondering what the steps are to get the logic from #1616 backported in to the `5.x` series. This PR attempts to do that.

Any steer on whether I'm going about this the correct way?

/cc @longlho @jasonmit @Turbo87 @buschtoens

achambers:
We're going to use patch-package to get us this functionality until we upgrade to v6.x

Turbo87:
you could also use a custom wrapper around the original instead ;)

PR #1734, opened 2022-11-18
build: add volta configuration for node and yarn versions

ascudder:
This is a small addition to make it easier to contribute to this project.

Since I use volta to manage the versions of node and the package managers that I use, I thought it would help in this project as well.


ijlee2:
@ascudder Thank you for the suggestion. At the moment, I prefer not to add Volta, and will address updating the versions of Node and package manager.

PR #1748, opened 2023-04-09
fix: Filepath matching is overly-eager when namespace wrapping

prsethre:
When the `wrapWithNamespaceIfNeeded` function runs in a project that has two or more addons sharing an initial substring (e.g addons named `foo` and `foo-bar`), the filepath to addon prefix matching can match an addon to the incorrect filepath. This is because the final separator is not used in the match, resulting in `'/path/to/addon/foo-bar'.startsWith('path/to/addon/foo')` evaluating to `true`, which in turn results in an inappropriately wrapped namespace of `-bar`, causing translation path lookups to fail.

The solution is to take the final path separator into account for the filepath matching, so that `'/path/to/addon/foo-bar/'.startsWith('path/to/addon/foo'/)` correctly evaluates to `false`.

### DIFF
--- a/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
+++ b/lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
@@ -28,7 +28,7 @@ function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
     for (let addon of normalizedAddonNames) {
       let addonPrefix = `${path.sep}${enums.addonNamespace}${path.sep}${addon}`;
 
-      if (dirname.startsWith(addonPrefix)) {
+      if (`${dirname}/`.startsWith(`${addonPrefix}/`)) {
         prefix = addonPrefix;
         break;
       }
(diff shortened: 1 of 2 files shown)


ijlee2 (approved):
Thank you for describing the issue as well as providing a fix and test.

lib/broccoli/translation-reducer/utils/wrap-with-namespace-if-needed.js
```diff
@@ -28,7 +28,7 @@ function wrapWithNamespaceIfNeeded(object, filepath, inputPath, addonNames) {
     for (let addon of normalizedAddonNames) {
       let addonPrefix = `${path.sep}${enums.addonNamespace}${path.sep}${addon}`;
 
-      if (dirname.startsWith(addonPrefix)) {
+      if (`${dirname}/`.startsWith(`${addonPrefix}/`)) {
```
ijlee2, line 31: When I tried to set up CI to run the Node tests on Windows, I found that the `/` causes the tests to fail. I'll create a pull request to replace the hardcoded character with `path.sep` (line 29 shows an example).

PR #1756, opened 2023-06-30
Checked compatibilities with Ember 4.4, 4.8, and 4.12

ijlee2:
## Why?

As seen in [the run for #1755](https://github.com/ember-intl/ember-intl/actions/runs/5427208700),

<div align="center">
    
</div>

the `3.24` and `3.28` scenarios are passing (though both LTS's are no longer supported by Ember), but the `release` scenario (`5.1.2`) is failing. We don't know what happens in-between.

## Solution?

I update the CI workflow, then added `4.4`, `4.8`, and `4.12` to the `ember-try` scenarios. I'll drop `3.24` support after releasing `6.0.0-beta.5` (i.e. target `6.0.0-beta.6`).

<div align="center">
    
</div>

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "deploy": "ember deploy production",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
     "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
+    "lint:glint": "glint",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "deploy": "ember deploy production",
     "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
     "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
+    "lint:glint": "glint",
```
ijlee2, line 30: **TODO**

Update `tsconfig.json` so that `lint:types` runs `glint`. (We may need a separate script to type-check Node files.)

PR #1759, opened 2023-07-01
Addressed failing CI

ijlee2:
## Why?

Since March 26, the floating dependencies check has failed because the project depends on Node 14.

Recently, the `ember-release`, `ember-beta`, and `ember-canary` began to fail as well, because these scenarios point to `ember-data@v5`, a dependency of `ember-cli-addon-docs`.

## Solution?

I allowed temporarily ignoring these failures and removed the floating dependencies check as a required GitHub status check. The reason is, these failures aren't directly caused by `ember-intl` code (as shown by the passing `ember-lts-3.24`, ..., `ember-lts-4.12` scenarios).

Soon, we'll want to:

- Drop Node 14 support
- Introduce pnpm workspace and have separate docs and test apps

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -137,7 +137,7 @@ jobs:
   deploy-documentation:
     name: Deploy documentation
     runs-on: ubuntu-latest
-    needs: [lint, test, test-compatibility, test-floating-dependencies, test-node]
+    needs: [lint, test, test-compatibility, test-node]
     timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -137,7 +137,7 @@ jobs:
   deploy-documentation:
     name: Deploy documentation
     runs-on: ubuntu-latest
-    needs: [lint, test, test-compatibility, test-floating-dependencies, test-node]
+    needs: [lint, test, test-compatibility, test-node]
```
ijlee2, line 140: The changes in this pull request will help us continuously deploy the documentation site again.

PR #1767, opened 2023-07-01
Checked the Node tests on Windows in CI

ijlee2:
## Why?

Retried #1757. Let's check if using Node 14 was the issue that had caused `yarn install` to stall.

Current findings:

- Even when Node 16 is used, running `yarn install` stalls on Windows in CI. It may be that the project dependencies are listed correctly?


ijlee2:
Closed in favor of #1765.

Issue #1774, opened 2023-07-21
bump typescript peer dependency to 5


Techn1x:
I've noticed a bunch of updates! I've updated to ember-intl 6 beta 6, seems to be working great. I did have to add `setupIntl(hooks, 'en')` to a _lot_ of my tests though, but that's OK and probably a good thing to be explicit. I'm also using the glint types from this addon too.

Just hoping that as part of these updates we can bump the typescript peer dependency? 4 is quite old, been on 5 for ages, and pnpm keeps complaining at me :)

![image](https://github.com/ember-intl/ember-intl/assets/1049837/8ff91fc5-1f0d-4484-9716-bf85ef3d82c0)

Thanks for the hard work! So glad to see this addon getting some TLC again. Hoping to see it in a stable place

---

- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 4.12.3
- Ember CLI Version: 5
- Ember Intl Version: 6 beta 6 (latest)
- Browser(s): chrome
- Node Version: 18

ijlee2:
@Techn1x Sorry for a late reply. I appreciate your trying out the latest beta version and providing feedback. I imagine there may be projects that are still on `typescript@v4`.

While there are some differences in TS configuration between `v4` and `v5`, I think we could try out supporting both versions in the following way:

```json
/* package.json */
"peerDependencies": {
  "typescript": "^4.8.0 || ^5.0.0"
},
```

What do you think? If the idea above sounds ok, can you make a pull request? We'll then make another beta release for people to try out.

Techn1x:
Expanding the peer dep range sounds great to me. Will try and find some time to put a PR together

Techn1x:
done https://github.com/ember-intl/ember-intl/pull/1776

PR #1776, opened 2023-08-29
Limited TypeScript v4 support to 4.8 and 4.9. Expanded TS support to 5.x.

Techn1x:
## Why?

Resolves https://github.com/ember-intl/ember-intl/issues/1774
So that consumers can use TS 5 without warnings

## Solution?

* expands (optional) peerDep range to TS 5
* uses TS 5 in the project
* fixes any issues that came up
  * replaced deprecated tsconfig option
  * provided TestContext to render calls that use context ie `await render<TestContext>(...)`
* replaced `@ts-ignore` directives with `@ts-expect-error`, the latter is better as it will error if the type error is fixed - making sure the dev removes the directive when able to do so. It also doesn't require ignoring an eslint rule.

### DIFF
--- a/tests/integration/helpers/format-message-test.ts
+++ b/tests/integration/helpers/format-message-test.ts
@@ -1,10 +1,20 @@
 import { findAll, render, rerender } from '@ember/test-helpers';
 import { tracked } from '@glimmer/tracking';
 import { hbs } from 'ember-cli-htmlbars';
-import type { TestContext } from 'ember-intl/test-support';
+import type { TestContext as BaseTestContext } from 'ember-intl/test-support';
 import { setupIntl } from 'ember-intl/test-support';
 import { setupRenderingTest } from 'ember-qunit';
 import { module, test } from 'qunit';
+
+interface TestContext extends BaseTestContext {
+  census_date?: number;
+  city?: string;
+  harvests?: { count: number; person: string }[];
+  instance?: SomeClass;
+  population?: number;
+  timeZone?: string;
+  translation?: string;
+}
 
 class SomeClass {
   @tracked someCondition = true;
@@ -85,7 +95,7 @@ module('Integration | Helper | format-message', function (hooks) {
         '{city} has a population of {population, number, integer} as of {census_date, date, long}.',
     });
 
-    await render(hbs`
+    await render<TestContext>(hbs`
       {{format-message
         this.translation
         city=this.city
@@ -112,7 +122,7 @@ module('Integration | Helper | format-message', function (hooks) {
         '{city} hat eine Bevölkerung von {population, number, integer} zum {census_date, date, long}.',
     });
 
-    await render(hbs`
+    await render<TestContext>(hbs`
       {{format-message
         this.translation
         city=this.city
@@ -139,7 +149,7 @@ module('Integration | Helper | format-message', function (hooks) {
         '{person} harvested {count, plural, one {# apple} other {# apples}}.',
     });
 
-    await render(hbs`
+    await render<TestContext>(hbs`
       {{#each this.harvests as |harvest|}}
         <div data-test-output>
           {{format-message this.translation person=harvest.person count=harvest.count}}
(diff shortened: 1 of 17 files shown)


tests/integration/helpers/format-message-test.ts
```diff
@@ -1,11 +1,21 @@
...
 
+interface TestContext extends BaseTestContext {
+  census_date?: number;
+  city?: string;
+  harvests?: { count: number; person: string }[];
+  instance?: SomeClass;
+  population?: number;
+  timeZone?: string;
+  translation?: string;
+}
```
Techn1x, line 17: this test was missing a TestContext interface with the values used by its tests

tests/integration/helpers/format-time-test.ts
```diff
@@ -1,17 +1,14 @@
-import {
-  render,
-  type TestContext as BaseTestContext,
-} from '@ember/test-helpers';
+import { render } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
 import type IntlService from 'ember-intl/services/intl';
+import type { TestContext as BaseTestContext } from 'ember-intl/test-support';
```
Techn1x, line 4: The TestContext provided by this addon is the standard test context interface, along with `intl: IntlService` added.

tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "alwaysStrict": true,
     "strictNullChecks": true,
     "strictPropertyInitialization": true,
-    "importsNotUsedAsValues": "error", // enforces type imports when imports are not used as values
+    "verbatimModuleSyntax": true, // enforces type imports when imports are not used as values
```
Techn1x, line 13: This was the deprecated tsconfig option
See https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues
and https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax

tests/unit/formatters/format-relative-test.ts
```diff
@@ -1,4 +1,4 @@
-import { createIntl, OnErrorFn } from '@formatjs/intl';
+import { createIntl, type OnErrorFn } from '@formatjs/intl';
```
Techn1x, line 1: fixing the deprecated config option brought this up - it's only used as a type, so we should import it as such.

ijlee2 (commented):
Much thanks for updating `typescript@5.2.2` and fixing TS errors in test files. While trying to update `@glint/*` packages in #1766, I was blocked by having to update `typescript` at the same time, so I'm glad to see that the work for the latter is now separated.

I noticed that the command for testing the `embroider-optimized` scenario now hangs ([in CI](https://github.com/ember-intl/ember-intl/actions/runs/6007468267/job/16296365218) and locally). Before it would finish in a couple of minutes (the scenario fails, but we had configured `ember-try` to allow tests to fail).

```sh
# Command (for local development)
yarn test:ember-compatibility embroider-optimized --- yarn test

# Output (we can ignore the warning messages from `testdouble`)
WARNING in ./tests/test-helper.ts 11:30-32
export 'default' (imported as 'td') was not found in 'testdouble' (possible exports unknown)
 @ ./assets/test.js 40:13-50 54:4-35

WARNING in ./tests/unit/services/intl-test.ts 14:26-33
export 'default' (imported as 'td') was not found in 'testdouble' (possible exports unknown)
 @ ./assets/test.js 52:13-62

webpack 5.88.2 compiled with 2 warnings in 21987 ms
cleaning up
cleaning up...
Built project successfully. Stored in "/tmp/tests-dist-2023729-1889-1qo2m01.seoa".

# Expected: After the message above, 2 tests should run and fail.
```

Do you have ideas about how updating `typescript` to `5.2.2` could have affected the run? If not, I could propose that we don't check `embroider-optimized` scenario for now, as a passing check in CI doesn't actually mean that `ember-intl` is optimized-ready.

Techn1x:
That's weird, I'm not sure sorry!

Techn1x:
I gave it a go locally, unfortunately neither embroider scenario runs for me on my machine, even for main branch. so I'm a bit useless in the debugging here :/

Screenshot 2023-08-30 at 2 09 03 pm

ijlee2:
@Techn1x No worries, I'll try to find time today to look more into the problem.

ijlee2 (approved):

Techn1x:
Thanks a bunch for looking into it

PR #1777, opened 2023-08-30
Temporarily skipped embroider-safe and embroider-optimized scenarios

ijlee2:
## Why?

Since CI was last run on July 1, something from the dependencies seems to have changed. The `embroider-optimized` scenario (as well as `embroider-safe`) hangs indefinitely, both in CI for #1776 as well as locally. We can see this problem occur from the CI run for this pull request's 1st commit (note, no code was changed).

## Solution?

In #1749, we had added the two scenarios to CI but configured `ember-try` so that failing tests would still result in a passing CI. I think an incorrectly passing CI doesn't provide us (maintainers as well as end-developers) useful information.

To unblock #1776, I will temporarily skip running `embroider-safe` and `embroider-optimized` scenarios in CI.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -71,8 +71,8 @@ jobs:
           - 'ember-beta'
           - 'ember-canary'
           - 'ember-classic'
-          - 'embroider-safe'
-          - 'embroider-optimized'
+          # - 'embroider-safe'
+          # - 'embroider-optimized'
     timeout-minutes: 5
     steps:
       - name: Check out a copy of the repo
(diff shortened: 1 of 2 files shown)


ynotdraw:
Hi @ijlee2 - any clues as to a path forward for this?  I may have some spare cycles to contribute if you have some idea as to how `ember-intl` could become embroider compatible/safe.

ijlee2:
Hi, @ynotdraw. I appreciate your offer to spend time to help out.

I'm not sure yet how `ember-intl` can become a v2 addon, as I still need to learn the ins and outs of this package. I do want to simplify code and remove unnnecessary dependencies, which may help with the v2 migration.

Perhaps a refactor task that you could start with is, to review and reorganize tests on the Ember side. I feel like, over time, tests had been added to an existing file in `integration` or `unit`, without someone maintaining the test code.

Do you have additional ideas for what you'd like to work on?

PR #1786, opened 2023-09-20
Remove unused and unnecessary development dependencies

ijlee2:
## Why?

To lower the maintenance cost and make it clear which dependencies are needed by the docs app.

### DIFF
--- a/tests/dummy/app/helpers/root-url.ts
+++ b/tests/dummy/app/helpers/root-url.ts
@@ -1,0 +1,16 @@
+import { getOwner } from '@ember/application';
+import Helper from '@ember/component/helper';
+
+export default class RootUrl extends Helper {
+  build(relativeURL: string): string {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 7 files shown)


tests/dummy/app/helpers/root-url.ts
```diff
@@ -0,0 +1,16 @@
...
+    // @ts-ignore Property `resolveRegistration` does not exist on type `Owner`
+    const ENV = getOwner(this).resolveRegistration('config:environment');
+
+    return `${ENV.rootURL}${relativeURL.replace(/^\//, '')}`;
+  }
+
+  compute([relativeURL]: [string]) {
+    return this.build(relativeURL);
+  }
+}
```
ijlee2, line 16: Adapted from https://github.com/ef4/ember-root-url/blob/v1.0.1/addon/services/root-url.js

ijlee2: We continue to see the screenshots for the compatibility matrix.

PR #1791, opened 2023-11-04
Pinned @babel/core, @babel/traverse, and fast-glob to address vulnerabilities

ijlee2:
## Why?

Dependabot raised a couple of vulnerability reports, caused by the dependencies of `ember-intl`'s dependencies.

- https://github.com/ember-intl/ember-intl/security/dependabot/93
- https://github.com/ember-intl/ember-intl/security/dependabot/80

## Solution?

Ideally, `ember-intl` would depend on fewer dependencies so that the chance of a vulnerability is low, and the remaining dependencies update their dependencies regularly.

Unfortunately, this is not the case, especially when it comes to broccoli packages. So instead, let's try fixing the two vulnerabilities by forcing a specific version to be installed. How this may cause a regression, not caught by tests, is unknown.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,6 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
+  '@babel/core': ^7.23.2
   '@types/ember': ^4.0.8
   '@types/ember__application': ^4.0.9
   '@types/ember__array': ^4.0.7
@@ -20,6 +21,7 @@ overrides:
   '@types/ember__utils': ^4.0.5
   '@types/htmlbars-inline-precompile': ^3.0.1
   '@types/qunit': ^2.19.7
+  fast-glob: ^3.3.1
 
 dependencies:
   '@formatjs/icu-messageformat-parser':
@@ -57,7 +59,7 @@ dependencies:
     version: 2.6.3(webpack@5.89.0)
   ember-cli-babel:
     specifier: ^8.2.0
-    version: 8.2.0(@babel/core@7.22.20)
+    version: 8.2.0(@babel/core@7.23.2)
   ember-cli-typescript:
     specifier: ^5.2.1
     version: 5.2.1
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1470,21 +1501,21 @@ packages:
...
       '@babel/code-frame': 7.22.13
-      '@babel/parser': 7.22.16
-      '@babel/types': 7.22.19
+      '@babel/parser': 7.23.0
+      '@babel/types': 7.23.0
 
-  /@babel/traverse@7.22.20:
-    resolution: {integrity: sha512-eU260mPZbU7mZ0N+X10pxXhQFMGTeLb9eFS0mxehS8HZp9o1uSnFeWQuG1UPrlxgA7QoUzFhOnilHDp0AXCyHw==}
+  /@babel/traverse@7.23.2:
+    resolution: {integrity: sha512-azpe59SQ48qG6nu2CzcMLbxUudtN+dOM9kDbUqGq3HXUJRlo7i8fvPoxQUzYgLZ4cMVmuZgm8vvBpNeRhd6XSw==}
```
ijlee2, line 1508: We no longer install `@babel/traverse@7.22.20`.

pnpm-lock.yaml
```diff
@@ -8824,20 +8865,6 @@ packages:
     resolution: {integrity: sha512-VxPP4NqbUjj6MaAOafWeUn2cXWLcCtljklUtZf0Ind4XQ+QPtmA0b18zZy0jIQx+ExRVCR/ZQpBmik5lXshNsw==}
     dev: true
 
-  /fast-glob@2.2.7:
-    resolution: {integrity: sha512-g1KuQwHOZAmOZMuBtHdxDtju+T2RT8jgCC9aANsbpdiDDTSnjgfuVsIBNKbUeJI3oKMRExcfNDtJl4OhbffMsw==}
```
ijlee2, line 8828: We no longer install `fast-glob@2.2.7`.

PR #1792, opened 2023-11-07
fix missing custom formats in setLocal

29decibel:
## Why?

Currently we call `setLocale` in the constructor, which in turn calls `getOrCreateIntl` then calls `createIntl`. However, the custom formats at this moment is not registered yet. Thus it ended up with some locale formats being null.

First time calling `createIntl` triggered by `setLocale` in constructor:

Screenshot 2023-11-06 at 4 52 04 PM

Second time calling `createIntl` which in this time is fine, as the first call finished which has `_formats` created:

Screenshot 2023-11-06 at 4 52 31 PM

## Solution?

Moving `this.setLocale` after `this._formats` solved the issue. So that it will guarantee we always have custom formats (`this._formats` and `this._formatters`) exists.


ijlee2:
Hi, @29decibel. Can you provide more information by presenting a minimum reproducible app? You were calling `this.intl.setLocale()` in the constructor of where?

To my knowledge, for a long while, `ember-intl` has asked developers to call `setLocale()` in the `application` route's `beforeEach()` hook. So, at the moment, I think your case may be a one-off instance. The proposed code change alters internal implementations and may break other apps, so I'm hesitant to allow the change.

I'd recommend that you don't patch the `intl` service as you showed in the code change, but spend additional time looking into what your project might have done differently. What do you think?

ijlee2:
Won't be pursuing the proposed change at the moment.

PR #1809, opened 2023-11-30
Remove inheritance among helpers

ijlee2:


ember-intl/addon/services/intl.js
```diff
@@ -344,13 +344,15 @@ export default class IntlService extends Service {
...
    * @private
-   * @param {Function} fn
+   * @param fn
+   * @param context
    * @returns {Function} unsubscribed from localeChanged
    */
-  onLocaleChanged(...args) {
-    this._ee.on('localeChanged', ...args);
+  onLocaleChanged(fn, context) {
+    this._ee.on('localeChanged', fn, context);
```
ijlee2, line 352: `eventemitter3` expects 2 arguments.

https://github.com/primus/eventemitter3/tree/5.0.1#contextual-emits

ember-intl/addon/helpers/format-date.ts
```diff
@@ -1,18 +1,21 @@
-import { getOwner } from '@ember/application';
 import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
```
ijlee2, line 3: I didn't write `import { service } from '@ember/service';`, because `ember-intl` still supports `ember-source@3.28`.

https://blog.emberjs.com/ember-4-1-released/

ember-intl/addon/helpers/format-date.ts
```diff
@@ -5,7 +5,19 @@ import { isEmpty } from '@ember/utils';
...
+type Value = Params[0];
+type Options = Params[1];
+
+interface FormatDateSignature {
+  Args: {
+    Named?: Options & { allowEmpty?: boolean };
+    Positional: [Value?, Options?];
+  };
+  Return: string;
+}
```
ijlee2: I based the initial run on:

- https://github.com/ember-intl/ember-intl/blob/v6.2.2/ember-intl/addon/helpers/-format-base.d.ts#L5-L14
- https://github.com/ember-intl/ember-intl/blob/v6.2.2/ember-intl/addon/helpers/format-date.d.ts#L4

ember-intl/addon/helpers/format-date.ts
```diff
@@ -53,6 +53,6 @@ export default class FormatDateHelper extends Helper<FormatDateSignature> {
       // }
     }
 
-    return this.intl.formatDate(value, options);
+    return this.intl.formatDate(value!, options);
```
ijlee2, line 56: **TODO**

Replace `isEmpty()` with `value === undefined || value === null` (target `ember-intl@v7`). The former is ambiguous (allows too many inputs to result in `true`) and doesn't work with TypeScript, as we see from the need to add the non-null assertion operator.

PR #1812, opened 2023-12-01
Updated signatures for helpers

ijlee2:
## Why?

By removing code inheritance and manually created declaration files in #1809, I found that the helpers started to provide signatures that worsen DX (developer experience).

Their return type included `SafeString` and `undefined` (i.e. types that may often be incompatible with other addons). We can consider this a regression, since `v6.1.0` had guaranteed that the return type is always a `string`.

## Solution?

As the helpers had been created before TypeScript and Glint were introduced, I had to resolve mismatches between the helper code and its signature. For example, in certain error cases, the helpers return now an empty string `''` instead of `undefined`. This may or may not break a consuming app.

Going forward, we'll try to limit the API and be more strict about types.

### DIFF
--- a/.changeset/violet-feet-glow.md
+++ b/.changeset/violet-feet-glow.md
@@ -1,0 +1,6 @@
+---
+"ember-intl": patch
+"test-app": patch
+---
+
+Updated signatures for helpers


ember-intl/addon/helpers/format-date.ts
```diff
@@ -14,7 +14,7 @@ interface FormatDateSignature {
     Named?: Options & { allowEmpty?: boolean };
     Positional: [Value?, Options?];
   };
-  Return: string | undefined;
+  Return: string;
```
ijlee2, line 17: Technically, I'm fixing a regression here, because the return type had been `string` for all helpers in `v6.1.0`.

https://github.com/ember-intl/ember-intl/blob/v6.1.0/addon/helpers/-format-base.d.ts#L13

ember-intl/addon/helpers/format-message.ts
```diff
@@ -1,7 +1,6 @@
 import Helper from '@ember/component/helper';
 import { registerDestructor } from '@ember/destroyable';
 import { inject as service } from '@ember/service';
-import type { SafeString } from '@ember/template/-private/handlebars';
```
ijlee2, line 4: **TODO**

I think we need to remove `SafeString` as a possible type in `ember-intl`. To me, `SafeString` is an implementation detail of Ember.js, and addons shouldn't rely on it. (The consumers of an addon can cast the type when needed.)

test-app/tests/integration/helpers/format-date/api-that-may-be-deprecated-test.ts
```diff
@@ -13,6 +13,7 @@ module(
     test('input is null and allowEmpty is false', async function (assert) {
       await render(hbs`
         <div data-test-output>
+          {{! @glint-expect-error }}
```
ijlee2: In commit 3, we see that the tests from #1811 cause TS errors, but Glint wasn't able to catch them. This was partly due to the monorepo setup (the test app is separate from the addon), and maybe partly due to `ember-intl` not being a v2 addon.

ember-intl/addon/helpers/format-date.ts
```diff
@@ -12,7 +12,7 @@ type Options = Params[1];
 interface FormatDateSignature {
   Args: {
     Named?: Options & { allowEmpty?: boolean };
-    Positional: [Value?, Options?];
+    Positional: [Value?] | [Value | null | undefined, Options];
```
ijlee2: This fixes a type error that had slipped through `v6.0.0-beta.3`, when we started to allow passing a POJO as the 2nd argument.

https://ember-intl.github.io/ember-intl/versions/v6.3.0/docs/guide/migration-5-0-to-6-1#dynamically-set-named-arguments

ijlee2: **TECHNICAL RISK**

To my understanding,`Options` is a combination of an options hash supported by `@formatjs/intl` and a user-defined hash that provides data.

So, we may be unintentionally allowing end-developers to provide `@formatjs/intl` as a part of their user-defined hash.

PR #1818, opened 2023-12-09
Allowed the test helper addTranslations() to call settled() 

ijlee2:
## Why?

Similarly to `setLocale()` as described in #1817, the test helper `addTranslations()` doesn't really help end-developers write tests.

Depending on how an end-developer lazily loads translations, they will need to add `await settled();` (or possibly use something from `@ember/runloop`) so that adding translations has an effect on their application.

## Solution?

I allowed the test helper to call `settled()` so that end-developers don't have to.

In most cases, this change shouldn't break their tests. End-developers will want to search their test files for `addTranslations(` and `ember-intl/test-support`, then migrate code as follows:

```diff
- import { render, settled } from '@ember/test-helpers';
+ import { render } from '@ember/test-helpers';
import { hbs } from 'ember-cli-htmlbars';
import { addTranslations, setupIntl } from 'ember-intl/test-support';
import { setupRenderingTest } from 'ember-qunit';
import { module, test } from 'qunit';

module('Integration | Component | lazy-hello', function (hooks) {
  setupRenderingTest(hooks);
  setupIntl(hooks);

  test('it renders', async function (assert) {
    await render(hbs`
      <LazyHello @name="Zoey" />
    `);

    assert
      .dom('[data-test-message]')
      .doesNotIncludeText(
        'Hello, Zoey!',
        'Before translations are loaded, we cannot write assertions against a string.',
      )
      .hasText(
        't:lazy-hello.message:("name":"Zoey")',
        'Before translations are loaded, we should see the missing-message string.',
      )
      .hasText(
        t('lazy-hello.message', { name: 'Zoey' }),
        'Before translations are loaded, we can write assertions against the test helper t().',
      );

-     addTranslations({
-       'lazy-hello': {
-         message: 'Hello, {name}!',
-       },
-     });
-     await settled();
+     await addTranslations({
+       'lazy-hello': {
+         message: 'Hello, {name}!',
+       },
+     });

    assert
      .dom('[data-test-message]')
      .hasText(
        'Hello, Zoey!',
        'After translations are loaded, we can write assertions against a string.',
      )
      .doesNotIncludeText(
        't:lazy-hello.message:("name":"Zoey")',
        'After translations are loaded, we should not see the missing-message string.',
      )
      .hasText(
        t('lazy-hello.message', { name: 'Zoey' }),
        'After translations are loaded, we can write assertions against the test helper t().',
      );
  });
});
```

I will update the documentations for testing in a separate pull request.

### DIFF
--- a/ember-intl/addon-test-support/add-translations.ts
+++ b/ember-intl/addon-test-support/add-translations.ts
@@ -1,7 +1,6 @@
 import { assert } from '@ember/debug';
 import { get } from '@ember/object';
-import type { TestContext } from '@ember/test-helpers';
-import { getContext } from '@ember/test-helpers';
+import { getContext, settled, type TestContext } from '@ember/test-helpers';
 import type { IntlService } from 'ember-intl';
 import type { Translations } from 'ember-intl/types';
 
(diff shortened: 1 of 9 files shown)


ember-intl/addon-test-support/add-translations.ts
```diff
@@ -14,42 +12,49 @@ function pickLastLocale(localeName: string | string[]): string {
...
- * parameter, the `localeName`, is optional and will default to the last
- * currently enabled locale. This means, that if you invoke this helper with
- * just translations, they will be added to the last locale and all other
- * locales will be tried before.
+ * Adds translations, as if the end-developer had somehow added them.
+ *
+ * The first parameter, `localeName`, is optional and defaults to
+ * the current locale (the last enabled locale). So if you invoke the
+ * test helper with just the translations, they will be added to the
+ * last locale and all other locales will be tried before.
```
ijlee2, line 20: I didn't understand the last part, so I left the description as is. I found the documentation for testing also unclear, and will need to update its texts.

ember-intl/addon-test-support/add-translations.ts
```diff
@@ -14,42 +12,49 @@ function pickLastLocale(localeName: string | string[]): string {
...
+export async function addTranslations(
   localeNameOrTranslations: string | Translations,
   translations?: Translations,
-): void {
+): Promise<void> {
   const { owner } = getContext() as TestContext;
 
   assert(
-    'The current test context has no owner. Did you forget to call `setupTest(hooks)`, `setupContext(this)` or some other test helper?',
-    typeof owner === 'object' && typeof owner.lookup === 'function',
```
ijlee2, line 40: This check is brittle, as it knows too much about Ember's implementation. Checking the existence of `owner` should be enough.

PR #1820, opened 2023-12-11
Rewrote tests for macros

ijlee2:
## Why?

The existing tests don't really show how the macros are to be used (in classic components), so I added rendering tests and rewrote the unit tests. These tests should help me fix type errors later.

(Note, the macros will likely be deprecated and removed in `v7.0.0`.)

### DIFF
--- a/test-app/tests/unit/macros/t-test.ts
+++ b/test-app/tests/unit/macros/t-test.ts
@@ -1,139 +1,77 @@
 import { setOwner } from '@ember/application';
 import EmberObject from '@ember/object';
-import { run } from '@ember/runloop';
 import type { TestContext as BaseTestContext } from '@ember/test-helpers';
-import type { IntlService } from 'ember-intl';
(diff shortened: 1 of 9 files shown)


test-app/tests/unit/macros/t-test.ts
```diff
@@ -1,139 +1,77 @@
...
+      }),
+    }).create();
 
-    Consider deleting the test.
-  */
-  skip('allows property to be overridden', function (this: TestContext, assert) {
-    this.object.set('tMacroProperty2', 'A new value');
-    assert.strictEqual(this.object.get('tMacroProperty2'), 'A new value');
+    setOwner(this.emberObject, this.owner);
   });
```
ijlee2, line 41: This test can no longer be valid, because setting a computed property has been deprecated.

https://deprecations.emberjs.com/v3.x#toc_computed-property-override

PR #1821, opened 2023-12-12
Fixed types for macros

ijlee2:


ember-intl/addon/macros/intl.ts
```diff
@@ -1,141 +1,42 @@
...
+  const dependentKeys = [
+    `${__intlInjectionName}.locale`,
+    ...(args as string[]),
+  ];
 
-  return computed(`${__intlInjectionName}.locale`, ...dependentKeys, function (
-    this: { [__intlInjectionName]?: IntlService },
-    propertyKey: string,
-  ) {
+  return computed(...dependentKeys, function (propertyKey: string) {
```
ijlee2, line 29: Based on https://github.com/emberjs/ember.js/blob/v5.4.0/packages/%40ember/-internals/metal/lib/computed.ts#L864-L869

PR #1823, opened 2023-12-12
Improved <template>-tag support

ijlee2:
## Why?

Currently, developers need to remember and write the full path to use one of `ember-intl`'s helpers in a `<template>`-tag component.

## Solution?

I re-exported the helpers in the `index` file so that they can be imported easily.

```diff
import type { TOC } from '@ember/component/template-only';
- import t from 'ember-intl/helpers/t';
+ import { t } from 'ember-intl';

interface HelloSignature {
  Args: {
    name: string;
  };
}

const HelloComponent: TOC<HelloSignature> =
  <template>
    <div data-test-message>
      {{t "hello.message" name=@name}}
    </div>
  </template>

export default HelloComponent;
```


test-app/tests/helpers/update-locale.ts
```diff
@@ -1,13 +0,0 @@
...
-
-export async function updateLocale(locale: 'de-de' | 'en-us'): Promise<void> {
-  const { owner } = getContext() as { owner: ApplicationInstance };
-
-  const intl = owner.lookup('service:intl') as IntlService;
-
-  intl.setLocale([locale]);
-
-  await settled();
-}
```
ijlee2, line 13: This should have been removed in #1817.

ember-intl/addon/index.ts
```diff
@@ -1 +1 @@
-export { default as IntlService } from './services/intl';
```
steveszc, line 2: Unfortunately this was a breaking change. It was previously possible to import the IntlService and subclass it in userland. With this change this is no longer possible. I will open an issue.

steveszc: Ahhh please disregard. I realized it is still possible to `import IntlService from 'ember-intl/services/intl';`.

ijlee2: No problem. Yep, the `index` file only re-exports things for an easy access, so you can continue to import the `intl` service with `import IntlService from 'ember-intl/services/intl';`.

I had meant to export only the type of the `intl` service in the `index` file (at least, for now), so it was my bad for introducing a bug that became a soft breaking change.

PR #1825, opened 2023-12-12
Support native types from ember-source

### LINKED ISSUE #1802: Incompatible with native `ember-source` types
`ember-intl` is trying to import `SafeString` from `@ember/template/-private/handlebars`, which doesn't exist without having the `@types/ember__template` package installed.

It also tries to import `EmberRunTimer` from `@ember/runloop/types` which fails as well.

jelhan:
Closes #1802


ember-intl/addon/services/intl.d.ts
```diff
@@ -1,6 +1,6 @@
-import type { EmberRunTimer } from '@ember/runloop/types';
+import type { Timer as EmberRunTimer } from '@ember/runloop';
 import Service from '@ember/service';
-import type { SafeString } from '@ember/template/-private/handlebars';
+import type { SafeString } from '@ember/template';
```
ijlee2, line 3: Something that may help solve the type problem more easily: I want to remove `SafeString` as a possible return type. That is, we assume that all helpers from `ember-intl` return `string`.

The error from `Timer as EmberRunTimer` might go away, if I can change `addon/services/intl.js` to `addon/services/intl.ts` (i.e. derive the types instead of manually defining a declaration file).

jelhan:
Incompatibility of type import for `Timer` from `@ember/runloop` between DefinetlyTyped and native types shipped with Ember has been fixed.

`SafeString` from `@ember/template` is still incompatible between DefinetlyTyped and native types.

jelhan:
> `SafeString` from `@ember/template` is still incompatible between DefinetlyTyped and native types.

I created a PR to DefinetlyTyped adding the missing type export of `SafeString`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68672

jelhan:
@ijlee2 Ready to be merged from my side.

Not sure if this resolves all incompatibilities. I struggle with testing locally as this addon does not build the types but publishes the TypeScript files. That's not what I expected. And it is at least not working when I try using this feature branch with `npm link`. But maybe not an issue when using the published package.

If it is still an issue, we can address publishing JavaScript files + type declarations instead of TypeScript files in a follow-up.

At least the incompatibilities between type imports here and supported by ember-source are fixed by this PR. And all those exports are now existing in the types published from DefinetlyTyped as well.

ember-intl/addon/services/intl.d.ts
ijlee2: I haven't forgotten about this, just haven't had motivation to work on `ember-intl` since the year. I will get to it some day. 🙂

ijlee2 (approved):
Let's try this out, much thanks for helping with updating the types on both sides (`ember-intl` as well as `@types/ember__*`).

PR #1832, opened 2024-02-20
WIP: Publish types in build package

jelhan:
This changes the publish setup to build the types with Glint in `declarations` directory and publish them.

The published package still contains TypeScript files. They are not transpiled to JavaScript before publishing. But they weren't before. So that's not a regression.

Not a TypeScript expert here. Mainly copied the setup from Ember Bootstrap.


jelhan:
This is still not fully working. The structure of `declarations` folder is not as expected. It is (simplified):

- declarations/
  - addon/
    - template-registry.d.ts
    - helpers/
      - t.d.ts
  - addon-test-support/
    - set-locale.d.ts
  - tests/
    - test-helper.d.ts

But I think it should be

- declarations/
  - template-registry.d.ts
  - helpers/
    - t.d.ts
  - test-support/
    - set-locale.d.ts

The wrong structure breaks type lookup in consuming applications.

I'm not sure why it's wrong. The TypeScript configuration looks the same as for Ember Bootstrap. And in Ember Bootstrap it is working fine.

Keeping it in WIP until we figured that one out.

ember-intl/package.json
```diff
@@ -17,21 +17,29 @@
   },
   "license": "MIT",
   "author": "jason.mitchell.w@gmail.com",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "declarations/*",
+        "declarations/*/index.d.ts"
```
jelhan, line 24: This is likely wrong causing the issues. See https://github.com/ember-cli/ember-cli/blob/fa8599058bde3c745a2b1156076955ca3543c857/blueprints/addon/index.js#L101-L107

ijlee2:
Will close this in favor of #1885.

PR #1836, opened 2024-02-27
Refactored intl service

ijlee2:
## Why?

The `intl` service, which lies at the core of the addon, is hard to understand.

## Solution?

I removed code that seemed to be unused and extracted methods that can be considered private. The refactors will help us reduce the unpacked size by ~1.5 kB.


ember-intl/addon/services/intl.d.ts
```diff
@@ -21,7 +21,7 @@ export interface TOptions {
 type FormatterProxy<T extends keyof IntlService['_formatters']> = (
   value: Parameters<IntlService['_formatters'][T]['format']>[1],
   formatOptions?: Parameters<IntlService['_formatters'][T]['format']>[2] & {
-    locale?: string | [string, ...string[]];
+    locale?: string | string[];
```
ijlee2, line 24: After merging the PR, I realized that `[string, ...string[]]` might have been used to indicate that the array has at least 1 element.

At a later point, consider reverting the type after refactoring the code in the addon more.

PR #1850, opened 2024-03-08
Refactored a few files in /lib

ijlee2:


packages/ember-intl/lib/utils/is-valid-locale-format.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * @deprecated Will be removed in v7.
+ *
+ * Do not use it.
+ */
```
ijlee2, line 5: Only used for `ember g blueprint` (this command will be deprecated).

Issue #1848, opened 2024-03-08
Memory leak regression after upgrading to `>= 6.3.0`


johanrd:
Hi. I have experienced memory leaks in my app after upgrading ember-intl from `v6.2.2` to `>v6.3.0`.

**Steps to reproduce:** 

1) Clone https://github.com/johanrd/ember-memory-leak
2) Check out the `ember-intl` branch
3) Run `pnpm start`
4) Transition between `about` and `dashboard` a few times, and see that e.g. the ChartJs Component piles up in memory each time you enter the `dashboard` route:

https://github.com/ember-intl/ember-intl/assets/4601554/21c7aaa6-2d5c-4c37-9dca-44763a5ec821

5) Downgrade to ember-intl `v6.2.2` and see that the memory leaks with the ChartJs Component are gone:

https://github.com/ember-intl/ember-intl/assets/4601554/7fa4c89f-2d31-47d8-89aa-0da3b10ade08

ijlee2:
Hi, @johanrd. Thanks for posting recordings to show some of the steps that you had taken. From the videos, though, I didn't understand what I'm supposed to look at, which suggests that there is a memory leak. Can you describe this to me?

Secondly, I found the [commit history](https://github.com/johanrd/ember-memory-leak/commits) hard to follow. Can you recreate the history so that each commit (each step) clearly indicates what's particular about that step?

At the moment, I have to doubt that `ember-intl` is the issue, because your reproduction repo includes a few other variables, such as `ember-truth-helpers` (in whose repo [you had raised a similar issue](https://github.com/jmurphyau/ember-truth-helpers/issues/205)), `ember-resources`, `ember-template-imports` (for `*.gts`), `ember-route-template`, and `chart.js`.

```json5
/* package.json */
{
  "dependencies": {
    "@ember/render-modifiers": "^2.1.0",
    "chart.js": "^4.4.2",
    "ember-intl": "^6.5.1",
    "ember-resources": "^7.0.0",
    "ember-route-template": "^1.0.3",
    "ember-truth-helpers": "^4.0.3",
    "reactiveweb": "^1.2.2"
  }
}
```

If you can write a reproduction app with the minimum necessary dependencies, I think it will help with your investigation more.

johanrd:
@ijlee2 Hi. Thanks for the quick reply.  Yes, the commit history is a mess – _memory leaks are a mess._

Don't follow the commit history. **Please follow my steps to reproduce instead**

[...]

PR #1851, opened 2024-03-10
Proof of concept fix for the memory leak documented in #1848.

johanrd:
## Why?
Proof of concept fix for the memory leak documented in #1848. I can verify that the leak is gone with this for the t-helper.

## Solution?
Currently, no destructor in ember-intl is called when destroying a component that consumes the t-helper (and possibly other helpers as well). Not sure why the destructor isn't called, but I found it worthwhile to try to explicitly destroy (as in pre v6.3.0).

It may not seem sexy to bring back the getOwner pattern, so I'm not sure that this is the best solution, yet it may be wise to be conservative until we know why the registerDestructor in `onLocaleChanged` did never run on component teardown.

Best regards,
johan

### DIFF
--- a/packages/ember-intl/addon/helpers/t.ts
+++ b/packages/ember-intl/addon/helpers/t.ts
@@ -1,5 +1,5 @@
+import { getOwner } from '@ember/application';
 import Helper from '@ember/component/helper';
-import { inject as service } from '@ember/service';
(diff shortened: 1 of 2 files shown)


johanrd:
Should be closed. There is a simpler fix in #1852

Issue #1858, opened 2024-04-12
Vulnerability in dependency `glob-parent` via `broccoli-merge-files`


ccdexcom:
- [X] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [X] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: N/A
- Ember CLI Version: N/A
- Ember Intl Version: 6.5.3
- Browser(s): N/A
- Node Version: v20.11.1 (likely N/A)

### Steps to Reproduce
[Example Repo](https://github.com/ccdexcom/pnpm-ember-intl)
1. Create a new node project (folder with root-level `package.json`)
2. `pnpm add ember-intl`
3. `pnpm audit`
4. `glob-parent` listed as a vulnerability
  ```
➜  pnpm-ember-intl git:(main) pnpm audit
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ glob-parent vulnerable to Regular Expression Denial of │
│                     │ Service in enclosure regex                             │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ glob-parent                                            │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Vulnerable versions │ <5.1.2                                                 │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Patched versions    │ >=5.1.2                                                │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Paths               │                                                        │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-ww39-953v-wcq6      │
└─────────────────────┴────────────────────────────────────────────────────────┘
1 vulnerabilities found
  ```

I've seen #1791 and #1666. I don't think #1791 fixes the root issue we're seeing with `ember-intl`. The `overrides` in #1791 I believe only have an effect on `ember-intl` as a project (cloned locally and installed and with its local lockfiles) but not on projects that use `ember-intl` and their lockfiles ([see example repo](https://github.com/ccdexcom/pnpm-ember-intl)).

There are other vulnerabilities coming in from `fast-glob` via `broccoli-merge-files` but `glob-parent` is an easy one to use as an example because it shows up with the `pnpm audit` command.

[...]

PR #1864, opened 2024-04-16
Remove broccoli-merge-files

LucasHill:
## Why?

There are multiple vulnerabilities being brought in by broccoli-merge-files. See: https://github.com/ember-intl/ember-intl/issues/1858

The overrides done here: https://github.com/ember-intl/ember-intl/pull/1791 actually do not assist consumers of this package, because they will be creating their own lockfile (with any package manager not just pnpm) which will not respect those overrides.

This branch is created off the v6.5.3 tag and it would be awesome if we could get it in as a patch version of 6.x, as waiting for a 7.0 release may be a while off and these are getting flagged in security systems right now. 

## Solution?

It turns out this dependency wasn't doing much for us as removing it was fairly trivial. Given that it appears that project may be abandoned, and there have been previous security issues related to it, I think it makes sense to just not use it.

The translation reducer class already exists to apply the logic around filtering/removing/defaulting translations and then creating intermediate json files that broccoli-merge-files then combined into one. Now the intermediate json files are not created, and instead the final `translations.js` is created without broccoli-merge-files. 

Note: There were 8 failing tests relating to date formatting that existed prior to any changes I made. I fixed all other broken tests though that this refactor caused.

Would be amazing if we could get this in and released! This addon has been extremely helpful for our team.

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -5,8 +5,6 @@
 const { existsSync } = require('node:fs');
 const { dirname, isAbsolute, join } = require('node:path');
 const mergeTrees = require('broccoli-merge-trees');
-const { BroccoliMergeFiles } = require('broccoli-merge-files');
-const stringify = require('json-stable-stringify');
 const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 
 const buildTranslationTree = require('./lib/broccoli/build-translation-tree');
@@ -100,21 +98,7 @@ module.exports = {
         },
       });
 
-      const flattenedTranslationTree = new BroccoliMergeFiles(
-        [translationTree],
-        {
-          outputFileName: 'translations.js',
-          merge: (entries) => {
-            const output = entries.map(([locale, translations]) => {
-              return [locale, JSON.parse(translations)];
-            });
-
-            return 'export default ' + stringify(output);
-          },
-        },
-      );
-
-      trees.push(flattenedTranslationTree);
+      trees.push(translationTree);
     }
 
     return this._super.treeForAddon.call(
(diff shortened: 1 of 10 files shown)


packages/ember-intl/lib/broccoli/translation-reducer/index.js
```diff
@@ -227,7 +227,7 @@ class TranslationReducer extends CachingWriter {
       }
 
       writeFileSync(
-        `${filepath}/${locale}.json`,
+        join(filepath, `${locale}.json`),
```
ijlee2, line 230: Not using `join()` may cause an issue on Windows.

packages/ember-intl/index.js
```diff
@@ -5,8 +5,6 @@
 const { existsSync } = require('node:fs');
 const { dirname, isAbsolute, join } = require('node:path');
 const mergeTrees = require('broccoli-merge-trees');
-const { BroccoliMergeFiles } = require('broccoli-merge-files');
-const stringify = require('json-stable-stringify');
```
ijlee2, line 9: Note to myself:

We may be able to replace `json-stable-stringify` with `JSON.stringify()`.

It might be that `json-stable-stringify` had been introduced in order to alphabetize the keys of nested objects for tests. If the order doesn't matter for consuming projects, we can uninstall the library and update the test assertions accordingly.

ijlee2 (approved):
Sorry for the long delay in testing and reviewing your pull request. I understand your code change now, and liked how you removed the dead code around `options.filename`.

I checked that, with your code change,

- Tests for the 3 new apps in the `docs` folder continued to pass.
- Tests in two production projects continued to pass.

PR #1865, opened 2024-04-18
Recommend using types/global.d.ts for glint

MichalBryxi:
## Why?

- When you bootstrap new app, there is no `types/my-app/index.d.ts`, should I create one? Would the `global.d.ts` do when it's already there? What is the format for translating my app name to `my-app` (kebab-case, snake_case, etc.)? All those questions should IMO be answered in easy to follow tutorial.

## Solution?

- I am *very* far from understanding *if* using `global.d.ts` is a bad idea, but it works, and the file is already there from a blueprint, so I _think_ it might make sense to use it.
- This PR is just to kickstart discussion on how to transfer the information in easiest possible way so that newcommers don't have to figure basic things.

### DIFF
--- a/.changeset/famous-olives-lie.md
+++ b/.changeset/famous-olives-lie.md
@@ -1,0 +1,5 @@
+---
+"docs-app-for-ember-intl": patch
+---
+
+Recommend using types/global.d.ts for glint.
--- a/docs/ember-intl/app/templates/docs/getting-started/quickstart.md
+++ b/docs/ember-intl/app/templates/docs/getting-started/quickstart.md
@@ -96,7 +96,7 @@ export default class ApplicationRoute extends Route {
 Update your template registry to extend this addon's. Check the [Glint documentation](https://typed-ember.gitbook.io/glint/environments/ember/using-addons#using-glint-enabled-addons) for more information.
 
 ```ts
-/* types/my-app/index.d.ts */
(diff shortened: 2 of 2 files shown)


ijlee2:
For some context, the text had been a copy-paste from another repo of mine. I'm still not sure of the differences between `types/global.d.ts` and `types/<package location name>/index.d.ts` in apps and v1 addons (when to use one over the other).

I think it's okay to show `types/global.d.ts` for apps, so that

- The path matches [that recommended for v1 addons](https://ember-intl.github.io/ember-intl/versions/v7.0.0-beta.2/docs/quickstart-addons#glint).
- The path matches [what Glint currently shows](https://typed-ember.gitbook.io/glint/environments/ember/using-addons#typing-your-dependencies).

Before I merge the PR, can you add a changeset (see [`CONTRIBUTING`](https://github.com/ember-intl/ember-intl/blob/v7.0.0-beta.2/CONTRIBUTING.md) for details) with a patch release for `docs-app-for-ember-intl`? It'd be also good to retitle the PR to something more specific, such as `Recommend using types/global.d.ts for glint`.

MichalBryxi:
@ijlee2 done and done. Hope this is the correct way. Happy to adjust if anything else is needed.

ijlee2 (approved):
Thanks!

PR #1879, opened 2024-05-14
DOCS: Fix migration test sample

mkszepp:
In migration sample there the assert in language `en` is not correct. It should be `Hello` instead of `Hallo`


ijlee2 (approved):
Thanks!

ijlee2 (approved):

PR #1880, opened 2024-05-14
Added @babel/core as a dependency (required by ember-cli-babel@v8)

mkszepp:
The standard of packages managers are, that they are installing automatically missing dependencies.
To avoid problems in consumer apps, we should disable auto installing dependencies by using a stricter dependency managment.

This is possible by adding `.npmrc` with some options.

Atm in `.npmrc` there was `strict-peer-dependencies=false`... this is the default value in pnpm 8 + 9 so we can remove it. https://pnpm.io/8.x/npmrc

### DIFF
--- a/.changeset/sweet-radios-care.md
+++ b/.changeset/sweet-radios-care.md
@@ -1,0 +1,6 @@
+---
+"ember-intl": patch
+"my-v2-addon": patch
+---
+
+Added @babel/core as a dependency (required by ember-cli-babel@v8)
--- a/.npmrc
+++ b/.npmrc
@@ -1,9 +1,1 @@
-# Docs: https://pnpm.io/npmrc
-# https://github.com/emberjs/rfcs/pull/907
-
-# we don't want addons to be bad citizens of the ecosystem
-auto-install-peers=false
-
-# we want true isolation,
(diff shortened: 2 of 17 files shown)


ijlee2 (approved):
@mkszepp While I agree with the spirit of [the RFC](https://rfcs.emberjs.com/id/0907-pnpm-support), I got the impression that `auto-install-peers=false` and `resolve-peers-from-workspace-root=false` aren't ready yet.

Looking at which packages you had to add to `{config,docs,tests}/**/package.json`, it seemed to me like `pnpm` didn't fully understand how the lint packages and Ember projects work, but applied its rules blindly. (Maybe turning these flags off will work better in `v9`?)

To minimize change, let's just add `@babel/core` to `ember-intl`'s dependencies, which follows the [upgrade guide](https://github.com/emberjs/ember-cli-babel/blob/v8.2.0/UPGRADING.md#upgrade-path-for-addons) and aligns with [Ember's blueprints](https://github.com/ember-cli/ember-cli/blob/v5.8.1/blueprints/addon/index.js#L59-L63). This would also help me cut a patch release for `ember-intl@v6`. Is this okay?

mkszepp:
@ijlee2 yes thats good

PR #1882, opened 2024-05-14
Refactored code in lib

ijlee2:
## Why?

To migrate `ember-intl` to v2 format, I need to understand the files in `lib` folder better and make `addon/index.js` as ordinary as possible.

I noticed that a couple of lines (e.g. `isLocalizationFramework`, `isAbsolutePath()`) don't apply anymore and should have been long removed. In addition to improving code readability, the pull request will reduce the package size by ~2kB.


packages/ember-intl/lib/utils/find-engine.js
```diff
@@ -1,18 +1,13 @@
...
-    addon.pkg.keywords.includes('ember-engine')
-  );
-}
-
-function findEngine(current) {
+function findEngine(addon) {
   do {
-    if (isEngine(current)) {
-      return current;
+    const isEngine = addon.pkg?.keywords.includes('ember-engine');
```
ijlee2, line 3: **TODO**

Needs an additional `?.`.

```ts
const isEngine = addon.pkg?.keywords?.includes('ember-engine');
```

PR #1885, opened 2024-05-20
Provided native types for intl service

ijlee2:
## Why?

Because the `intl` service is complex, it's difficult to [hand-write its declaration file](https://github.com/ember-intl/ember-intl/blob/v7.0.0-beta.5/packages/ember-intl/addon/services/intl.d.ts) and ensure that types are correct.

By reimplementing the service, with a minimum API and native types, we can help end-developers write code correctly. Note, the package now weighs at 87.2 kB (before: 98.5 kB).

## Solution?

I started with a blank canvas, reintroduced the code in pieces, and let the tests in `tests/ember-intl` indicate if I'm on the right track. To minimize the API, I analyzed which getters and methods are important.


packages/ember-intl/addon/-private/utils/locale.ts
```diff
@@ -38,11 +38,3 @@ export function hasLocaleChanged(
...
   return locale1.toString() !== locale2.toString();
 }
-
-/**
- * @private
- * @hide
- */
-export function normalizeLocale(locale: string): string {
-  return locale.replace(/_/g, '-').toLowerCase();
-}
```
ijlee2, line 48: **breaking**

End-developers are expected to provide the locale in the correct form.

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+  hasLocaleChanged,
+} from '../-private/utils/locale';
+import { flattenKeys, type Translations } from '../-private/utils/translations';
+import translations from '../translations';
+
+type OnFormatjsError = (error: Parameters<OnErrorFn>[0]) => void;
+
+export default class IntlService extends Service {
+  @tracked private _intls: Record<string, IntlShape> = {};
+  @tracked private _locale?: string[];
```
ijlee2, line 43: **breaking**

The `locale` getter (a copy of `_locale)` and setter no longer exist. End-developers are to use `setLocale()` to update `_locale`.

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+
+    if (!this._formats) {
+      // @ts-expect-error: Property 'resolveRegistration' does not exist on type 'Owner'
+      this._formats = getOwner(this).resolveRegistration('formats:main') ?? {};
+    }
+
+    // Hydrate
+    translations.forEach(([locale, translations]: [string, Translations]) => {
+      this.addTranslations(locale, translations);
+    });
```
ijlee2, line 74: **technical risk**

I'm not sure how this exactly works and what we can do instead to prepare for Embroider.

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  formatDate(
+    value: FormatDateParameters[0] | undefined | null,
+    options?: FormatDateParameters[1] & {
+      locale?: string;
+    },
+  ): string {
+    if (value === undefined || value === null) {
+      return '';
+    }
```
ijlee2, line 114: **breaking**

For consistency, all `format*()` methods will return an empty string when `value` is `undefined` or `null`.

(`@formatjs/intl` may or may not accept these values.)

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore: Argument of type 'Timer | undefined' is not assignable to parameter of type 'EmberRunTimer | undefined'
+      // eslint-disable-next-line ember/no-runloop
+      this._timer = next(() => {
+        this._eventEmitter.emit('localeChanged');
+
+        this.updateDocumentLanguage();
+      });
+    }
```
ijlee2, line 236: These lines of code come from the `locale` setter (now removed).

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+    options?: FormatMessageParameters[1] & {
+      htmlSafe?: boolean;
+      locale?: string;
+    },
+  ): string {
+    const locales = options?.locale ? [options.locale] : this._locale!;
+
+    let translation: string | undefined;
+
+    for (let i = 0; i < locales!.length; i++) {
```
ijlee2, line 252: **breaking**

`options.default` is no longer allowed. End-developers are also responsible for ensuring that `key` is a string.

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+    if (!messages) {
+      return;
+    }
+
+    const translation = (messages as unknown as Record<string, string>)[key] as
+      | string
+      | undefined;
+
+    return translation;
+  }
```
ijlee2, line 330: **breaking**

The code, now private, used to be a part of [`lookup()`](https://github.com/ember-intl/ember-intl/blob/v7.0.0-beta.5/packages/ember-intl/addon/services/intl.js#L146-L160) (public, now removed).

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+    registerDestructor(context, () => {
+      this._eventEmitter.off('localeChanged', fn, context);
+    });
+  }
+
+  private updateDocumentLanguage(): void {
+    const dom = getDOM(this);
+
+    if (dom) {
+      const html = dom.documentElement;
```
ijlee2, line 357: **technical risk**

The `getDom()` relies on implementation details from Ember, starting with many years ago. Check if it's possible to inject the `-document` service instead.

packages/ember-intl/addon/services/intl.ts
```diff
@@ -0,0 +1,397 @@
...
+
+      html.setAttribute('lang', this.primaryLocale);
+    }
+  }
+
+  private updateIntl(
+    locale: string | string[],
+    messages?: Record<string, unknown>,
+  ) {
```
ijlee2, line 366: Used to be called [`getOrCreateIntl()`](https://github.com/ember-intl/ember-intl/blob/v7.0.0-beta.5/packages/ember-intl/addon/services/intl.js#L182-L202) (a misnomer).

PR #1889, opened 2024-05-20
Restricted API for helpers

### LINKED ISSUE #1735: Consider more forgiving DX

### Environment

- Ember Version: ~4.8.0
- Ember CLI Version: ~4.8.0
- Ember Intl Version: ^5.7.2
- Browser(s): Any :)
- Node Version: v16.18.0

### Steps to Reproduce

1. As a developer I'm naturally having only limited resources / knowledge / foresight to what _might_ happen. Especially when it comes to UI developers & APIs that might change / stop working without any warning and without prior testing.
2. As a user I would rather have an app that works _somehow_ rather than _not at all_. Having incorrectly displayed date, translation, etc. is not great, but it's way better than having an UI that does not work at all.

Because of the two points above I'd like to propose `ember-intl` not being overly aggressive on missing (undefined) values. Any code example below will blow up _the entire application_ with the message attached:

```hbs
## format-time
{{!-- Uncaught Error: <sandbox@helper:format-time::ember117> helper requires value attribute. --}}
{{!-- {{format-time this.nonexistent}} --}}

## format-number
{{!-- Uncaught Error: <sandbox@helper:format-number::ember117> helper requires value attribute. --}}
{{!-- {{format-number this.nonexistent}} --}}

## format-number(currency):
{{!-- Uncaught Error: <sandbox@helper:format-number::ember117> helper requires value attribute. --}}
{{!-- {{format-number this.nonexistent currency="USD" style="currency"}} --}}

format-time:
{{!-- Uncaught Error: <sandbox@helper:format-time::ember117> helper requires value attribute. --}}
{{!-- {{format-time this.nonexistent format="hhmmss"}} --}}

format-message:
{{!-- Uncaught Error: <sandbox@helper:format-message::ember117> helper requires value attribute. --}}
{{!-- {{format-message
  this.nonexistent
}} --}}
```

Since `this.nonexistent` is a variable that could come from many various sources (usually API) I as a developer have no control of it _always_ being filled. Thus I will need to put `if / else` block around every single call of these helpers.

Now I'm not saying the developers should not do that to catch error states, but that's not how real world works and people will have code without those safeguards. And then a component that _formats_ a date (or similar) just blows up entire application. And I think that should never happen.

Thoughts?

ijlee2:
## Why?

Closes #1735.

Before, the helpers had used the internal state `allowEmpty`, the user preference `options.allowEmpty`, and `isEmpty()` from `@ember/utils`, to handle an input that isn't well-defined. `isEmpty()` had made maintenance especially difficult, because [it returns `true` for many JavaScript values](https://api.emberjs.com/ember/5.8/functions/@ember%2Futils/isEmpty).

## Solution?

From now on, the `intl` service's `format*()` and `t()` methods form the source of truth. The helpers simply call these methods.


packages/ember-intl/addon-test-support/add-translations.ts
```diff
@@ -15,14 +14,7 @@ export async function addTranslations(
   locale: string,
   translations: Translations,
 ): Promise<void> {
-  const context = getContext();
-
-  assert(
-    'To use `addTranslations()`, make sure to call `setupTest()`, `setupRenderingTest()`, or `setupApplicationTest()`.',
-    context,
-  );
```
ijlee2, line 23: I decided to remove `assert()`, as it's unlikely that developers can run a test without `setupTest()`, `setupRenderingTest()`, or `setupApplicationTest()`.

Note, the `assert()` had provided a more user-friendly error message.

packages/ember-intl/addon/helpers/format-date.ts
```diff
@@ -10,17 +9,15 @@ type Options = FormatParameters[1];
 
 interface FormatDateSignature {
   Args: {
-    Named?: Options & { allowEmpty?: boolean };
-    Positional: [Value] | [Value, Options];
+    Named?: Options;
+    Positional: [Value];
```
ijlee2, line 13: Due to the use of inheritance (`addon/helpers/-format-base.js`), #1633 had accidentally increased the API of `{{format-*}}` helpers.

PR #1895, opened 2024-06-06
Reintroduced extend to handle deeply nested translation folders

ijlee2:


docs/my-app-with-namespace-from-folders/translations/components/de-de.yml
```diff
@@ -0,0 +1 @@
+title: Komponenten
```
ijlee2, line 1: When these "in-between" translation files are added, existing assertions in `tests/acceptance/index-test.ts` fail.

Issue #1908, opened 2024-06-26
Cannot silence missing translations output on build


Techn1x:
I have `errorOnMissingTranslations: false` in my config. While it (correctly) doesn't error, it forces the missing translations to appear in output.

eg
```
[ember-intl] "promo.msUpsell.intro" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
[ember-intl] "promo.msUpsell.getStarted" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
[ember-intl] "promo.msUpsell.settingUp" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
[ember-intl] "promo.msUpsell.youCanAlso" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
[ember-intl] "promo.msUpsell.copyClasses" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
[ember-intl] "promo.msUpsell.ifYouHaveOne" was not found in "en-au", "en-ca", "en-gb", "en-us", "ko", "zh"
```

The relevant code is here;
https://github.com/ember-intl/ember-intl/blob/215f867e467a76dfacdc2b0b002146d7dece6dde/packages/ember-intl/lib/broccoli/translation-reducer/index.js#L175-L190

In theory, I just need to set `verbose: false` in my options. I tried this, however it is not used - options.verbose is only defined by the following line, where `this.isSilent` seems to be a dead reference, so verbose is _always_ true
https://github.com/ember-intl/ember-intl/blob/215f867e467a76dfacdc2b0b002146d7dece6dde/packages/ember-intl/index.js#L118 

---

I'd be happy to put a PR together, likely a simple fix, but I am not sure the approach we want. Do we want to expose `verbose` as a configurable option for users, with default as false? Or something else?

ijlee2:
@Techn1x I've seen similar log messages in my project, but didn't get to look into the cause. I'm glad that you got to a bit.

Feel free to try out possible solutions using the "docs" projects in the `docs` folder, and create a pull request if you find a good solution. I'll be on vacation and won't be able to look into the issue for some time. If you need the log messages gone soon, I'd suggest patching `ember-intl` for now.

[...]

PR #1909, opened 2024-07-05
Silence missing translations output on build by default

Techn1x:
## Why?

Resolves: https://github.com/ember-intl/ember-intl/issues/1908

## Solution?

Set its default to false

### DIFF
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -25,6 +25,7 @@ const defaultConfig = {
   },
   stripEmptyTranslations: false,
   wrapTranslationsWithNamespace: false,
+  verbose: false,
 };
 
 module.exports = {
@@ -93,6 +94,7 @@ module.exports = {
       requiresTranslation,
       stripEmptyTranslations,
       wrapTranslationsWithNamespace,
+      verbose,
     } = this.configOptions;
 
     const [translationTree, addonsWithTranslations] = buildTranslationTree(
@@ -115,8 +117,8 @@ module.exports = {
       outputPath: options.outputPath ?? outputPath,
       requiresTranslation,
       stripEmptyTranslations,
-      verbose: !this.isSilent,
       wrapTranslationsWithNamespace,
+      verbose,
     });
   },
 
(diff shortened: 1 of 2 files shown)


packages/ember-intl/index.js
```diff
@@ -115,8 +117,8 @@ module.exports = {
       outputPath: options.outputPath ?? outputPath,
       requiresTranslation,
       stripEmptyTranslations,
-      verbose: !this.isSilent,
```
Techn1x, line 118: this.isSilent points at nothing, so ended up being always true

packages/ember-intl/index.js
```diff
@@ -93,6 +94,7 @@ module.exports = {
       requiresTranslation,
       stripEmptyTranslations,
       wrapTranslationsWithNamespace,
+      verbose,
     } = this.configOptions;
```
Techn1x, line 98: technically the user config could override this if they really wanted to see the missing translations in build output but not error for them. But we don't have to make that public

When embroider macros is implemented for config, a cleanup of options can be done then

ijlee2: I agree with how you had set `verbose: false` in line 28, to still allow the possibility of an end-developer to log missing translations, but not to document it as a public API (by updating the blueprint for `config/ember-intl.js`), in case we want to drop it later.

packages/ember-intl/index.js
ijlee2: It looks like the variable `isSilent` is [a remnant from `v4.x` or earlier](https://github.com/ember-intl/ember-intl/blob/v4.0.1/index.js#L61). We can definitely remove it.

ijlee2 (commented):
Thanks again for investigating the cause and creating the pull request. Can you make 2 changes before we merge it?

- Keep the object key names alphabetized (lines 28, 97, 121).
- Add a changeset ([from the workspace root, run `pnpm changeset`](https://github.com/ember-intl/ember-intl/blob/v7.0.3/CONTRIBUTING.md)) to summarize your change (the summary will later be included in the `CHANGELOG`). You can answer `patch` to the changeset CLI's first question.

Techn1x:
Sure thing, I'll jump on that first thing tomorrow

Techn1x:
Done!

ijlee2 (approved):
✨

PR #1913, opened 2024-08-13
Used pnpm@9 to manage dependencies

ijlee2:


package.json
```diff
@@ -24,7 +23,8 @@
     "@changesets/get-github-info": "^0.6.0"
   },
   "engines": {
-    "node": "18.* || >= 20"
+    "node": "18.* || >= 20",
+    "pnpm": ">= 9"
```
ijlee2, line 27: This only affects local development (i.e. projects that consume `ember-intl` are free to use whichever package manager and version they want), because the workspace root `package.json` describes the project.

PR #1918, opened 2024-08-26
Listed @ember/test-helpers as a peer dependency

### LINKED ISSUE #1916: Import error after updating `@ember/test-helpers` to version `4.x`
### Description:

After updating the `@ember/test-helpers` package to version 4.x, I encountered a build error when trying to build my Ember application that uses ember-intl.

### Steps to reproduce:
 - Update `@ember/test-helpers` to version 4.x.
 - Build the Ember application.
 - The following error occurs:
 ```shell
Build Error (WebpackBundler)

ember-intl tried to import "@ember/test-helpers" in "ember-intl/test-support/add-translations.js" from addon code, but "@ember/test-helpers" is a devDependency. You may need to move it into dependencies.
```

### Expected behavior:
The application should build without errors after updating the `@ember/test-helpers` package.

### Actual behavior:
The build fails due to an improper import of `@ember/test-helpers` from the `ember-intl/test-support/add-translations.js` file. The error message suggests that `@ember/test-helpers` is currently listed as a devDependency, but it is being imported from within addon code, which causes the build to fail.

### Suggested solution:
It seems that `@ember/test-helpers` should be moved from devDependencies to dependencies in the `ember-intl` package's package.json. This would ensure that it is correctly included in production builds.

### Environment:
- Ember Version: ~5.4.x
- Ember CLI Version: ~5.4.x
- Ember Intl Version: ~7.0.5
- Browser(s): Chome
- Node Version: ~20.x

### PR DESCRIPTION
**Pull request by ijlee2**
## Why?

Closes #1916.

## Solution?

I listed `@ember/test-helpers` as a peer dependency so that consuming projects can decide whether they need to install it. The range of supported versions is currently wide (`2.9.4` and above) because `ember-intl@v7` supports Ember 3.28.

### DIFF
--- a/packages/ember-intl/package.json
+++ b/packages/ember-intl/package.json
@@ -56,7 +56,7 @@
   "devDependencies": {
     "@ember/optional-features": "^2.1.0",
     "@ember/string": "^3.1.1",
-    "@ember/test-helpers": "^3.3.1",
+    "@ember/test-helpers": "^4.0.2",
     "@embroider/test-setup": "^4.0.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
--- a/packages/ember-intl/package.json
+++ b/packages/ember-intl/package.json
@@ -97,9 +97,13 @@
     "webpack": "^5.94.0"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "^2.9.4 || ^3.2.0 || ^4.0.0",
     "typescript": "^5.0.0"
   },
   "peerDependenciesMeta": {
+    "@ember/test-helpers": {
+      "optional": true
+    },
     "typescript": {
       "optional": true
     }
--- a/packages/ember-intl/package.json
+++ b/packages/ember-intl/package.json
@@ -56,7 +56,7 @@
   "devDependencies": {
     "@ember/optional-features": "^2.1.0",
     "@ember/string": "^3.1.1",
-    "@ember/test-helpers": "^4.0.2",
+    "@ember/test-helpers": "^4.0.4",
     "@embroider/test-setup": "^4.0.0",
     "@glimmer/component": "^1.1.2",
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: packages/ember-intl/package.json [hunk 1]
```diff
@@ -97,9 +97,13 @@
     "webpack": "^5.94.0"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "^3.2.0 || ^4.0.0",
     "typescript": "^5.0.0"
   },
   "peerDependenciesMeta": {
+    "@ember/test-helpers": {
+      "optional": false
```
elwayman02, line 105: This should likely be `optional: true`, because you can't guarantee that your consumers are all going to use the testing util which depends on this package.

packages/ember-intl/package.json
```diff
@@ -97,9 +97,13 @@
     "webpack": "^5.94.0"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "^2.9.4 || ^3.0.0 || ^4.0.0",
```
ijlee2, line 100: **TODO**

Remove `2.9.4` when dropping support for Ember 3.28.

packages/ember-intl/package.json
```diff
@@ -97,9 +97,13 @@
     "webpack": "^5.94.0"
   },
   "peerDependencies": {
+    "@ember/test-helpers": "^2.9.4 || ^3.2.0 || ^4.0.0",
```
ijlee2, line 100: **TODO**

Remove `^2.9.4` when dropping support for Ember 3.28.

packages/ember-intl/package.json
ijlee2: Thanks for the suggestion. I think it makes sense to set `optional` to `true` for `@ember/test-helpers`. While we can assume that `@ember/test-helpers` is always present in apps and v1 addons, it may not be in v2 addons that only use helpers from `ember-intl`.

PR #1921, opened 2024-09-02
Created a page that explains how to use ember-intl's helpers in <template> tag

MichalBryxi:
- Add documentation on how to use `ember-intl` together with `<template>` tag

## Why?

- All the examples in the documentation show classic way of using helpers from the global scope
- Template imports require two bits of new knowledge:
  - Invocation is now `camelCased` instead of `kebab-cased`. Although this is generic knowledge from template tags, the docs of ember-intl were lacking to provide *an* example.
  - Where to import the files from? This can be from virtually anywhere, so having this documented is IMO a big win.

## Solution?

- New `Helpers / <template> tag` documentation page

### DIFF
--- a/.changeset/real-years-jog.md
+++ b/.changeset/real-years-jog.md
@@ -1,0 +1,5 @@
+---
+"docs-app-for-ember-intl": minor
+---
+
+Created a page that explains how to use ember-intl's helpers in <template> tag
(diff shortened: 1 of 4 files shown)


ijlee2:
@MichalBryxi I rebased your branch and updated the code example to show all helpers in the `import` statement. What do you think?

MichalBryxi:
@ijlee2 ship it! ❤️

ijlee2 (approved):
Thanks!

Issue #1829, opened 2024-02-02
How to use `publicOnly: true` in Embroider apps


amk221:
In a classic app, it's possible to use the instructions [here](https://github.com/ember-intl/ember-intl/blob/v6.5.0/docs-app/app/templates/docs/guide/asynchronously-loading-translations.md) to produce JSON files in a translations directory. The app knows the name of the fingerprinted file to load, because it can read the assetMap.

But in an Embroider build, the `translations/*.json` files no longer get processed by broccolli-asset-rev, and as such they are not fingerprinted. And since there is no longer an assetMap, there doesn't seem to be a way to load the translation files into the app. I thought perhaps I could leave `publicOnly` as false, and somehow use Webpack to produce the json files, but I'm not sure that's possible.

Could the guidance in the [asynchronously-loading-translations.md](https://github.com/ember-intl/ember-intl/blob/v6.5.0/docs-app/app/templates/docs/guide/asynchronously-loading-translations.md) be updated?

- [x] I am on the [latest](https://github.com/ember-intl/ember-intl/releases/latest) ember-intl version
- [x] I have searched the [issues](https://github.com/ember-intl/ember-intl/issues) of this repo and believe that this is not a duplicate

### Environment

- Ember Version: 5.5.0
- Ember CLI Version: 5.3.0
- Ember Intl Version: 6.4.0
- Browser(s): 
- Node Version: 18.19.0

amk221:
There's some more context in this thread

https://github.com/embroider-build/embroider/issues/755

veelci:
I've just run into this as well w/ embroider.

Ember Version: 4.12.3
Ember Cli Version: 4.12.1
Ember Intl Version: 6.4.0
Node Version: 18

ijlee2:
@amk221 I didn't fully understand the problem that you had encountered. Without seeing how your project is set up (conversely, your seeing mine), I think it may be hard to discuss the issue.

What I can show at the moment is, setting `publicOnly: true` may work in an Embroider app, with all `static*` flags set to `true` and route splitting enabled as well.

The screenshots below show an Embroider app with `ember-source@5.6.0`. To my knowledge, this app is a normal Ember app (no special configurations). When I changed `publicOnly` from `false` to `true`, I saw that translations are missing, as we expect.

<img width="900" alt="" src="https://github.com/ember-intl/ember-intl/assets/16869656/58be0a16-18e3-4187-94da-f4b2360077b1">

[...]

PR #1927, opened 2024-10-15
Update instructions for lazy-loading translations for Embroider apps

ijlee2:
## Why?

Documents the latest findings, reported in https://github.com/ember-intl/ember-intl/issues/1829#issuecomment-2368364173. By replacing `asset/resource` with `asset/source`, we can avoid calling `fetch()`.

### DIFF
--- a/docs/my-app-with-lazy-loaded-translations/ember-cli-build.js
+++ b/docs/my-app-with-lazy-loaded-translations/ember-cli-build.js
@@ -28,11 +28,8 @@ module.exports = function (defaults) {
         module: {
           rules: [
             {
-              generator: {
-                filename: '[path][name]-[hash][ext][query]',
-              },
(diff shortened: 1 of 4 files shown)


docs/my-app-with-lazy-loaded-translations/ember-cli-build.js
```diff
@@ -28,11 +28,8 @@ module.exports = function (defaults) {
         module: {
           rules: [
             {
-              generator: {
-                filename: '[path][name]-[hash][ext][query]',
-              },
```
ijlee2, line 33: I believe we can remove this when using `asset/source`.

The [documentation for `generator.filename`](https://webpack.js.org/configuration/module/#rulegeneratorfilename) said that it "works only with `asset` and `asset/resource` module types." The file names in `dist/translations` didn't show any hash.

Issue #1669, opened 2022-05-18
add support for formatRange API


mtoygar:
`Intl.DateTimeFormat` and formatjs have a support for localizing [date ranges](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange). Do you have plans to add support for this API?

Adding formatjs docs for the sake of completeness
https://formatjs.io/docs/polyfills/intl-datetimeformat/#features
https://formatjs.io/docs/react-intl/api/#formatdatetimerange

PR #1931, opened 2024-12-20
Created {{format-date-range}} helper

miguelcobain:
This PR implements the `formatDateTimeRange` helper which uses formatjs' own `formatDateTimeRange`, similar to what we have on `formatDate` and `formatTime`.

Couple of notes:
- formatjs `formatDateTimeRange` apparently does not support `defaultFormats` perhaps we should add this to formatjs?
- currently, `formatDateTimeRange` does not support date strings as inputs. I submitted a [PR to formatjs repo](https://github.com/formatjs/formatjs/pull/4784) to add this, in order to be consistent with the other methods.

For these reasons I removed the strings as input tests and commented out the format='hhmmss' assertions.

This PR is still missing documentation, but I wanted to get an thumbs up from maintainers before tackling that part.

### DIFF
--- a/packages/ember-intl/addon/-private/formatjs/format-date-time-range.ts
+++ b/packages/ember-intl/addon/-private/formatjs/format-date-time-range.ts
@@ -1,0 +1,12 @@
+import type { IntlShape } from '@formatjs/intl';
+
+export type FormatDateTimeRangeParameters = Parameters<
+  IntlShape<string>['formatDateTimeRange']
+>;
+
+export function formatDateTimeRange(
+  intlShape: IntlShape,
+  ...[from, to, formatOptions]: FormatDateTimeRangeParameters
+): string {
+  return intlShape.formatDateTimeRange(from, to, formatOptions);
+}
--- a/packages/ember-intl/addon/helpers/format-date-time-range.ts
+++ b/packages/ember-intl/addon/helpers/format-date-time-range.ts
@@ -19,14 +19,6 @@ interface FormatDateTimeRangeSignature {
 export default class FormatDateTimeRangeHelper extends Helper<FormatDateTimeRangeSignature> {
   @service declare intl: IntlService;
 
-  constructor() {
-    // eslint-disable-next-line prefer-rest-params
-    super(...arguments);
-
-    // @ts-expect-error: Property 'onLocaleChanged' is private and only accessible within class 'IntlService'.
-    this.intl.onLocaleChanged(this.recompute, this);
-  }
-
   compute(
     [from, to]: FormatDateTimeRangeSignature['Args']['Positional'],
     options: FormatDateTimeRangeSignature['Args']['Named'],
--- a/packages/ember-intl/addon/helpers/format-date-time-range.ts
+++ b/packages/ember-intl/addon/helpers/format-date-time-range.ts
@@ -1,0 +1,36 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+import type IntlService from '../services/intl';
+
+type FormatParameters = Parameters<IntlService['formatDateTimeRange']>;
+type From = FormatParameters[0];
+type To = FormatParameters[1];
+type Options = FormatParameters[2];
+
+interface FormatDateTimeRangeSignature {
+  Args: {
+    Named?: Options;
+    Positional: [From, To];
+  };
+  Return: string;
+}
+
+export default class FormatDateTimeRangeHelper extends Helper<FormatDateTimeRangeSignature> {
+  @service declare intl: IntlService;
+
+  constructor() {
+    // eslint-disable-next-line prefer-rest-params
+    super(...arguments);
+
+    // @ts-expect-error: Property 'onLocaleChanged' is private and only accessible within class 'IntlService'.
+    this.intl.onLocaleChanged(this.recompute, this);
+  }
+
+  compute(
+    [from, to]: FormatDateTimeRangeSignature['Args']['Positional'],
+    options: FormatDateTimeRangeSignature['Args']['Named'],
+  ) {
+    return this.intl.formatDateTimeRange(from, to, options);
+  }
+}
--- a/packages/ember-intl/addon/services/intl.ts
+++ b/packages/ember-intl/addon/services/intl.ts
@@ -6,7 +6,7 @@ import { tracked } from '@glimmer/tracking';
 
 import type {
   FormatDateParameters,
-  FormatDateTimeRangeParameters,
+  FormatDateRangeParameters,
   FormatListParameters,
   FormatMessageParameters,
   FormatNumberParameters,
@@ -19,7 +19,7 @@ import {
   createIntl,
   createIntlCache,
   formatDate,
-  formatDateTimeRange,
+  formatDateRange,
   formatList,
   formatMessage,
   formatNumber,
@@ -137,10 +137,10 @@ export default class IntlService extends Service {
     return formatDate(intlShape, value, options);
   }
 
-  formatDateTimeRange(
-    from: FormatDateTimeRangeParameters[0] | undefined | null,
-    to: FormatDateTimeRangeParameters[1] | undefined | null,
-    options?: FormatDateTimeRangeParameters[2] & {
+  formatDateRange(
+    from: FormatDateRangeParameters[0] | undefined | null,
+    to: FormatDateRangeParameters[1] | undefined | null,
+    options?: FormatDateRangeParameters[2] & {
       locale?: string;
     },
   ): string {
@@ -154,7 +154,7 @@ export default class IntlService extends Service {
 
     const intlShape = this.getIntlShape(options?.locale);
 
-    return formatDateTimeRange(intlShape, from, to, options);
+    return formatDateRange(intlShape, from, to, options);
   }
 
   formatList(
--- a/packages/ember-intl/addon/services/intl.ts
+++ b/packages/ember-intl/addon/services/intl.ts
@@ -6,6 +6,7 @@ import { tracked } from '@glimmer/tracking';
 
 import type {
   FormatDateParameters,
+  FormatDateTimeRangeParameters,
   FormatListParameters,
   FormatMessageParameters,
   FormatNumberParameters,
@@ -18,6 +19,7 @@ import {
   createIntl,
   createIntlCache,
   formatDate,
+  formatDateTimeRange,
   formatList,
   formatMessage,
   formatNumber,
@@ -133,6 +135,27 @@ export default class IntlService extends Service {
     const intlShape = this.getIntlShape(options?.locale);
 
     return formatDate(intlShape, value, options);
+  }
+
+  formatDateTimeRange(
+    from: FormatDateTimeRangeParameters[0] | undefined | null,
+    to: FormatDateTimeRangeParameters[1] | undefined | null,
+    options?: FormatDateTimeRangeParameters[2] & {
+      locale?: string;
+    },
+  ): string {
+    if (
+      from === undefined ||
+      from === null ||
+      to === undefined ||
+      to === null
+    ) {
+      return '';
+    }
+
+    const intlShape = this.getIntlShape(options?.locale);
+
+    return formatDateTimeRange(intlShape, from, to, options);
   }
 
   formatList(
--- a/packages/ember-intl/addon/services/intl.ts
+++ b/packages/ember-intl/addon/services/intl.ts
@@ -144,12 +144,11 @@ export default class IntlService extends Service {
       locale?: string;
     },
   ): string {
-    if (
-      from === undefined ||
-      from === null ||
-      to === undefined ||
-      to === null
-    ) {
+    if (from === undefined || from === null) {
+      return '';
+    }
+
+    if (to === undefined || to === null) {
       return '';
     }
 
--- a/tests/ember-intl/tests/integration/helpers/format-date-range/input-is-a-string-test.ts
+++ b/tests/ember-intl/tests/integration/helpers/format-date-range/input-is-a-string-test.ts
@@ -15,7 +15,7 @@ interface TestContext extends BaseTestContext {
 }
 
 module(
-  'Integration | Helper | format-date-time-range > input is a string',
+  'Integration | Helper | format-date-range > input is a string',
   function (hooks) {
     setupRenderingTest(hooks);
     setupIntl(hooks, 'en-us');
@@ -28,7 +28,7 @@ module(
     test('it returns a string', async function (this: TestContext, assert) {
       await render<TestContext>(hbs`
         <div data-test-output>
-          {{format-date-time-range this.from this.to}}
+          {{format-date-range this.from this.to}}
         </div>
       `);
 
@@ -41,7 +41,7 @@ module(
     test('it returns a new value when the locale is changed', async function (this: TestContext, assert) {
       await render<TestContext>(hbs`
         <div data-test-output>
-          {{format-date-time-range this.from this.to}}
+          {{format-date-range this.from this.to}}
         </div>
       `);
 
@@ -56,7 +56,7 @@ module(
     test('we can format the date', async function (this: TestContext, assert) {
       await render<TestContext>(hbs`
         <div data-test-output="2">
-          {{format-date-time-range
+          {{format-date-range
             this.from
             this.to
             hour="numeric"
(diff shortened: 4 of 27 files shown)


ijlee2:
Hi, @miguelcobain. Thanks for opening a pull request to start a conversation and providing a fix to the `formatjs` repo.

Your PR seems to be related to a feature request from the past (#1669). There are a few things that I don't understand at the moment (as I have yet to encounter the need for [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat)) and I prefer keeping an API to a minimum to reduce maintenance cost, so I'd like to get some clarifications from you before reviewing your PR.

- What are real-life use cases for a date-time range (perhaps from your team's product)? If we look at the asserted output `23.–26.01.2014`, for example, I don't see why a company or person would want such text to be shown in their app.

- Why does `ember-intl` need to provide a helper that wraps `Intl.DateTimeFormat`, other than "because `Intl` and `formatjs` do so"?

    If I understood some of the use cases right (e.g. to show `1/23/2014 – 2/26/2015`), wouldn't it be simpler (even more accurate/accommodating) if `ember-intl` asks end-developers to provide a translation that can handle a date or time range?

    ```yml
    # End-developers get to choose the format
    my-date-time-range: '{from} – {to}'
    ```

    ```hbs
    {{t "my-date-time-range" from=this.from to=this.to}}
    ```

    Note, the example above assumes that `this.from` and `this.to` are strings; they may or may not have been created using `ember-intl`'s `{{format-date}}` helper or another date-formatting library like `luxon`. If end-developers need to render ranges often, they would create a helper in their project.

- [`formatjs`'s documentation for `FormattedDateTimeRange`](https://formatjs.github.io/docs/react-intl/components/#formatteddatetimerange) mentions the need to install their polyfill `@formatjs/intl-datetimeformat`. If this is true, `ember-intl` would have 1 additional dependency, possibly increasing the bundle size for many end-developers who don't need the date-time range, and further coupling its implementation details to `formatjs`'s (i.e. more difficult to move away from `formatjs` when needed). What are your thoughts on this matter?

miguelcobain:
@ijlee2 those are good questions. Allow me to provide my point of view on these things.

> What are real-life use cases for a date-time range (perhaps emanating from your team's product)? If we look at the asserted output 23.–26.01.2014, for example, I don't see why a company or person would want such text to be shown in their app.

Basically, any time you want to display a range of dates, the [`formatRange`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange) is really useful. Keep in mind that `formatRange(from, to)` is **very** different than doing `formatDate(from) - formatDate(to)`:
- different locales have different was to present date ranges. In your example, it seems that `23.–26.01.2014` is the localized way in german to display "from 23 to 26 of January 2014".
- `formatRange` is smart. E.g, consider this case: 
```js
const date1 = new Date(Date.UTC(1906, 0, 10, 10, 0, 0)); // Wed, 10 Jan 1906 10:00:00 GMT
const date2 = new Date(Date.UTC(1906, 0, 20, 10, 0, 0)); // Sat, 20 Jan 1906 10:00:00 
const fmt = new Intl.DateTimeFormat("en", {
  year: "numeric",
  month: "short",
  day: "numeric",
});
console.log(fmt.formatRange(date1, date2)); // 'Jan 10 – 20, 1906'
```

Because we set our month to `short`, we get a nicely compact representation without duplicating the `Jan`.  If you use 
```js
const fmt2 = new Intl.DateTimeFormat("en", {
  month: "short",
  day: "numeric",
});
```
You would get `Jan 10 – 20`, omitting the year. However, if the year was different, you would get something like `Jan 10, 1906 – Jan 20, 2000`. In short, `formatRange` is aware of what date components are similar to smartly omit them if needed, **and** it is aware of locales.

The polyfill's readme is a good read on the motivation behind `formatRange`: https://github.com/tc39/proposal-intl-DateTimeFormat-formatRange

> Why does ember-intl need to provide a helper that wraps Intl.DateTimeFormat, other than "because Intl and formatjs do so provide it"?

The main reason is to take advantage of automatic locale wiring to ember-intl's locale. If ember-intl doesn't provide this helper, then one would have to do that wiring manually. This includes recomputing the helper when the locale changes.
Another advantage is to use the built-in `formats.js` file to store our formats. Unfortunately, right now formatjs doesn't support that, but this has to be a bug on their end. Once that's solved, we would benefit from that.

> If I understood some of the use cases right (e.g. to show 1/23/2014 – 2/26/2015), wouldn't it be simpler (even more accurate/accommodating) if ember-intl asks end-developers to provide a translation that can handle a date or time range?

As I explained above, `{from} – {to}` is not an accurate implementation of what `formatRange` does. It's much smarter and locale-aware than that. I don't think you can implement `formatRange` in terms of simple string concatenation like that.

> [formatjs's documentation for FormattedDateTimeRange](https://formatjs.github.io/docs/react-intl/components/#formatteddatetimerange) mentions the need to install their polyfill @formatjs/intl-datetimeformat. If this is true, ember-intl would have 1 additional dependency, possibly increasing the bundle size for many end-developers who don't need the date-time range, and further coupling its implementation details to formatjs's (i.e. more difficult to move away from formatjs when needed). What are your thoughts on this matter?

I think browser support for `formatRange` is very acceptable: https://caniuse.com/mdn-javascript_builtins_intl_datetimeformat_formatrange

However, if for some reason people need to support very old browsers, then then can easily install the polyfill themselves. As far as I know, ember-intl doesn't need to concern itself with that.

Installation of the polyfill is just:
1. install `npm i @formatjs/intl-datetimeformat`
2. add imports in `app.js`, for example: 
```js
import '@formatjs/intl-datetimeformat/polyfill'
import '@formatjs/intl-datetimeformat/locale-data/en' // locale-data for en
```

Feel free to ask any further questions. I'm glad to provide any clarification needed.

packages/ember-intl/addon/-private/formatjs/format-date-time-range.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type FormatDateTimeRangeParameters = Parameters<
+  IntlShape<string>['formatDateTimeRange']
+>;
+
+export function formatDateTimeRange(
+  intlShape: IntlShape,
+  ...[from, to, formatOptions]: FormatDateTimeRangeParameters
+): string {
+  return intlShape.formatDateTimeRange(from, to, formatOptions);
```
ijlee2, line 11: Can you check whether `@formatjs/intl@v3` uses positional or named arguments for `formatDateTimeRange`?

While I personally prefer named arguments (with names `from` and `to`) since there is more than 1 input, I also think it may be better for now to follow the same conventions as `formatjs`.

packages/ember-intl/addon/helpers/format-date-time-range.ts
```diff
@@ -0,0 +1,36 @@
...
+  @service declare intl: IntlService;
+
+  constructor() {
+    // eslint-disable-next-line prefer-rest-params
+    super(...arguments);
+
+    // @ts-expect-error: Property 'onLocaleChanged' is private and only accessible within class 'IntlService'.
+    this.intl.onLocaleChanged(this.recompute, this);
+  }
+
```
ijlee2, line 29: I removed the usages of `recompute()` in #1932.

```suggestion
```

packages/ember-intl/addon/services/intl.ts
```diff
@@ -138,6 +140,27 @@ export default class IntlService extends Service {
...
+    },
+  ): string {
+    if (
+      from === undefined ||
+      from === null ||
+      to === undefined ||
+      to === null
+    ) {
+      return '';
+    }
```
ijlee2, line 157: Can you temporarily comment out this code, to check how `formatjs` handles these 3 edge cases?

- Only `from` is `undefined` or `null`.
- Only `to` is `undefined` or `null`.
- Both `from` and `to` are `undefined` or `null`.

Assuming that we want to show an empty string for all 3 cases, let's split the conditional into two for simplicity:

```ts
if (from === undefined || from === null) {
  return '';
}

if (to === undefined || to === null) {
  return '';
}
```

packages/ember-intl/addon/-private/formatjs/format-date-time-range.ts
miguelcobain: These are the docs for `formatDateTimeRange`: https://formatjs.github.io/docs/react-intl/api/#formatdatetimerange

The signature is
```ts
function formatDateTimeRange(
  from: number | Date | string,
  to: number | Date | string,
  options?: Intl.DateTimeFormatOptions & {format?: string}
): string
```

If meant the react package of formatjs, then it uses named arguments:
```tsx
<FormattedDateTimeRange
  from={new Date('2020-1-1')}
  to={new Date('2020-1-15')}
/>
```

But this is because (afaik) react components don't support positional arguments. Even with a simple formatted date, they use them
```tsx
<FormattedDate
  value={new Date(1459832991883)}
  year="numeric"
  month="long"
  day="2-digit"
/>
```

In the case of ember-intl, to be consistent with the other helpers, I think it makes sense to have dates as positional and formatting options as named. I am open to suggestions, of course.

packages/ember-intl/addon/services/intl.ts
miguelcobain: | from/to     | null                                | undefined | new Date() | false |
| ---            | ---                                 | ---  |--- |---
| null           | `1/1/1970`                     | (throws Invalid time value error) | `1/1/1970 – 12/23/2024` | `1/1/1970`  |
| undefined  |  (throws Invalid time value error)        | (throws Invalid time value error) | (throws Invalid time value error) | (throws Invalid time value error) |
| new Date() | `12/23/2024 – 1/1/1970` |  (throws Invalid time value error) | `12/23/2024` | `12/23/2024 – 1/1/1970` |
| false           | `1/1/1970`                     | (throws Invalid time value error) | `1/1/19 70 – 12/23/2024` | `1/1/1970` |

In summary:
- `null` and `false` behave just like `0`, which is the beginning of unix epoch "1/1/1970"
- `undefined` is invalid and hard errors

Keep in mind that:
- in ember-intl's service `formatDate`, we are already doing
```ts
if (value === undefined || value === null) {
  return '';
}
```

So, we are forcing `null` being `''`  instead of `1/1/1970`.
I replicated the same conditions to keep things consistent between helpers.

I will do the separate arg validations as you suggested. Please let me know if these findings change something and you would like me to change other validations across the board.

ijlee2: Thanks for checking how `formatjs` handles `undefined` and `null`.

For me, at least, it makes more sense to show an empty string by default (versus showing `1/1/1970` or causing a runtime error) when the values aren't well-defined. So let's keep the early exits as is, similarly to what we do for other `format*()` methods.

packages/ember-intl/addon/-private/formatjs/format-date-time-range.ts
ijlee2: To clarify, I was wondering if `intlShape.formatDateTimeRange()`, in `v3`, still accepts `from` and `to` as positional inputs/arguments, instead of named (via a POJO).

According to its types, it still expects positional arguments, so let's define our helper to expect positional arguments as well.

tests/ember-intl/tests/integration/helpers/format-date-range/input-is-a-string-test.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable no-irregular-whitespace, qunit/no-commented-tests */
```
ijlee2, line 1: **TODO**

Uncomment these tests after https://github.com/formatjs/formatjs/pull/4784 is merged and released.

ijlee2:
@miguelcobain What do you think about renaming the helper to `{{format-date-range}}`? I think this has a couple of benefits:

- The term `DateTime` is a technical term and an implementation detail of `Intl`/`formatjs`. The [documentation for `Intl.DateTimeFormat.formatRange()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange) refers a few times to the layman term "date range." The layman term can make the helper's purpose and expected output more obvious to first-time users.
- There are fewer characters to type and 1 fewer word to camel-case in `*.{gjs,gts}`.

If you strongly believe that `{{format-date-time-range}}` is better, because it follows the same name as `formatjs`, I am open to reverting the 5th commit. Because `react-intl` uses the name `<FormattedDateTimeRange>` and `vue-intl` the name `formatTimeRange`, I think it's okay for us (Ember) to arrive at a name that best suits our needs.

tests/ember-intl/tests/integration/helpers/format-date-time-range/input-is-a-date-test.ts
```diff
@@ -30,7 +30,10 @@ module(
         </div>
       `);
 
-      assert.dom('[data-test-output]').hasText('1/23/2014 – 2/26/2015');
+      assert
+        .dom('[data-test-output]')
+        .hasText(/1\/23\/2014\s–\s1\/26\/2014/)
+        .hasText('1/23/2014 – 1/26/2014');
```
ijlee2: These assertions show how end-developers can either use a regular expression or the special character corresponding to `<0x2009>`.

tests/ember-intl/tests/integration/helpers/format-date-time-range/input-is-a-date-test.ts
```diff
@@ -42,15 +45,14 @@ module(
...
+        .hasText(/23\.–26\.01\.2014/)
+        .hasText('23.–26.01.2014');
     });
 
     test('we can format the date', async function (this: TestContext, assert) {
       await render<TestContext>(hbs`
-        <div data-test-output="1">
-          {{format-date-time-range this.from this.to format='hhmmss'}}
-        </div>
-
```
ijlee2: As @miguelcobain noted, [`@formatjs/intl`'s `formatDateTimeRange()`](https://github.com/formatjs/formatjs/blob/%40formatjs/intl%403.0.4/packages/intl/src/dateTime.ts#L118-L123
) doesn't support passing `format` as of `v3.0.4`, like it does with `formatDate()` and `formatTime()`.

To avoid wrong expectations, I decided to not document what would happen if we pass `format`.

tests/ember-intl/tests/integration/helpers/format-date-time-range/input-is-falsy-test.ts
```diff
@@ -12,44 +12,45 @@ module(
...
-    `);
+        <div data-test-output>
+          {{format-date-time-range 0 0}}
+        </div>
+      `);
 
       assert
         .dom('[data-test-output]')
-        .hasText(new Intl.DateTimeFormat('en-us').format(0));
+        .hasText(new Intl.DateTimeFormat('en-us').formatRange(0, 0));
```
ijlee2: **bugfix**

`format()` wasn't the right method from `Intl.DateTimeFormat` to use. By using `formatRange()`, we also ensure that it works on modern browsers without the need for a polyfill.

tests/ember-intl/tests/integration/helpers/format-date-range/input-is-a-string-test.ts
miguelcobain: Another option would be to keep the tests uncommented and use `test.skip()` instead of `test()` (just in case you didn't know 😊).

miguelcobain:
@ijlee2 I'm totally in agreement with renaming to `{{format-date-range}}`.
Something to note is that this will matter less the more people adopt template imports, since they can just alias the helper to what they want.
But as ember-intl's imports/api, I totally agree. Maps better to `{{format-date}}`.

tests/ember-intl/tests/integration/helpers/format-date-range/input-is-a-string-test.ts
ijlee2: I do. But the unsupported string case causes Glint to error and it's not easy to add a `@glint-expect-error` in a multi-line invocation.

ijlee2 (approved):
Thanks again for your help.

Before releasing `v7.1.0`, I'd like to see `docs/ember-intl` updated so that there is a new documentation for page `{{format-date-range}}`. Can you create a separate pull request to help out with this?

Starting next week, I'll be on holidays until early January, so I can only help out with review and release this Friday. There's no rush, so feel free to enjoy your holidays first.

PR #1938, opened 2025-01-02
Update @formatjs/intl to 3.1.0, to allow passing ISO strings to {{format-date-range}} helper

miguelcobain:
A new version of`@formatjs/intl` has been released which addresses the known issues of `formatDateTimeRange`, mopre specifically the ability to pass strings as arguments and the ability to use custom formats.
For reference, these were the commits/PR that added these features.

- https://github.com/formatjs/formatjs/pull/4784
- https://github.com/formatjs/formatjs/commit/91cc5f1e0d1295807e985906530abec671e1304e#diff-d87396f7c259160fc92e7f53bd4edd5a33b731a2f536781c3b2dd95a020ad195R42

This PR brings back the strings test ~and adds the cases where we use the `format` argument~.
It also updates the docs appropriately.


tests/ember-intl/tests/integration/helpers/format-date-range/input-is-falsy-test.ts
```diff
@@ -25,7 +25,6 @@ module(
     test('input is an empty string', async function (assert) {
       await render(hbs`
         <div data-test-output>
-          {{! @glint-expect-error }}
```
miguelcobain, line 28: This was giving me lint errors, not sure why.
I removed it and everything seems to work fine. Tests and lint pass.

docs/ember-intl/app/templates/docs/helpers/format-list.md
```diff
@@ -1,6 +1,6 @@
 # &#123;&#123;format-list&#125;&#125;
 
-Uses [`Intl.ListFormat`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) to join an array of strings.
+Uses [`Intl.ListFormat#format()`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/format) to join an array of strings.
```
miguelcobain, line 3: I did this kind of update to the specific format method, because I think the method resembles much more the helper.
Also, since we introduced `{{format-date-range}}`, we need a way to differentiate between `format()` and `formatRange()`.
In the future, a similar thing might happen if we introduce `{{format-number-range}}`.

ijlee2:
Hi, @miguelcobain. I'm still on travel and don't have my laptop to help with release, hopefully at the end of this week.

I'm hesitant to expand the role of the configuration file `format.js`, because this file didn't work when I had tried to make `ember-intl` a v2 addon. Since DateRange, in most cases, may be used to show dates, I wonder if we can use the existing default format for date instead. I'd need more time to think about this issue (how to define defaults in the future).

For now, can you update the PR so that the latest version of formatjs package(s) are installed, tests for the string case are enabled, and the documentation page(s) show that ISO strings are allowed? In addition, if you can create a changeset (let's go with a patch release) and include it in your PR (instructions in CONTRIBUTING), I can merge it without modifying it with help from my laptop.

Thanks!

miguelcobain:
> I'm hesitant to expand the role of the configuration file format.js, because this file didn't work when I had tried to make ember-intl a v2 addon. Since DateRange, in most cases, may be used to show dates, I wonder if we can use the existing default format for date instead.

The shape of the object that the `format.js` file exports is completely up to formatjs itself. In this case, the upstream authors decided to use a separate key named `dateTimeRange`. You can check here where that was done: https://github.com/formatjs/formatjs/commit/91cc5f1e0d1295807e985906530abec671e1304e#diff-d87396f7c259160fc92e7f53bd4edd5a33b731a2f536781c3b2dd95a020ad195R134

So, unless we want to "massage" the object before feeding it into formatjs (which I don't think is good because it breaks expectations), we will have to stick with the shape that formatjs expects.

Regarding the v2 addon migration, I think that whatever path we end up using for the other cases will also be applied here. Since they're all in the same file/object, I think the migration will not be impacted by this.

ijlee2:
To clarify, it's not the key name `dateTimeRange` that currently concerns me. More that, allowing default formats for `{{date-time-range}}` increases this PR's scope (update `@formatjs/intl` to the latest so that end-developers can pass ISO strings) and maintenance work. The blueprint for `ember-intl` needs to be changed, too.

I'd like additional time to think about default formats, so let's please decouple this from updating the package.

miguelcobain:
Ok, I can do that.
But just be aware that, as soon as the package is updated, users will immediately be able to use custom formats in date-time range.
So, it's just a matter of us informing our users of that fact or not.

miguelcobain:
@ijlee2 please check https://github.com/ember-intl/ember-intl/pull/1939 and let me know if something else is needed 🙏

docs/ember-intl/app/templates/docs/helpers/format-list.md
ijlee2: I think changing the URL to a more specific page is okay, but let's leave the text that end-developers see to what it is now (something that is easy to read and understand). I'll make this change.

tests/ember-intl/tests/integration/helpers/format-date-range/input-is-falsy-test.ts
ijlee2: Until `@formatjs/intl@3.1.0`, its `formatDateTimeRange()` didn't allow strings to be passed, so we had to tell Glint to ignore the type errors. Now that strings can be passed, Glint is letting us know that we don't need this override anymore.

ijlee2 (approved):
I appreciate your keeping track of the updates in `@formatjs/intl`. I will reach out to you when I have had more time to think about the default formats.

docs/ember-intl/app/formats.js
```diff
@@ -28,4 +28,12 @@ export default {
       second: 'numeric',
     },
   },
+
+  dateTimeRange: {
```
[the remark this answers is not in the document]
ijlee2: I didn't see #1939 until after reviewing this PR, my bad.

PR #1939, opened 2025-01-11
bump @formatjs/intl

miguelcobain:
Same as https://github.com/ember-intl/ember-intl/pull/1938, but without documenting `format` for `format-date-range`.


ijlee2:
Fixed in #1938.

PR #1944, opened 2025-01-28
Allowed @ember/test-helpers@5.x

mkszepp:
## Why?

When we update `@ember/test-helpers` to v5.x, we are runing into this `peerDependency` issue.

## Solution?

We should allow latest `@ember/test-helpers` to remove the `peerDependency` issue inside an app


ijlee2 (approved):
Thanks for updating the dependencies!

PR #1949, opened 2025-02-11
Enabled recommendedTypeChecked and removed DefinitelyTyped packages

ijlee2:
## Why?

By consuming (shared) base configurations, I can more easily maintain this repo. After consuming native types from `ember-source` in the addon, I forgot to remove DefinitelyTyped packages.

## Solution?

The pull request (1) enables [`recommendedTypeChecked` from `typescript-eslint`](https://typescript-eslint.io/users/configs/#recommended-type-checked) and (2) removes DefinitelyTyped packages (i.e. rely on native types from `ember-source`) from the addon.

The former occurs through wrapping packages from [`@ijlee2-frontend-configs`](https://github.com/ijlee2/frontend-configs/) in the shared configurations (in the `configs` folder). These packages are used (i.e. tested) in other open-sourced projects of mine.

I updated the source code in the addon so that I could either fix or ignore lint (type) errors.


packages/ember-intl/addon-test-support/setup-intl.ts
```diff
@@ -23,15 +22,15 @@ export function setupIntl(
...
+    const intl = owner.lookup('service:intl');
 
     intl.setLocale(locale);
     intl.setOnMissingTranslation(onMissingTranslation);
 
     if (translations) {
-      addTranslations(locale, translations);
+      await addTranslations(locale, translations);
+    } else {
+      await settled();
```
ijlee2, line 33: Note, [`addTranslations()` calls `await settled()`](https://github.com/ember-intl/ember-intl/blob/v7.1.3/packages/ember-intl/addon-test-support/add-translations.ts#L23).

To fix the lint error, I went with an if-else statement so that `setupIntl()` doesn't call `await settled()` twice in the if-case.

PR #1960, opened 2025-03-31
Removed inject as service (while still supporting ember-source@3.28)

mkszepp:
## Why?

In ember v6.3 there was added deprecation for service import from inject https://deprecations.emberjs.com/id/importing-inject-from-ember-service/

This changes allows us to hold support for `ember-source` <= 4.0 ([see changelog](https://github.com/emberjs/ember.js/blob/main/CHANGELOG.md#v410-december-28-2021)) and removes deprecation in newer versions

### DIFF
--- a/.changeset/hungry-singers-help.md
+++ b/.changeset/hungry-singers-help.md
@@ -1,0 +1,6 @@
+---
+"ember-intl": patch
+"test-app-for-ember-intl": patch
+---
+
+Removed inject as service (while still supporting ember-source@3.28)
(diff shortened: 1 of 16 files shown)


ijlee2 (approved):
Thanks for coming up with a solution that we can use now, without having to drop support for `ember-source@3.28` early.

I updated `ember-source` to `6.3.0` in #1961, then rebased your branch, so that I can verify your solution in CI.

Before:



After:

PR #1963, opened 2025-04-23
Warn instead of error for MISSING_DATA

lfloyd117:
## Why?

Chromium stopped supporting eu-es and is-is locales, which we localize for. This has been the case for a long time and yet our localization has still worked in those locales.
This should really be a warning instead of breaking a whole app. 
See this issue for more info: https://github.com/formatjs/formatjs/issues/4352 

We've been working around this by overriding the `onIntlError` method in our wrapper service, but as of v7.1.6 that method has been removed and the error handling can't be hooked into more. You can use `setOnFormatjsError` but the errors for the aforementioned locales throw during the translation load in `intl`'s constructor so we cannot set it early enough. 

If you would rather a configurable override be added instead, I am happy to change gears to that, but I saw others having this same exact issue and thought it best to change this.

## Solution?

Log a warning in the console instead of throwing an error.

### DIFF
--- a/.changeset/fresh-pianos-carry.md
+++ b/.changeset/fresh-pianos-carry.md
@@ -1,0 +1,5 @@
+---
+"ember-intl": patch
+---
+
+Warn MISSING_DATA issues instead of erroring
--- a/docs/ember-intl/app/templates/docs/migration/v7.md
+++ b/docs/ember-intl/app/templates/docs/migration/v7.md
@@ -125,6 +125,8 @@ for (const locale of this.intl.locales) {
 }
 ```
 
+- The `onIntlError()` method has been removed. See `setOnFormatjsError()` below.
+
 - The `translationsFor()` method has been removed. If you need to check the existence of translations, use `exists()` instead.
 
 ```diff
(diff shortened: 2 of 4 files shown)


ijlee2:
I shared my thoughts [on Discord](https://discord.com/channels/480462759797063690/487648547685269515/1365290838762192908):

> I think your solution is the best approach. I checked that `MISSING_DATA` is only used "when some native Intl APIs don't support certain locales, or missing locale-data when polyfills are setup." That is, existing users likely already fixed these issues, so it isn't a problem if the next patch/minor logs an error message instead.
>
> https://formatjs.github.io/docs/guides/develop/#error-codes
>
> Will merge the PR and release a new version shortly.
>
> It is annoying that the "hydrate" code in the `constructor()` (which I'm guessing had been added so that we can use `ember-intl` right away without having to load translations) creates an order dependency. It's also one of the things that I didn't know how to recreate, when I had attempted to convert `ember-intl` to `v2`. So I wonder if, going forward, we should ask users to load translations.

These links showed, at the time of writing, Chromium has yet to support `eu-es` and `is-is`.

- https://issues.chromium.org/issues/402532201
- https://issues.chromium.org/issues/40624456

PR #1972, opened 2025-06-09
Deprecated defining formats in app/formats.js

### LINKED ISSUE #1942: Create setFormats()
## Background

While trying to convert `ember-intl` to a v2 addon, I noticed that [how `ember-intl` retrieves an end-developer's default formats](https://github.com/ember-intl/ember-intl/blob/v7.1.1/packages/ember-intl/addon/services/intl.ts#L93) won't work. We should use `@embroider/macros` so that `ember-intl` runs on classic Ember as well as on Embroider.

## TODOs

- Plan and document migration (from `app/formats.js` to `ember-cli-build.js`).
- Decide on name conventions for keys (e.g. `date-range`, `dateRange`, or `format-date-range` for `{{format-date-range}}` helper).
- Write a codemod.

https://github.com/embroider-build/embroider/tree/main/packages/macros

ijlee2:
## Why?

Closes #1942.

The use of `app/formats.js` is problematic if `ember-intl` were to become a v2 addon. Currently, the `intl` service has [an order dependency in `constructor()`](https://github.com/ember-intl/ember-intl/blob/v7.2.0/packages/ember-intl/addon/services/intl.ts#L99) and needs [`getOwner(this).resolveRegistration('formats:main')`](https://github.com/ember-intl/ember-intl/blob/v7.2.0/packages/ember-intl/addon/services/intl.ts#L313) to read the file defined in the consuming app.

In addition, the parent key names in `app/formats.js` (i.e. `date`, `dateTimeRange`, `number`, `relative`, and `time`) refer to the names that `@formatjs/intl` uses. In other words, the file `app/formats.js` leaks implementation details.

## Solution?

For simplicity, I went with @bertdeblock's idea of calling a method (named `setFormats()`), instead of asking people to install and configure `@embroider/macros`. The formats (parent keys have been renamed) are to be defined in `app/ember-intl.{js,ts}`.

```ts
/* app/ember-intl.ts */
import type { Formats } from 'ember-intl';

export const formats: Formats = {
  'format-time': {
    hhmmss: {
      hour: 'numeric',
      minute: 'numeric',
      second: 'numeric',
    },
  },
};
```

A few things to note:

- With the new formats, the key names match the names of `ember-intl`'s helpers.
- `ember-intl` provides `Formats`, the expected type for formats, in the barrel file. People can use it to check for typos and incorrect values.
- Currently, the path `app/ember-intl.{js,ts}` doesn't matter, because people will explicitly import their `formats`, then pass it to `ember-intl` via `setFormats()`.
- `setFormats()` is expected to be called in `app/routes/application.{js,ts}`, along with `setLocale()`, etc. Currently, the order doesn't matter: `setFormats()` can appear before or after `setLocale()`.

People can move away from `app/formats.js` today. See https://github.com/ember-intl/ember-intl/pull/1972/commits/08bcaa7c6a52fe28cf203c1cd18c88cf92bc47c2 for an example.

### DIFF
--- a/packages/ember-intl/addon/services/intl.ts
+++ b/packages/ember-intl/addon/services/intl.ts
@@ -96,7 +96,15 @@ export default class IntlService extends Service {
     // eslint-disable-next-line prefer-rest-params
     super(...arguments);
 
-    this.getDefaultFormats();
+    const hasNewConfiguration = Boolean(
+      // @ts-expect-error: Property 'resolveRegistration' does not exist on type 'Owner'
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      getOwner(this).resolveRegistration('ember-intl:main'),
+    );
+
+    if (!hasNewConfiguration) {
+      this.getDefaultFormats();
+    }
 
     // Hydrate
     translations.forEach(([locale, translations]: [string, Translations]) => {
--- a/packages/ember-intl/addon/services/intl.ts
+++ b/packages/ember-intl/addon/services/intl.ts
@@ -8,16 +8,19 @@ import { tracked } from '@glimmer/tracking';
 import type {
   FormatDateParameters,
   FormatDateRangeParameters,
+  FormatjsFormats,
   FormatListParameters,
   FormatMessageParameters,
   FormatNumberParameters,
   FormatRelativeParameters,
   FormatRelativeTimeParameters,
+  Formats,
   FormatTimeParameters,
   IntlShape,
   OnErrorFn,
 } from '../-private/formatjs/index';
 import {
+  convertToFormatjsFormats,
   createIntl,
   createIntlCache,
   formatDate,
@@ -40,6 +43,8 @@ import {
 import { flattenKeys, type Translations } from '../-private/utils/translations';
 import translations from '../translations';
 
+export type { Formats };
+
 type OnFormatjsError = (error: Parameters<OnErrorFn>[0]) => void;
 
 type OnMissingTranslation = (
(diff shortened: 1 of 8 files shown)


packages/ember-intl/addon/services/intl.ts
```diff
@@ -96,7 +101,15 @@ export default class IntlService extends Service {
     // eslint-disable-next-line prefer-rest-params
     super(...arguments);
 
-    this.getDefaultFormats();
+    const hasNewConfiguration = Boolean(
+      // @ts-expect-error: Property 'resolveRegistration' does not exist on type 'Owner'
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      getOwner(this).resolveRegistration('ember-intl:main'),
```
bertdeblock, line 107: I would let users decide where to store their formats, and not force them to use a specific file. Otherwise, I don't see how this is any different from the deprecated `app/formats.js` file. This solution also uses the resolver, which is something the Ember team wants to get rid of if I'm not mistaken.

bertdeblock:
```
Currently, the path app/ember-intl.{js,ts} does matter, because the check for the new configuration file uses getOwner(this).resolveRegistration('ember-intl:main'). ember-intl may allow arbitrary paths (e.g. app/configs/ember-intl.{js,ts}) in v8 (TBD), since people can explicitly import their formats, then pass it to ember-intl via setFormats().
```

Missed this in the description. Migration path still feels a bit weird though.

ijlee2: No problem. Yeah, a possible extra step in migration isn't ideal; I didn't think of an easier way to check for the existence of the new configuration file, which would allow people to start creating this file in `>=7.3.0`.

PR #1974, opened 2025-06-09
Passed the correct locales to _onMissingTranslation()

ijlee2:
## Why?

The `intl` service's `t()` didn't pass the correct `locales` to the user's `onMissingTranslation()`. Existing tests didn't catch this, because [`_onMissingTranslation()` used by `setupIntl()` doesn't use `locales`](https://github.com/ember-intl/ember-intl/blob/v7.3.0/packages/ember-intl/addon-test-support/-private/on-missing-translation.ts#L42-L48).

### DIFF
--- a/tests/ember-intl/tests/integration/helpers/t/input-is-a-translation-key-test.ts
+++ b/tests/ember-intl/tests/integration/helpers/t/input-is-a-translation-key-test.ts
@@ -39,14 +39,20 @@ module(
         </div>
 
         <div data-test-output="2">
+          {{t "smoke-tests.hello.world" locale="en-us"}}
+        </div>
+
+        <div data-test-output="3">
           {{t "smoke-tests.hello.world" locale="fr-fr"}}
         </div>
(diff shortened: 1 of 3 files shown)


tests/ember-intl/tests/integration/helpers/t/input-is-a-translation-key-test.ts
```diff
@@ -39,14 +39,20 @@ module(
...
         </div>
       `);
 
       assert.dom('[data-test-output="1"]').hasText('Hallo Welt!');
 
+      assert.dom('[data-test-output="2"]').hasText('Hello world!');
+
       assert
-        .dom('[data-test-output="2"]')
+        .dom('[data-test-output="3"]')
```
ijlee2, line 55: With the default `_onMissingTranslations()`, users and this test will see the following text:

```
Missing translation "smoke-tests.hello.world" for locale "fr-fr"
```

PR #1975, opened 2025-06-09
Converted ember-intl to v2 addon

ijlee2:


NullVoxPopuli:
Making this compatible with a v2 environment is going to require some iteration, and I think it'd be a little unwieldily to do in one PR.

If we merge the runtime/v2-addon changes, then I can PR separately some build changes -- (I'm kinda having to re-implement a bunch in private code, and would love to bring that in to this repo)

ijlee2:
Hi, @frederikbosch. I plan on converting `ember-intl` to v2 addon for `v8.0.0`. I first wanted to remove excess code to ease the update, so that's why the beta versions still show `ember-intl` as a v1 addon.

PR #1990, opened 2025-09-30
fix: remove `ember-cli-typescript` and move `typescript` to devDeps

aklkv:
## Why?

It took us some time to figure out why our builds are failing for incorrect dependency chain. We were able to pinpoint to `typescript` being in peerDeps in the `ember-intl` package. I am not sure if there is a good reason to keep it as such, but it should normally be under `devDeps`.

## Solution?

* move `typescript` to `devDeps` 
* bonus, removed `ember-cli-typescript` and add config to handle typescript processing on `ember-cli-babel` level

### DIFF
--- a/.changeset/clean-streets-search.md
+++ b/.changeset/clean-streets-search.md
@@ -1,0 +1,5 @@
+---
+"ember-intl": patch
+---
+
+Removed ember-cli-typescript from dependencies and moved typescript to devDependencies
--- a/packages/ember-intl/ember-cli-build.js
+++ b/packages/ember-intl/ember-cli-build.js
@@ -7,6 +7,10 @@ module.exports = function (defaults) {
   const { project } = defaults;
 
   const options = {
+    'ember-cli-babel': {
+      enableTypeScriptTransform: true,
+    },
+
     autoImport: {
       webpack: {
         node: {
--- a/packages/ember-intl/index.js
+++ b/packages/ember-intl/index.js
@@ -32,6 +32,12 @@ module.exports = {
   name: 'ember-intl',
   configOptions: null,
 
+  options: {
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -27,10 +27,10 @@
     "sort-package-json": "^3.4.0",
     "update-workspace-root-version": "^2.0.5"
   },
-  "packageManager": "pnpm@9.15.9",
+  "packageManager": "pnpm@10.17.1+sha512.17c560fca4867ae9473a3899ad84a88334914f379be46d455cbf92e5cf4b39d34985d452d2583baf19967fa76cb5c17bc9e245529d0b98745721aa7200ecaf7a",
   "engines": {
     "node": "18.* || >= 20",
-    "pnpm": ">= 9"
+    "pnpm": ">= 10"
```
ijlee2, line 33: Let's update `pnpm` in a separate PR. I can help with this.

aklkv:
> Hi, @aklkv. Thanks for reporting a possible issue.
> 
> I'm a bit confused because `typescript` is currently listed as an optional peer dependency and this repo has several test apps and addons. I'd would like to know more what kind of setup your project has (e.g. package manager? monorepo? JavaScript-only?). Could you provide me a minimal reproduction?

Well our setup is hard to describe:

* pnpm monorepo
* multiple local addons
* engines
* one app
* all consume [design system](https://github.com/hashicorp/design-system), which recently added `ember-intl` as peerDependancy, this is where the problem started

We started having two different version, but while pairing with @mansona on this we have narrowed down the issues to `ember-intl` having `typescript` in peerDeps which seems to be very strange

ijlee2:
Thanks! I can check out Hashicorp's `design-system` repo later and see what's going on.

PR #2006, opened 2025-10-22
Limited build options

ijlee2:
## Why?

[Before `v6`](https://github.com/ember-intl/ember-intl/blob/v6.0.0-beta.1/index.js#L53-L63), the number of build options that can be configured became too much. The more build options ember-intl allows, the harder it is to maintain code, standardize usage, and pave a way forward for Vite apps.

## Solution?

The following options have been removed.

- `excludeLocales`
- `includeLocales`
- `outputPath`
- `requiresTranslation`
- `stripEmptyTranslations`
- `verbose`

I limited the options to the following (subject to change):

```ts
const defaultConfig = {
  errorOnMissingTranslations: false,
  errorOnNamedArgumentMismatch: false,
  fallbackLocale: undefined,
  inputPath: 'translations',
  publicOnly: false,
  wrapTranslationsWithNamespace: false,
};
```

Now that there's fewer code, we can make additional refactors. (I will do this in a separate pull request.)

> [!NOTE]
>
> We can even remove `errorOnMissingTranslations` and `errorOnNamedArgumentMismatch` if `ember-intl` provides a separate package for linting translations.

### DIFF
--- a/.changeset/mighty-taxis-show.md
+++ b/.changeset/mighty-taxis-show.md
@@ -1,0 +1,6 @@
+---
+"ember-intl": major
+"test-ember-intl-node": minor
+---
+
+Limited build options


packages/ember-intl/lib/broccoli/translation-reducer/index.js
```diff
@@ -98,9 +76,7 @@ class TranslationReducer extends CachingWriter {
     mkdirSync(filePath, { recursive: true });
 
     for (const locale in translations) {
-      if (this.options.verbose) {
-        this.validateMessages(translations[locale], locale);
-      }
+      this.validateMessages(translations[locale], locale);
```
ijlee2: **bugfix**

I think, even when `verbose` is `false`/`undefined`, we do want to validate translation messages. Not sure when this bug had been introduced.

Eventually, we'll want to handle linting translations separately from building them.

PR #2015, opened 2025-10-30
Simplified message for setOnMissingTranslation() in tests

ijlee2:
## Why?

When a translation is missing, the `intl` service uses [`setOnMissingTranslation()`](https://github.com/ember-intl/ember-intl/blob/v8.0.0-beta.4/packages/ember-intl/src/services/intl.ts#L323-L325) to display some string. The default value for development and production has been [`Missing translation "${key}" for locale "${locale}"`](https://github.com/ember-intl/ember-intl/blob/v8.0.0-beta.4/packages/ember-intl/src/services/intl.ts#L81). The value is different for testing.

Until now, `setupIntl()` replaced the default value with a string that weakly asserts the translation key and data. The code that converts data (an arbitrary object) to a string wasn't easy to maintain. It increased the package size while providing limited value, since we can't know from weak assertions what our end-users see in reality.

## Solution?

I simplified the value by asserting only the translation key. This matches the approach that @buschtoens had initially taken in [#590](https://github.com/ember-intl/ember-intl/pull/590/commits/af5143d503fe16f3b96bd5bc8bb5540cdae54419#diff-7fced4221531ecb0a077caea61e86fbbd31be82dee2731dba0e0d46f0230820eR54).

End-developers will need to update their `hasText()` or `includesText()` assertions, similarly to what's shown below:

```diff
// Without data
- assert.dom().hasText('t:hello.message:()');
+ assert.dom().hasText('t:hello.message');

// With data
- assert.dom().hasText('t:hello.message:("name":"Zoey")');
+ assert.dom().hasText('t:hello.message');
```

<details>

<summary>Update alternative 1</summary>

The best way to update is to ensure that translations are loaded in tests, so that you can write assertions that are strong and don't leak implementation details.

```diff
- assert.dom().hasText('t:hello.message:("name":"Zoey")');
+ assert.dom().hasText('Hello, Zoey!');
```

</details>

<details>

<summary>Update alternative 2</summary>

If you can't load translations and want to weakly assert data, call `setOnMissingTranslation()` in your custom `setupRenderingTest()`.

```ts
/* tests/helpers/index.ts */
import { setupIntl } from 'ember-intl/test-support';
import {
  setupRenderingTest as upstreamSetupRenderingTest,
  type SetupTestOptions,
} from 'ember-qunit';

function setupRenderingTest(hooks: NestedHooks, options?: SetupTestOptions) {
  upstreamSetupRenderingTest(hooks, options);
  setupIntl(hooks, 'en-us');

  hooks.beforeEach(function () {
    const intl = this.owner.lookup('service:intl');

    intl.setOnMissingTranslation((key, _locale, data) => {
      // Return some string
    });
  });
}

export { setupRenderingTest };
```

</details>

### DIFF
--- a/.changeset/tall-regions-cross.md
+++ b/.changeset/tall-regions-cross.md
@@ -1,0 +1,11 @@
+---
+"ember-intl": major
+"my-v1-classic-app-with-lazy-loaded-translations": patch
+"my-v1-app-with-lazy-loaded-translations": patch
+"my-v1-app-with-namespace-from-folders": patch
+"my-v1-app-with-fallbacks": patch
+"test-app-for-ember-intl": patch
+"docs-app-for-ember-intl": patch
+---
+
+Simplified message for setOnMissingTranslation() in tests


buschtoens (approved):
Good riddance! Thank you. 😄

PR #2145, opened 2026-05-14
feat(typescript): expose public method types

BobrImperator:
## Why?

`ember-intl` consumers can't easily import public interface's types and must derive from the `IntlService` shape. Where the types themselves are derived from `@formatjs/intl` as well described by the `-private/formatjs/*` module.  Although its worth noting that `ember-intl` also extends `@formatjs/intl` types.

Consider a below example:
```ts
const options: Intl.NumberFormatOptions = {}
```

Currently it can only be typed as:
```ts
import type { IntlService } from 'ember-intl';
const options: Parameters<IntlService['formatNumber']>[1] = {}
```

## Solution?

The PR proposes standardizing and exporting types of public methods so users can easily re-define their signatures and use concrete types without additional churn.

```ts
import type {  FormatNumberOptions } from 'ember-intl';
const options: FormatNumberOptions = {}
```

## Changes

- Refactors `services/intl.ts` to declare types before using them with methods
- Adds public exports of the `ember-intl` types
Additionally exports `TranslationJson` which has to be copied by consumers currently.

### DIFF
--- a/.changeset/bold-swans-push.md
+++ b/.changeset/bold-swans-push.md
@@ -1,0 +1,10 @@
+---
+"ember-intl": major
+---
+
+Exports argument types for each public method of the `IntlService`.
+```ts
+import type { FormatNumberOptions } from 'ember-intl';
+const options: FormatNumberOptions = {}
+```
+
--- a/packages/ember-intl/src/index.ts
+++ b/packages/ember-intl/src/index.ts
@@ -8,4 +8,27 @@ export { default as formatRelativeTime } from './helpers/format-relative-time.ts
 export { default as formatTime } from './helpers/format-time.ts';
 export { default as t } from './helpers/t.ts';
 export { default as tKey } from './helpers/t-key.ts';
-export type { Formats, default as IntlService } from './services/intl.ts';
+export type {
+  FormatDateArg,
+  FormatDateOptions,
+  FormatDateRangeFrom,
+  FormatDateRangeOptions,
+  FormatDateRangeTo,
+  FormatDisplayNameArg,
+  FormatDisplayNameOptions,
+  FormatListArg,
+  FormatListOptions,
+  FormatMessageArg,
+  FormatMessageOptions,
+  FormatNumberArg,
+  FormatNumberOptions,
+  FormatRelativeTimeArg,
+  FormatRelativeTimeOptions,
+  Formats,
+  FormatTimeArg,
+  FormatTimeOptions,
+  default as IntlService,
+  OnFormatjsError,
+  OnMissingTranslation,
+  TranslationJson,
+} from './services/intl.ts';
--- a/packages/ember-intl/src/services/intl.ts
+++ b/packages/ember-intl/src/services/intl.ts
@@ -42,7 +42,48 @@ import {
   type TranslationJson,
 } from '../-private/utils/translations.ts';
 
-export type { Formats };
+type FormatDateArg = FormatDateParameters[0] | undefined | null;
+type FormatDateOptions = FormatDateParameters[1] & {
+  locale?: string;
+};
+
+type FormatDateRangeFrom = FormatDateRangeParameters[0] | undefined | null;
+type FormatDateRangeTo = FormatDateRangeParameters[1] | undefined | null;
+type FormatDateRangeOptions = FormatDateRangeParameters[2] & {
+  locale?: string;
+};
+
+type FormatDisplayNameArg = FormatDisplayNameParameters[0] | undefined | null;
+type FormatDisplayNameOptions = FormatDisplayNameParameters[1] & {
+  locale?: string;
+};
+
+type FormatListArg = FormatListParameters[0] | undefined | null;
+type FormatListOptions = FormatListParameters[1] & {
+  locale?: string;
+};
+
+type FormatMessageArg = FormatMessageParameters[0] | string | undefined | null;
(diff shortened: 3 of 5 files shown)


ijlee2 (commented):
Hmm, it's an interesting suggestion. I'm hesitant to follow it at the moment.

In general, I strongly believe in keeping an API to a minimum and covering the 80% case, so that it's easier to maintain and extend code. Currently, the pull request exposes 20 additional types and I'm not sure how many people or files really need to know the exact type for `intl`'s `format*` methods (I wrote "really" in the sense that maybe a refactor could be done on the consumer's side).

> so that users can easily re-define their signatures

Can you provide me more context (examples of code would help) for how you are using the `format*` methods such that you needed access to the types?

BobrImperator:
You're right to say the status quo covers the vast majority of cases, in our experience there's been only a few cases that needed this. However in my experience complex types should generally be exported as a best practice. In Typescript this isn't particularly necessary, since consumers can cheat using Typescript utilities such as `Parameters` and type lookups. But for example the Rustlang compiler requires type aliases and interfaces to be exported alongside public functions, because otherwise consumers aren't able to construct these types themselves.

To expand on the example from the description, the example code is a parameterized currency formatting helper. The helper can be parameterized to display currencies in an alternative way. The `options` object needs to be typed ahead of an if statement that modifies it.

How the function must look like today:
```ts
  import type { IntlService } from 'ember-intl';
  const numericValue = isNaN(Number(value)) ? 0 : Number(value);
  const precisionValue = precision !== undefined ? Number(precision) :
  DEFAULT_PRECISION;

  const options: Parameters<IntlService['formatNumber']>[1] = {
    style: 'currency',
    currency: currency ?? DEFAULT_CURRENCY,
    minimumFractionDigits: precisionValue,
    maximumFractionDigits: precisionValue,
  };

  if (compact) {
    options.notation = 'compact';
    options.compactDisplay = 'short';
    options.minimumFractionDigits = 0;
    options.maximumFractionDigits = precisionValue;
  }

  return this.intl.formatNumber(numericValue, options);
```

What this PR proposes:
```ts
  import type { FormatNumberOptions } from 'ember-intl';
  const numericValue = isNaN(Number(value)) ? 0 : Number(value);
  const precisionValue = precision !== undefined ? Number(precision) :
  DEFAULT_PRECISION;

  const options: FormatNumberOptions = {
    style: 'currency',
    currency: currency ?? DEFAULT_CURRENCY,
    minimumFractionDigits: precisionValue,
    maximumFractionDigits: precisionValue,
  };

  if (compact) {
    options.notation = 'compact';
    options.compactDisplay = 'short';
    options.minimumFractionDigits = 0;
    options.maximumFractionDigits = precisionValue;
  }

  return this.intl.formatNumber(numericValue, options);
```

ijlee2:
When I write codemods, I sometimes use TypeScript's `Parameters` to refer to a type from `recast`, because the package that I use, `@codemod-utils/ast-javascript`, doesn't provide the type. This occurs relatively rarely, though, compared to the number of times I call a "visit method" from `recast`, so I've been okay with the `Parameters`-approach.

In the case of your code, I recommend refactoring it as follows. I'll present it in the context of a helper.

```ts
import { type Registry as Services, service } from '@ember/service';

const DEFAULT_CURRENCY = 'USD';
const DEFAULT_PRECISION = 2;

type Value = number | null | undefined;

type Options = {
  compact?: boolean;
  currency?: string;
  precision?: number;
};

interface MyFormatCurrencySignature {
  Args: {
    Named?: Options;
    Positional: [Value];
  };
  Return: string;
}

export default class MyFormatCurrencyHelper extends Helper<MyFormatCurrencySignature> {
  @service declare intl: Services['intl'];

  compute(
    [value]: MyFormatCurrencySignature['Args']['Positional'],
    options: MyFormatCurrencySignature['Args']['Named'],
  ): MyFormatCurrencySignature['Return'] {
    const currency = options?.currency ?? DEFAULT_CURRENCY;
    const numericValue = value ?? 0;
    const precision = options?.precision ?? DEFAULT_PRECISION;

    const defaultOptions = {
      currency,
      style: 'currency',
    } as const;

    if (options?.compact) {
      return this.intl.formatNumber(numericValue, {
        ...defaultOptions,
        compactDisplay: 'short',
        maximumFractionDigits: precision,
        minimumFractionDigits: 0,
        notation: 'compact',
      });
    }

    return this.intl.formatNumber(numericValue, {
      ...defaultOptions,
      maximumFractionDigits: precision,
      minimumFractionDigits: precision,
    });
  }
}
```

Some things to note:

- `ember-intl` (really, `Intl`) allows two additional types for `value`: `BigInt` and `Intl.StringNumericLiteral`. The helper above restricts the type to a minimum, `number | null | undefined`, defines it independently of the underlying dependency (no `Parameters` or exposed type like `FormatNumberOptions`) so that you may more easily swap the dependency later if needed.
- Use `?.`, `??`, and early exits to simplify logic and help TypeScript and other developers.
- Having two `return` statements may seem excessive. It could be that having a separate for the `compact` case (some business logic) allows you to more easily change the definition later.

(I'll think about whether it'd make sense to expose types for `options`.)

ijlee2:
I have decided not to go with the feature suggestion to keep the public API and maintenance cost small. See https://github.com/ember-intl/ember-intl/pull/2145#issuecomment-4459261306 for possible alternatives.

PR #2146, opened 2026-05-14
feat(ember-intl): make `primaryLocale`  to have a definite type

BobrImperator:
## Why?

`ember-intl` has changed the type of `primaryLocale` between versions, relaxing it to allow an `undefined` locale. However an `undefined` value means locales aren't set and any `intl` method will throw anyway.

## Solution?

This PR tightens the types back to either return a `string` or and `Error` saying to `setLocale` first.

### DIFF
--- a/.changeset/primary-locale-definite.md
+++ b/.changeset/primary-locale-definite.md
@@ -1,0 +1,10 @@
+---
+"ember-intl": minor
+---
+
+`IntlService.primaryLocale` is now typed as `string` (no longer `string | undefined`) and throws when accessed before `setLocale()` has been called with a valid locale. 
+An will be thrown error with a helpful message, instructing to make sure to `setLocale` before accessing any methods. 
+```js
+"ember-intl: locale is missing. Make sure to call `intl.setLocale(...)` first."
+```
+
--- a/packages/ember-intl/src/services/intl.ts
+++ b/packages/ember-intl/src/services/intl.ts
@@ -88,12 +88,16 @@ export default class IntlService extends Service {
     return Object.keys(this._intls);
   }
 
-  get primaryLocale(): string | undefined {
-    if (!this._locale) {
-      return;
+  get primaryLocale(): string {
+    const primary = this._locale?.[0];
+
+    if (!primary) {
+      throw new Error(
+        'ember-intl: locale is missing. Make sure to call `intl.setLocale(...)` first.',
+      );
     }
 
-    return this._locale[0];
+    return primary;
   }
 
   addTranslations(locale: string, translations: TranslationJson): void {
@@ -379,13 +383,12 @@ export default class IntlService extends Service {
 
   private updateDocumentLanguage(): void {
     const html = getHtmlElement(this);
-    const { primaryLocale } = this;
 
-    if (!html || !primaryLocale) {
+    if (!html) {
       return;
     }
 
-    html.setAttribute('lang', primaryLocale);
+    html.setAttribute('lang', this.primaryLocale);
   }
 
   private updateIntl(
--- a/tests/ember-intl/tests/unit/services/intl/getters-test.ts
+++ b/tests/ember-intl/tests/unit/services/intl/getters-test.ts
@@ -1,6 +1,5 @@
 import type { TestContext as BaseTestContext } from '@ember/test-helpers';
 import type { IntlService } from 'ember-intl';
-import { setupIntl } from 'ember-intl/test-support';
 import { module, test } from 'qunit';
 import { setupTest } from 'test-app-for-ember-intl/tests/helpers';
 
@@ -10,7 +9,6 @@ interface TestContext extends BaseTestContext {
 
 module('Unit | Service | intl > getters', function (hooks) {
   setupTest(hooks);
-  setupIntl(hooks, 'en-us');
 
   hooks.beforeEach(function (this: TestContext) {
     this.intl = this.owner.lookup('service:intl');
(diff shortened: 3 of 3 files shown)


ijlee2 (changes requested):
Hi, @BobrImperator. Thanks for suggesting a DX improvement and providing a changeset.

While I agree that fewer possibilities for a value are nice to have, this is a breaking change from versions `5.7.x` to `8.x`, where the type for `primaryLocale` has been `string | undefined`. In `v5` and `v6`, when `ember-intl` didn't have native types, it incorrectly listed the type as `string` ([v5.7.0](https://github.com/ember-intl/ember-intl/blob/v5.7.0/addon/services/intl.d.ts#L54), [v6.5.6](https://github.com/ember-intl/ember-intl/blob/v6.5.6/packages/ember-intl/addon/services/intl.d.ts#L48)), but the implementation shows that `undefined` is, in theory, possible ([v5.7.0](https://github.com/ember-intl/ember-intl/blob/v5.7.0/addon/services/intl.js#L59), [v6.5.6](https://github.com/ember-intl/ember-intl/blob/v6.5.6/packages/ember-intl/addon/services/intl.js#L66-L68)).

In addition, the `throw` statement that you added still doesn't solve the issue that I'd like to solve: If a user forgets to call `intl.setLocale` but makes use of `intl.locales`, they will see an error message that is cryptic, e.g.

```sh
TypeError: locales is not iterable
    at IntlService.t (intl.ts:354:26)
    at THelper.compute
    at ClassicHelperManager.getValue
```

It'd be nicer if we can use `assert` from `@ember/debug` somewhere to indicate that `setLocale` needs to be called.

Can you provide me more context for why you needed a change in the type? Were you perhaps trying to update `ember-intl` from `v5` or `v6`? If you were making use of `intl.primaryLocale`, can you provide me the code so that I can make suggestion(s) for an alternative?

BobrImperator:
Thanks for the response. I'd like to also add that this is also helpful for moving away from `locale` property which was also removed.

I agree with you overall, the thing I propose here is a stop-gap and I do believe it'd be best if the service could have a clear "setup" step that _must_ happen. When it doesn't then the service is essentially in a broken state.
[ember-simple-auth#setup](https://github.com/mainmatter/ember-simple-auth/blob/master/packages/ember-simple-auth/src/services/session.ts#L368) does this by requiring `#setup` to be called. When it isn't then any methods used will begin to throw. 
I'd argue this is the same for `ember-intl` today since `setLocale` must be called now, unlike before (if I'm not mistaken it used to have default en-us locale).

I've been debating myself whether we'd want a real error or an assert so I think we should ultimately use your suggestion to use `@ember/debug` here.

Lastly to give you some context. What motivated the PR specifically was an upgrade from `5.7.2` -> `8.2.2` which removed some properties and changed some type expectations. In the simplest case, consider such code:

The following is allowed under `5.x` because `primaryLocale` is a `string` type. The function expects a string and there no guards anywhere.
```ts
// Function definition
  export function firstDayOfWeek(locale: string): DayOfWeekValue {
      const intlLocale = new Intl.Locale(locale);                                                                                                                                                          
      return (intlLocale.weekInfo.firstDay % 7) as DayOfWeekValue;
   }
   
// Callsite
firstDayOfWeek: DayOfWeekValue = this.args.firstDayOfWeek ?? firstDayOfWeek(this.intl.primaryLocale);
```

The same example adjusted for `>5.x` would have to implement guards somehow. Be it at at parent-level component, inside the function itself or a getter.
```ts
// Function definition
  export function firstDayOfWeek(locale: string): DayOfWeekValue {
      const intlLocale = new Intl.Locale(locale);                                                                                                                                                          
      return (intlLocale.weekInfo.firstDay % 7) as DayOfWeekValue;
   }
   
// Callsite
firstDayOfWeek: DayOfWeekValue | undefined =  this.intl.primaryLocale && (this.intl. this.args.firstDayOfWeek ?? firstDayOfWeek(this.intl.primaryLocale));
```

It probably doesn't matter really where it's done, but the fact that it has to be checked. Which overall seems unsound for the `intl` service because without calling `setLocale` any of the methods called will begin to fail exactly as you've mentioned.
In fact what we've ended up doing is to patch / extend the ember-intl service with a similar implementation to this PR  in order to provide such guarantee.

ijlee2:
I see, much appreciate the additional context and your help with updating `ember-intl` to the latest major version.

> it'd be best if the service could have a clear "setup" step that must happen.

I agree. You might have noticed, the documentation site currently recommends creating a private method called `setupIntl`, where end-developers are expected to call 1 or more `set` methods from the `intl` service. I think it'd be nice if we can call 1 setup method in the future.

> `setLocale` must be called now, unlike before (if I'm not mistaken it used to have default en-us locale).

Yep, [`en-us` was included by default until `v6.5.6`](https://github.com/ember-intl/ember-intl/blob/v6.5.6/packages/ember-intl/addon/services/intl.js#L122). Related breaking change in `7.0.0`: https://ember-intl.github.io/ember-intl/docs/migration/v7#breaking-changes-required-locale-in-test-helpers.

---

I'd migrate the code that you shared by using the non-null assertion operator `!`, because you the developer have more knowledge than TypeScript about when the code will be run (some time after `setLocale` is called) and what type `primaryLocale` will have then.

```ts
this.args.firstDayOfWeek ?? firstDayOfWeek(this.intl.primaryLocale!)
```

You've likely installed `typescript-eslint` and use [`@typescript-eslint/no-unnecessary-type-assertion`](https://typescript-eslint.io/rules/no-unnecessary-type-assertion/) (enabled in `recommended`). If `ember-intl` does change the type of `primaryLocale` to `string` only, the autofix for `eslint` will remove the unnecessary `!`'s.

Once I can [get CI for `main` to pass](https://github.com/ember-intl/ember-intl/pull/2148), I can update https://ember-intl.github.io/ember-intl/docs/migration/v7 so that it lists the type change as a breaking change. Feel free to open the PR for this.

ijlee2:
For an update, I will try to include the feature suggestion in `ember-intl@9.0.0`. This assumes a minimal change to how end-developers use `ember-intl`.

ijlee2:
Closed this in favor of #2192.

PR #2158, opened 2026-05-22
fix(v1-compat): funnel project translations folder to ignore non-translation files

### LINKED ISSUE #2153: v1-compat: empty non-translation file (.keep) in translations/ produces ["", {}] entry → INVALID_CONFIG at runtime
## Summary

`@ember-intl/v1-compat` (v1.0.9) emits an `["", {}]` entry into the inlined `translations.js` when the project's `translations/` folder contains an empty non-translation file (e.g. a `.keep` placeholder used to keep an empty dir in git). At runtime the generated `load-translations` instance initializer calls `intl.addTranslations("", {})`, which makes `@formatjs/intl` throw:

```
[@formatjs/intl Error INVALID_CONFIG] "locale" was not configured, using "" as fallback.
```

## Reproduction

1. New Ember app with `ember-intl@8.x` + `@ember-intl/v1-compat@1.0.9`.
2. `config/ember-intl.js`:
   ```js
   module.exports = () => ({ publicOnly: false, inputPath: 'translations' });
   ```
3. `touch translations/.keep` (zero-byte file).
4. `ember build`.
5. Open app — initializer fires, throws `INVALID_CONFIG`.

`grep '\[\["",{' dist/assets/vendor*.js` confirms the bad entry in the inlined translations array.

## Root cause

Two cooperating issues:

**1. `lib/broccoli/build-translation-tree.js` does not filter the project's translation tree.**

Addon translations get funneled with an extension filter:

```js
function funnelTranslations(tree, destDir) {
  return funnel(tree, {
    destDir: destDir || '.',
    include: ['**/*.yaml', '**/*.yml', '**/*.json'],
  });
}
```

But the project's own translations folder is added unfiltered:

```js
if (existsSync(projectTranslations)) {
  trees.push(new WatchedDir(projectTranslations));   // no include filter
}
```

So any non-translation file in `./translations/` (`.keep`, `.DS_Store`, README, etc.) reaches the reducer.

**2. `lib/utils/translation-reducer/get-translations.js` returns `{}` for empty files.**

```js
if (file === '') {
  return {};
}
```

Combined with the reducer's skip check:

```js
let translationObject = getTranslations(filePath);
if (!translationObject) {
  return accumulator;
}
```

`{}` is truthy, so the file is NOT skipped. Then the locale is derived from the filename:

```js
const fileName = basename(filePath).split('.')[0];   // '.keep' -> '' -> ''
const locale = normalizeLocale(fileName);            // ''
```

…and `accumulator[""] = {}` is committed. In `mergeTranslationFiles: true` mode this becomes `["", {}]` at the head of the generated array, which the v1-compat `load-translations` initializer iterates and feeds into `intl.addTranslations`.

## Suggested fix

Either fix alone resolves it; both would be defensive:

1. In `build-translation-tree.js`, funnel the project tree the same way addon trees are funneled:
   ```js
   trees.push(funnelTranslations(new WatchedDir(projectTranslations)));
   ```
2. In `translation-reducer.js`, skip files where `basename(filePath).split('.')[0]` is empty, or have `get-translations.js` return `undefined` (not `{}`) for empty files so the existing `if (!translationObject)` guard works.

## Workaround

Remove `.keep` (or any other non-translation file) from `translations/`. If the empty directory needs to stay in git, use `.gitkeep` outside of `translations/`, or `.gitignore` the contents and let the build/fetch script `mkdir -p` it.

## Environment

- `ember-intl`: 8.2.2
- `@ember-intl/v1-compat`: 1.0.9
- `ember-source`: 5.12.0
- `ember-cli`: 5.12.0
- Node: 20.x

mlenderink:
Fixes #2153.

## Summary

The addon translation trees built by `@ember-intl/v1-compat` are already funneled with an include filter of `*.{json,yaml,yml}`, but the project's own `translations/` folder was added to the merged tree as a `WatchedDir` without any filter. Any non-translation file (e.g. a `.gitkeep` placeholder used to keep an empty `translations/` folder in git) therefore reached the translation reducer, which derived an empty locale from the filename and produced a `["", {}]` entry at the head of the inlined `translations.js` array.

At runtime, the `load-translations` instance initializer then called `intl.addTranslations("", {})`, which made `@formatjs/intl` throw:

```
[@formatjs/intl Error INVALID_CONFIG] "locale" was not configured, using "" as fallback.
```

## Fix

Wrap the project tree with the same `funnelTranslations` helper that is already used for addon trees, so only `*.{json,yaml,yml}` files reach the reducer.

## Test

Added `tests/ember-intl-v1-compat/tests/lib/broccoli/build-translation-tree/ignores-non-translation-files.test.ts`. The test sets up a project with `translations/.gitkeep` (empty) alongside a real `en-us.json` and asserts the merged tree contains only the JSON file. Verified the test fails on `main` without the fix and passes with it.

## Changeset

Added a `minor` changeset for `@ember-intl/v1-compat`, per the issue discussion.

### DIFF
--- a/.changeset/v1-compat-funnel-project-translations.md
+++ b/.changeset/v1-compat-funnel-project-translations.md
@@ -1,0 +1,5 @@
+---
+"@ember-intl/v1-compat": minor
+---
+
+Filter the project's `translations/` folder through the same funnel (`*.{json,yaml,yml}`) that is already applied to addon translation trees. Previously, non-translation files such as a `.gitkeep` placeholder used to keep an empty `translations/` folder in git would reach the translation reducer, which derived an empty locale from the filename and inserted a `["", {}]` entry in the inlined `translations.js`. At runtime, the `load-translations` initializer then called `intl.addTranslations("", {})`, which made `@formatjs/intl` throw `INVALID_CONFIG`. See [#2153](https://github.com/ember-intl/ember-intl/issues/2153).
--- a/packages/ember-intl-v1-compat/lib/broccoli/build-translation-tree.js
+++ b/packages/ember-intl-v1-compat/lib/broccoli/build-translation-tree.js
@@ -80,7 +80,10 @@ function buildTranslationTree(project, inputPath, treeGenerator) {
   const projectTranslations = join(project.root, inputPath);
 
   if (existsSync(projectTranslations)) {
-    trees.push(new WatchedDir(projectTranslations));
+    // Funnel so that non-translation files (e.g. a `.gitkeep` used to keep
+    // the folder in git) are filtered out before reaching the reducer. The
+    // addon translation trees above are already funneled the same way.
+    trees.push(funnelTranslations(new WatchedDir(projectTranslations)));
   }
 
   return [mergeTrees(trees, { overwrite: true }), addonsWithTranslations];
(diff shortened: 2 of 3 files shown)


ijlee2 (approved):
Thanks very much!

PR #2192, opened 2026-08-12
Strictened types for primaryLocale and setLocale()

ijlee2:
## Why?

Supercedes #2146.

From `5.x` to `8.x`, the getter `intl.primaryLocale` had the type of `string | undefined`. The type `undefined` shouldn't occur in practice, because `intl.setLocale()` is assumed to have been called first (in the `application` route).

Similarly, from `5.x` to `8.x`, the method `intl.setLocale` had the type of `string | string[]`. The type `string[]` allows an empty array, which shouldn't occur in practice.

## Solution?

`primaryLocale`: After refactoring code, I cherry-picked the commits from #2146, then replaced `throw new Error()` with `assert()` from `@ember/debug`. This is so that the conditional check doesn't occur in production.

To reduce scope, I didn't check which of the remaining `string`'s in the `intl` service could be changed to the "singular" type `Locale`. The type `Locale` (defined to be `string`) would be defined in `packages/ember-intl/src/-private/utils/locale.ts`.

### DIFF
--- a/packages/ember-intl/src/services/intl.ts
+++ b/packages/ember-intl/src/services/intl.ts
@@ -1,3 +1,4 @@
+import { assert } from '@ember/debug';
 import { cancel, next, type Timer as EmberRunTimer } from '@ember/runloop';
 import Service from '@ember/service';
 import { htmlSafe } from '@ember/template';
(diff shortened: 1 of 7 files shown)


packages/ember-intl/src/services/intl.ts
```diff
@@ -369,17 +377,16 @@ export default class IntlService extends Service {
...
     const html = getHtmlElement(this);
-    const { primaryLocale } = this;
 
-    if (!html || !primaryLocale) {
+    if (!html) {
       return;
     }
 
-    html.setAttribute('lang', primaryLocale);
+    html.setAttribute('lang', this.primaryLocale);
```
ijlee2, line 385: We can skip the `primaryLocale` check because `setLocale()` doesn't allow passing an empty array anymore.

PR #2202, opened 2026-09-03
Recommended installing ember-intl in addons as a peer dependency

abeforgit:
## Why?

After [a discussion with the maintainer](https://discord.com/channels/480462759797063690/487648547685269515/1545050435164045344), it became clear that ember-intl expects to be run in an app where only one version is present, due to the potential for clashes when 2 versions of the `intl` service would exist in the app.

This is not particularly unique to ember-intl, but it is one of the few addons that have explicit docs for usage in other addons.

## Solution?

Remove the mention of installing it as a dep, adjust the example `pnpm` invocations, and add a brief explainer at the top of the file

### DIFF
--- a/.changeset/warm-bananas-wear.md
+++ b/.changeset/warm-bananas-wear.md
@@ -1,0 +1,5 @@
+---
+"docs-app-for-ember-intl": minor
+---
+
+Recommended installing ember-intl in addons as a peer dependency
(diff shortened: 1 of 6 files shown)


ijlee2 (dismissed):
Thanks again for your question and helping out with updating the documentation. I didn't know about `--save-peer` (quite cool)!

PR #2203, opened 2026-09-07
Improved assertion that the user has called `setLocale`

bertdeblock:
Got the following error: `TypeError: locales is not iterable`, because I had not called `setLocale` yet. This should hopefully make it a bit more clear.

### DIFF
--- a/packages/ember-intl/src/services/intl.ts
+++ b/packages/ember-intl/src/services/intl.ts
@@ -84,7 +84,7 @@ export default class IntlService extends Service {
 
   private get _localesOrThrow(): Locales {
     assert(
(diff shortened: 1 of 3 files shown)


packages/ember-intl/src/services/intl.ts
```diff
@@ -84,7 +84,7 @@ export default class IntlService extends Service {
 
   private get _localesOrThrow(): Locales {
     assert(
-      'No locales set. Did you call `intl.setLocale()` with one or more locales?',
+      'No locales set. Did you call `intl.setLocale()`?',
```
ijlee2, line 87: Let's let TypeScript tell developers what they need to pass to `setLocale`.

ijlee2 (approved):
✨

PR #2207, opened 2026-09-14
Add translationHelpers config to @ember-intl/lint

johanrd:
## Why

Be able to use the nice linter also for other import paths than  from`ember-intl`.

The linter only recognizes `t` and `tKey` imported from `ember-intl` (plus the `intl` service). Apps that move translation to module scope, so strict-mode templates, template-only components and plain-function helpers can import one `t`, otherwise get no `no-missing-keys` or `no-unused-keys` coverage: the keys are silently invisible.

The option is data, matched against the import source exactly as written. When a source matches, its local binding is treated as `t` or `tKey`, so the existing template (`{{t "key"}}`, `(t "key")`) and JavaScript (`t("key")`) detection applies unchanged. Default is no extra helpers, so behaviour is unchanged. `.hbs` files have no imports and are not affected.

## Solution
```js
// ember-intl.config.mjs
export default {
  translationHelpers: [
    { source: 'my-app/utils/intl', export: 't', kind: 't' },
    { source: 'my-app/utils/translation-key', export: 'default', kind: 'tKey' },
  ],
};
```

Cowritten by claude


ijlee2:
@johanrd Thanks for keeping your branch up-to-date.

Since expanding an API may increase maintenance cost and shouldn't be undone without a major release, I want to take some time to think about your solution and possible alternatives. I also need to limit my time with open source at the moment. I hope you'll be okay with the delay in review.

johanrd:
@ijlee2 sure, no problem. Running it currently as a patch locally, so no hurry at all.

